PR #22, opened 2025-11-15
Removed non applicable db functionality

LandynDev:
We store to DB to serve dashboard. This means we'll only be storing and don't need get functions so those are removed.

Changes:
- removed get functions from repositories
- merged repositories into single repository.py file (each specific repository only contained one function each after being removed of extraneous code)

### DIFF
--- a/auto_updater/auto_update/pm2_interface.py
+++ b/auto_updater/auto_update/pm2_interface.py
@@ -1,13 +1,15 @@
-import subprocess
 import json
-import time
 import logging
-from typing import Optional, Dict, List, Any
+import subprocess
+import time
 from dataclasses import dataclass
+from typing import Any, List, Optional
+
 
 @dataclass
 class PM2ProcessInfo:
     """Information about a PM2 process."""
+
     name: str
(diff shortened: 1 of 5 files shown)


gittensor/validator/storage/queries.py
```diff
@@ -8,87 +8,7 @@
...
-"""
-
-BULK_UPSERT_REPOSITORIES = """
-INSERT INTO repositories (full_name, name, owner, weight)
-VALUES %s
-ON CONFLICT (full_name)
-DO NOTHING
-"""
-
 BULK_UPSERT_PULL_REQUESTS = """
```
entrius: I notice when storing to pull_requests previously we used `SET_PULL_REQUEST`, should we update this query so that we actually upsert and `hotkey`, `uid`, `earned_score` and perhaps any other fields are updated? 

this may cause discrepancies in the dashboard when we update scoring especially and the dashboard won't reflect the base earned_score, or if we want to display miner stats the associated miner could possibly be stale (neuron dereg but reregisters)

Issue #25, opened 2025-11-18
Penalize typo fix PRs


anderdc:
Some miners are harvesting score through implementing typo fixes. We'd like to move away from that behavior by adjusting score for those types of PRs.

Strategy:
- For this issue we only wish to solve the typo fix issue, no need for other ad hoc scoring adjustments
- adjust score at the **PR level**. No need to penalize at the miner level or repo level.
- core solution should be added at `gittensor/validator/evaluation/reward.py` in the `score_pull_requests` function
- preferably right before setting PR score (`pr.set_earned_score`), some logic is implemented to determine if the PR is a typo only PR. If logic determines PR is typo only, PR score is set to `0.01`, this value should be in `constants.py`
- you can make helper functions as needed, please put them in files that are most sensible/create if necessary
- confidence of PR being typo only must be `>=85%`. this threshold should be in `constants.py`
- add one `.debug` log that says if a PR was determined to be typo only fix. include miner UID and PR.

Constraints:
- total changes should amount to **250 or less**
- follow repository conventions when appropriate (commenting style, variable names, sensible helper functions, etc)
- senseless code. drawn out code, or code that can be obviously optimized to try to hit the max line constraint will **not be accepted**
- do not provide/add any tests

PR #26, opened 2025-11-18
feat: typo detection

James-4u:
Detect typo only PR and apply penalty

Fixed issue #25 

### DIFF
--- a/gittensor/validator/evaluation/spam_detection.py
+++ b/gittensor/validator/evaluation/spam_detection.py
@@ -10,16 +10,10 @@ from gittensor.validator.utils.spam_detection import (
 from gittensor.utils.utils import mask_secret
 
 def detect_typo_only_pr(pr: PullRequest) -> bool:
-    if not pr.file_changes:
-        return False
-    
     file_patches = [
         fc.patch for fc in pr.file_changes
         if fc.patch and isinstance(fc.patch, str)
     ]
-
-    if not file_patches:
-        return False
     
     is_typo_only = is_typo_only_pr(file_patches)
     return is_typo_only


gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,209 @@
...
+
+        return True
+
+    typo_like = 0
+    total = len(pairs)
+
+    for old, new in pairs:
+        category = classify_change(old, new)
+
+        if category in ("formatting", "punctuation", "comment_typo", "typo", "safe_small_edit"):
```
anderdc: why is `safe_small_edit` considered a typo change?

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,209 @@
...
+
+                # Allow up to 3-character spelling change
+                if dist <= 3 or sim >= 0.7:
+                    continue
+
+                all_ok = False
+
+    return all_ok
+
+def classify_change(old: str, new: str) -> str:
```
anderdc: this function is not optimal, it runs the same for loop twice,  I like the end of the function more, this can be made better

SmartDever02: got your point, two loop looks the same but they are used for different purpose.
they are all word-by-word checks for the typo but
- first block is to detect typo changes in comment logic
- second block is to detect typo changes in normal/code text 
but am going to extract the shared logic

gittensor/validator/utils/spam_detection.py
SmartDever02: safe_small_edit is actually not a typo change it's more like typo-like changes. for example

```
- print("Hello world")
+ print("Hello world ")
```

```
- "Connection timed out"
+ "Connection timeout"
```

```
- "Loading..."
+ "Loading.."
```

```
- /* wait for response */
+ /* waiting for response */
```

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,209 @@
...
+            return False
+
+    return (typo_like / total) >= TYPO_RATIO_THRESHOLD
+
+def is_typo_only_pr(file_patches: List[str]) -> bool:
+    if not file_patches:
+        return False
+
+    for patch in file_patches:
+        if not is_typo_only_patch(patch):
```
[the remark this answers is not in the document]
SmartDever02: okay, sounds great - btw I think we should increase the threshold. 25 lines are possible to hit.
thinking about 50 as default, let me know your thoughts.

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,169 @@
...
+    for line in lines:
+        if line.startswith(("@@", "+++", "---")):
+            continue
+        if line.startswith("-") and not line.startswith("--"):
+            removed.append(line[1:])
+        elif line.startswith("+") and not line.startswith("++"):
+            added.append(line[1:])
+
+    count = min(len(removed), len(added))
+    return list(zip(removed[:count], added[:count]))
```
LandynDev: Can we simplify this function to something like this? Feel like its doing a bit now that can otherwise read easier.

```
def extract_line_pairs_from_patch(patch: str) -> List[Tuple[str, str]]:
    removed = []
    added = []
    
    for line in patch.split("\n"):
        if line.startswith("-") and not line.startswith(("--", "---")):
            removed.append(line[1:])
        elif line.startswith("+") and not line.startswith(("++", "+++")):
            added.append(line[1:])
    
    return list(zip(removed, added))
```

Benefits:
Removed redundant skip logic - Lines starting with `@@, +++, ---` naturally fail the other conditions
Removed manual truncation - zip() automatically stops at the shorter list, so count and slicing are unnecessary

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,169 @@
...
+
+    for old, new in pairs:
+        category = classify_change(old, new)
+
+        if category in ("formatting", "punctuation", "comment_typo", "typo", "safe_small_edit"):
+            typo_like += 1
+        elif category == "danger":
+            return False
+
+    return (typo_like / total) >= TYPO_RATIO_THRESHOLD
```
LandynDev: I don't find the classification to be the most useful here either. It seems to be a complexity additional for marginal benefits / insights. I mean, there's a lot of 'lower level' stuff going on that's difficult to read with parsing, regex, etc. This is obviously kind of a requirement to detect typos, but the complexity that comes with it makes me favor only implementing what is **absolutely necessary** in order to keep complexity minimal.

I almost think that this should simply read as:
```
for old, new in pairs:
        if typo_change_detected_for_pair(old,new):
             typo_like +=1

return (typo_like / len(pairs)) >= TYPO_RATIO_THRESHOLD
```

I'd also think that this `typo_change_detected_for_pair` that I just made up would be a remodeled version of `classify_change`. I almost imagine that this `typo_change_detected_for_pair` function may be as simple as this:

```
def typo_change_detected_for_pair(old: str, new: str) -> bool:
    return is_token_typo(old,new) or is_comment_typo(old, new)
```

I really don't see a need to detect things beyond tokenized detected typos and comment typos. Other stuff seems to be a good bit of complexity for marginal gain. IMO if it caught an extra 2% of PRs or something but with more complexity, we're fine without it.

gittensor/validator/utils/spam_detection.py
SmartDever02: agreed - it's amazing

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,169 @@
...
+    pairs = extract_line_pairs_from_patch(patch)
+
+    if not pairs:
+        lines = patch.split("\n")
+        has_added = any(l.startswith("+") and not l.startswith("++") for l in lines)
+        has_removed = any(l.startswith("-") and not l.startswith("--") for l in lines)
+
+        # No added/removed = no content changes → typo-like
+        if not has_added and not has_removed:
+            return True
```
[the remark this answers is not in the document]
SmartDever02: I don't remember the edge case I found from dashboard but as long as we are mega simplifying everything it's okay

gittensor/validator/utils/spam_detection.py
SmartDever02: well `is_comment_typo()` was designed to detect typo changes inside comment as it was not getting caught by `is_token_typo()` - but now I don't feel any needs to keep it - we can just use  is_token_typo()

gittensor/validator/utils/spam_detection.py
```diff
@@ -0,0 +1,169 @@
...
+            return "comment_typo"
+
+    if not is_comment_line(old) and not is_comment_line(new):
+        in_string = ('"' in old or "'" in old or '"' in new or "'" in new)
+
+        if not in_string:
+            if introduces_code_symbols(new) or contains_keywords(new):
+                return "danger"
+
+    return "unknown"
```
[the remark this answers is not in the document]
SmartDever02: removed them all - make it mega simple

PR #54, opened 2025-12-02
Fix confusing recycle allocation logic in dynamic emissions

NikolaPantel:
## Problem

The dynamic emissions recycle allocation had confusing logic with a backwards ternary expression that would incorrectly add 1.0 emissions when nothing should be recycled.

**Buggy code (line 60):**
```python
scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, 1 if total_recycled <= 0 else 0)
```

**Logic evaluation:**
- If `total_recycled > 0`: `max(total_recycled, 0)` = `total_recycled` ✅
- If `total_recycled <= 0`: `max(total_recycled, 1)` = `1.0` ❌

This means when no emissions should be recycled (total_recycled ≤ 0), the system would incorrectly allocate 1.0 emissions to RECYCLE_UID.

## Impact

**Incorrect Behavior:**
- When `final_scalar = 1.0` (100% emissions distributed), `total_recycled = 0`
- Buggy code adds `1.0` to RECYCLE_UID
- **Result:** Extra 1.0 emissions created out of thin air

**Scenario:**
```python
# High network activity: final_scalar = 1.0
total_original = 100.0
total_recycled = 0.0  # Nothing should be recycled

# Buggy: max(0.0, 1) = 1.0
scaled_rewards[RECYCLE_UID] += 1.0  # ❌ Incorrect

# Total emissions = 100.0 + 1.0 = 101.0 (inflation!)
```

## Fix (1 line)

Simplified to ensure non-negative recycled amount:

```python
scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, 0.0)
```

**Correct behavior:**
- If `total_recycled > 0`: Recycle that amount ✅
- If `total_recycled <= 0`: Recycle 0.0 (nothing) ✅

## Impact

- **Correctness**: Prevents emission inflation when no recycling should occur
- **Minimal**: 1-line change, zero breaking changes
- **Clarity**: Removes confusing backwards ternary logic

### DIFF
--- a/gittensor/validator/evaluation/dynamic_emissions.py
+++ b/gittensor/validator/evaluation/dynamic_emissions.py
@@ -55,11 +55,14 @@ def apply_dynamic_emissions_using_network_contributions(
     total_original = sum(normalized_rewards.values())
     scaled_rewards = {uid: reward * final_scalar for uid, reward in normalized_rewards.items()}
     total_recycled = total_original * (1 - final_scalar)
+    
+    # Dynamic bound: full recycle (1.0) if no earned scores, otherwise 0
+    dynamic_recycle_bound = 1 if total_original <= 0 else 0
 
     # Allocate recycled emissions
-    scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, 0.0)
+    scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, dynamic_recycle_bound)
 
-    recycle_percentage = (total_recycled / total_original * 100) if total_original > 0 else 0.0
+    recycle_percentage = (total_recycled / total_original * 100) if total_original > 0 else 100.0
     bt.logging.info(
         f"Dynamic emissions: lines_scalar={lines_scalar:.3f}, repo_scalar={repo_scalar:.3f}, "
         f"final={final_scalar:.2f}, recycled={total_recycled:.2f} ({recycle_percentage:.2f}%)"


LandynDev:
Yeah effectively this should never happen because these scalars only approach the upper limit of 1.0 asymptotically (they never hit 1.0). 

Therefore, the `total_recycled` will always be non zero, because `final_scalar` will never be 1.
`total_recycled = total_original * (1 - final_scalar)`.

But if we get deeper into this proposed fix:
- This does fix a logical bug and makes the code read more correctly.
- Your solution presents another bug where if `total_original` is 0, then `total_recycle` is 0. `total_recycle` should be 1 in such case.


Similar to your point and using the bug that you pointed out, I believe that the real fix is to change the original code:

```
    total_original = sum(normalized_rewards.values())
    scaled_rewards = {uid: reward * final_scalar for uid, reward in normalized_rewards.items()}
    total_recycled = total_original * (1 - final_scalar)

    # Allocate recycled emissions
    scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, 1 if total_recycled <= 0 else 0)

    recycle_percentage = (total_recycled / total_original * 100) if total_original > 0 else 0.0
```

to:

```
    total_original = sum(normalized_rewards.values())
    scaled_rewards = {uid: reward * final_scalar for uid, reward in normalized_rewards.items()}
    total_recycled = total_original * (1 - final_scalar)
    dynamic_recycle_bound = 1 if total_original <= 0 else 0 # Full recycle if no earned scores

    # Allocate recycled emissions
    scaled_rewards[RECYCLE_UID] = scaled_rewards.get(RECYCLE_UID, 0.0) + max(total_recycled, dynamic_recycle_bound)

    recycle_percentage = (total_recycled / total_original * 100) if total_original > 0 else 1.0 # This needs to be updated too
```

hsparks-codes:
Great catch @LandynDev! You're absolutely right - I missed the edge case where total_original = 0. Updated the fix to use dynamic_recycle_bound as you suggested. The code now correctly handles both cases:
- Normal operation: recycles based on scalars
- No earned scores: full recycle (1.0) to RECYCLE_UID

Thanks for the detailed reply! 🙏

LandynDev (approved):

Issue #64, opened 2025-12-04
[Bug]  `*` comment pattern causes false positives for C/C++ pointer operations


NikolaPantel:
### Summary
Similar to PR #56 (preprocessor directive bug), the comment pattern `r'^\s*\*'` incorrectly detects C/C++ pointer operations as comments, causing valid code to be marked as non-scoreable.

### Severity
**CRITICAL** - Affects heavily-weighted languages (C/C++) and is extremely common

### Problem
The pattern `r'^\s*\*'` (line 63 in `constants.py`) was intended to detect C-style multi-line comment continuations like:
```c
/*
 * This is a comment
 */
```

However, it also matches **valid C/C++ code** that starts with `*`:

#### False Positive Examples:
```c
// These are VALID CODE but detected as comments:
*ptr = malloc(sizeof(int));      // ❌ Pointer assignment
*result = calculate();            // ❌ Pointer dereference  
*p++ = value;                     // ❌ Pointer increment
*array[i] = data;                 // ❌ Pointer array access
```

### Root Cause
In C/C++, `*` at the start of a line is almost always:
- **Pointer dereference**: `*ptr = value;`
- **Pointer declaration**: `*p = &var;`
- **Multiplication**: `*result = a * b;`

It is **rarely** a comment. Multi-line comments use `/*` and `*/` patterns (already covered), not standalone `*`.

### Impact
- **High frequency**: Pointers are fundamental in C/C++ - used in almost every file
- **Significant score loss**: Miners lose points for legitimate pointer operations
- **Unfair penalty**: C/C++ contributions are systematically under-scored
- **Worse than PR #56**: Preprocessor directives appear once per file, but pointers appear dozens/hundreds of times

### Comparison to PR #56

| Aspect | PR #56 (`#` pattern) | This Bug (`*` pattern) |
|--------|---------------------|------------------------|
| Pattern | `#` for preprocessor | `*` for pointers |
| Languages | C, C++, Rust | C, C++, Java, Go |
| Frequency | Low (top of files) | **Very High** (throughout code) |
| Impact | Significant | **Critical** |

### Proposed Solution
Use the same approach as PR #56: skip the `*` pattern for languages where it's not a comment.

#### Option 1: Context-Aware Detection (Recommended)
```python
# In constants.py
NON_STAR_COMMENT_LANGUAGES = [
    'c', 'cpp', 'cc', 'cxx', 'h', 'hpp',  # C/C++
[...]

PR #68, opened 2025-12-04
Fix: Refactor is_comment_line to use Pygments for improved accuracy

langverse2023:
This PR replaces the previous regex-based, line-by-line comment detection with a more robust solution using the `Pygments` library.

**The Problem:**
The previous approach analyzed each line in isolation. This caused issues with multi-line comments (e.g., C-style `/* ... */`), where middle lines starting with `*` were often invalidly interpreted as multiplication operators or code syntax.

**The Solution:**
I have introduced a new helper function `get_comment_line_indices` that preserves the context of the code block.

**Implementation Details:**
1.  **Reconstruct Source:** The function strips diff markers (`+`, `-`) from the patch to reconstruct the continuous source text.
2.  **Context-Aware Lexing:** Instead of checking line by line, we pass the **full text** to the `Pygments` lexer. This allows the lexer to correctly identify multi-line comment blocks (like `""" docstrings """` or `/* block comments */`).
3.  **Token Mapping:** We iterate through the tokens generated by `Pygments` and map them back to the original line numbers.
4.  **Filtering:** A line is marked as "non-scoreable" only if it contains comment tokens and strictly **no code tokens**.

**Key Changes:**
- Removed `is_comment_line` (regex-based).
- Added `get_comment_line_indices` (Pygments-based).
- Updated `count_non_scoreable_lines` to pre-calculate comment indices before iterating.

**Fixes** #64
**Improves** #56

---
Contribution by Gittensor, learn more at [https://gittensor.io/](https://gittensor.io/)

### DIFF
--- a/gittensor/validator/utils/config.py
+++ b/gittensor/validator/utils/config.py
@@ -3,7 +3,7 @@ import os
 import bittensor as bt
 
 # NOTE: bump this number when we make new updates
-__version__ = "2.0.5"
+__version__ = "2.0.6"
 
 
 VALIDATOR_WAIT = 60  # 60 seconds
--- a/gittensor/validator/weights/master_repositories.json
+++ b/gittensor/validator/weights/master_repositories.json
@@ -8483,7 +8483,7 @@
     ]
   },
   "opentensor/subtensor": {
-    "weight": 200.00,
+    "weight": 78.01,
     "additional_acceptable_branches": [
       "devnet-ready"
     ]


hsparks-codes:
@langverse2023 Close this PR, otherwise will report. You're stealing my effort.

langverse2023:
> ```python
> # Test cases
> if __name__ == "__main__":
>     # Test 1: Comments only
>     """
>     /*
>     * Hello world
>     */
> 
>     """
>     test_string = """
> * Hello world
> """
>     print(f"Rust comments only: {is_comment_line(test_string, '.rs')}")  
>     
>     # Test 2: Code
>     test_string = """
> #include <stdio.h>
> """
>     print(f"C program code: {is_comment_line(test_string, '.c')}")  
>     
>     # Test 3: multi-line comment
>     """
>     '''
>     This is a comment
>     '''
>     """
>     test_string = '''
> This is a comment
> '''
>     print(f"Multi-line comment: {is_comment_line(test_string, '.py')}")
> ```
> 
> I don't think Pygments library works better than current solution as we check line by line

Thanks for the feedback! You are absolutely right that checking line-by-line is problematic for multi-line comments (e.g., Pygments might interpret an isolated * Hello world as a multiplication or pointer instead of a comment).
To address this, I have refactored the logic to analyze the entire patch context at once.

langverse2023:
@LandynDev could u check my pr

gittensor/validator/utils/spam_detection.py
```diff
@@ -27,14 +29,82 @@ def is_token_typo(old: str, new: str, max_dist=TYPO_MAX_DIST, min_sim=TYPO_MIN_S
...
+        else:
+            # Empty lines or diff headers are treated as empty strings to preserve index alignment
+            clean_content_parts.append("")
     
-    return any(re.match(pattern, content) for pattern in patterns_to_check)
+    full_text = "\n".join(clean_content_parts)
+
+    # 2. Determine the appropriate lexer
+    try:
+        filename = f"dummy{file_extension}" if file_extension else "dummy.txt"
```
anderdc: double check the dummy filename construction, I don't believe `file_extension` has the `.`. 

also why not just use the actual filename? seems strange to just make up a dummy filename

PR #73, opened 2025-12-04
fix: improve test file detection to avoid false positives

0xsatoshi99:
## Problem

The current test file detection uses a simple substring check:

```python
def is_test_file(self) -> bool:
    return "test" in self.filename.lower()
```

This causes **FALSE POSITIVES** for legitimate source files that contain "test" as part of another word:

| File | Contains "test" | Actually a test? | Current Result |
|------|-----------------|------------------|----------------|
| `contest/main.py` | con**test** | NO | ❌ Penalized |
| `latest_version.py` | la**test** | NO | ❌ Penalized |
| `attest/verify.py` | at**test** | NO | ❌ Penalized |
| `protest/handler.go` | pro**test** | NO | ❌ Penalized |
| `testimony/record.py` | **test**imony | NO | ❌ Penalized |

These files are incorrectly penalized with `0.25x` weight (`TEST_FILE_CONTRIBUTION_WEIGHT`), reducing miner scores unfairly.

## Solution

Use pattern matching for common test file naming conventions:

```python
def is_test_file(self) -> bool:
    # Check for test directories
    if "/test/" in filename or "/tests/" in filename or "/__tests__/" in filename:
        return True
    
    # Check for common test file naming patterns
    test_patterns = [
        r"^test_",           # test_main.py (Python)
        r"_test\.",          # main_test.py, main_test.go
        r"\.test\.",         # main.test.js (JavaScript/TypeScript)
        r"_spec\.",          # main_spec.rb (Ruby RSpec)
        r"^spec_",           # spec_main.rb
        r"\.spec\.",         # main.spec.ts (Angular)
    ]
    
    return any(re.search(pattern, basename) for pattern in test_patterns)
```

## Test Results

| File | Expected | Old Result | New Result |
|------|----------|------------|------------|
| `test_main.py` | True | True ✓ | True ✓ |
| `main_test.py` | True | True ✓ | True ✓ |
| `src/tests/main.py` | True | True ✓ | True ✓ |
| `main.test.js` | True | True ✓ | True ✓ |
| `main.spec.ts` | True | True ✓ | True ✓ |
| `contest/main.py` | False | True ❌ | False ✓ |
| `latest_version.py` | False | True ❌ | False ✓ |
| `attest/verify.py` | False | True ❌ | False ✓ |
| `protest/handler.go` | False | True ❌ | False ✓ |

## Impact

- **Fairer scoring**: Legitimate source files no longer penalized
- **Accurate detection**: Only actual test files get reduced weight
- **Multi-language support**: Works with Python, Go, JavaScript, TypeScript, Ruby conventions

---

*Contribution by Gittensor, learn more at https://gittensor.io/*

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,87 @@
+# Gittensor Contributor Guide
+
+## Getting Started
+
+Before contributing, please:
+
+1. Read the [README](README.md) to understand the project goals
+2. Review the [miner guide](gittensor/miner/README.md) and [validator guide](gittensor/validator/README.md)
+3. Check existing issues and discussions to avoid duplicate work
+
+## Lifecycle of a Pull Request
+
+### 1. Create Your Branch
+
+- Branch off of `test` and target `test` with your PR
+- Ensure there are no conflicts before submitting
+
+### 2. Make Your Changes
+
+- Write clean, well-documented code
+- Follow existing code patterns and architecture
+- Update documentation if applicable
+- Do NOT add comments that are over-explanatory, redundant, or extraneous to inflate line count.
+- We will NOT accept PRs that are purely testing changes, documentation changes, or typo fixes
+- When making your changes, ask yourself: will this raise the value of the repository?
+
+### 3. Submit Pull Request
+
+1. Push your branch to the repository
+2. Open a PR targeting the `test` branch
+3. Use draft PRs for work-in-progress changes
+4. Fill out the PR template with:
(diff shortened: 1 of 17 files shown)


0xsatoshi99:
@LandynDev fixed for above requirements. please check again.

anderdc:
your PRs are not pointed to the `test` branch. These will not get merged unless that is fixed

anderdc (approved):

PR #78, opened 2025-12-04
Updated logic on pr success ratio to account for closed date too

LandynDev:


SmartDever02:
I actually thought this was a room for exploits (it's what I meant in our dm) but after a deep check, I noticed that prs are only merged_prs not closed ones - so no need to add check there @LandynDev

PR #85, opened 2025-12-11
fix: check uid availability when get all uids

eureka0928:
# Fix UID filtering in `get_all_uids()` to respect serving status and validator permit limits

## Summary
Updates `get_all_uids()` to filter UIDs using `check_uid_availability()`, ensuring only eligible miners are queried. This addresses ### the known issue where validators attempted to query non-serving or inappropriate UIDs.

## Changes

`gittensor/utils/uids.py`

Updated `get_all_uids()` to use `check_uid_availability()` for filtering
Filters out high-stake validators: Excludes validators with stake above vpermit_tao_limit (default: 4096 TAO)
Improved documentation: Updated docstring to reflect the filtering behavior

## Why this matters
Previously, get_all_uids() returned all UIDs without checking:
Whether the UID was a high-stake validator (shouldn't be queried as a miner)
This caused validators to waste resources querying unavailable or inappropriate UIDs, relying on later failure handling instead of proactive filtering.

### DIFF
--- a/gittensor/utils/uids.py
+++ b/gittensor/utils/uids.py
@@ -5,13 +5,15 @@ import numpy as np
 
 
 def check_uid_availability(metagraph: "bt.metagraph.Metagraph", uid: int, vpermit_tao_limit: int) -> bool:
-    """Check if uid is available. The UID should be available if it is serving and has less than vpermit_tao_limit stake
+    """Return whether a UID is eligible for querying.
+
     Args:
-        metagraph (:obj: bt.metagraph.Metagraph): Metagraph object
-        uid (int): uid to be checked
-        vpermit_tao_limit (int): Validator permit tao limit
+        metagraph: Metagraph containing axon and stake state.
+        uid: UID to check.
+        vpermit_tao_limit: Maximum allowed stake for validator-permit UIDs.
+
     Returns:
-        bool: True if uid is available, False otherwise
+        True if the UID is serving and within the validator-permit stake limit.
     """
     # Filter non serving axons.
     if not metagraph.axons[uid].is_serving:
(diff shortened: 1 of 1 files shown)


gittensor/utils/uids.py
```diff
@@ -25,14 +25,26 @@ def check_uid_availability(metagraph: "bt.metagraph.Metagraph", uid: int, vpermi
...
+
+    Notes:
+        - ``exclude`` may be used to skip specific UIDs.
+        - UID ``0`` is always included (subnet requirement), even if it
+          would otherwise be filtered out.
     """
-    # Get all available miner UIDs, excluding specified ones
-    available_miner_uids = {uid for uid in range(self.metagraph.n.item()) if uid not in exclude}
+    metagraph: "bt.metagraph.Metagraph" = self.metagraph
+    vpermit_tao_limit: int = getattr(self.config.neuron, "vpermit_tao_limit", 4096)
```
anderdc: what if, for whatever reason, a miner had more than the vpermit tao limit?

I guess this would effectively make it so validators cannot dual mine, are you ok with this @LandynDev? I am fine with it. this is a good optimization overall so we're not querying dud UID, or vali UIDs

gittensor/utils/uids.py
```diff
@@ -25,14 +25,26 @@ def check_uid_availability(metagraph: "bt.metagraph.Metagraph", uid: int, vpermi
 
 
 def get_all_uids(self, exclude: List[int] = []) -> set[int]:
-    """Returns all uids from the metagraph, excluding specified ones.
```
anderdc: please use our typical function docstring/comments pattern
i.e. brief description, args, return

no need for 'notes',

PR #94, opened 2025-12-15
Fix case-insensitive repository name matching in master_repositories check

James-4u:
## Summary
Updated the repository name comparison logic to perform case-insensitive matching when checking if a repository exists in `master_repositories`.

## Changes
- Modified line 351 in `github_api_tools.py` to convert both `repository_full_name` and all keys in `master_repositories` to lowercase before comparison
- This ensures that repository names are matched case-insensitively, preventing false negatives when repository names have different casing

## Why This Change?
GitHub repository names are case-insensitive, but the comparison was previously case-sensitive. This could cause valid repositories to be incorrectly filtered out if there was a casing mismatch between the repository name from the API and the key stored in `master_repositories.json`.

## Example
image
image

## Testing
- Verified that repositories are correctly identified regardless of case differences
- Ensures backward compatibility with existing lowercase keys in `master_repositories.json`

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -496,14 +496,7 @@ def get_user_merged_prs_graphql(
                 if not pr_raw['mergedAt']:
                     continue
 
-                # Filter by master_repositories
-                if repository_full_name.lower() not in [repo_full_name.lower() for repo_full_name in master_repositories.keys()]:
-                    bt.logging.debug(
-                        f"Skipping PR #{pr_raw['number']} in {repository_full_name} - ineligible repo"
-                    )
-                    continue
-
-                # Parse merge date and filter by time window
+                # Parse merge date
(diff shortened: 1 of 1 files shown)


gittensor/utils/github_api_tools.py
```diff
@@ -495,7 +496,14 @@ def get_user_merged_prs_graphql(
...
 
-                # Parse merge date
+                # Filter by master_repositories
+                if repository_full_name.lower() not in [repo_full_name.lower() for repo_full_name in master_repositories.keys()]:
+                    bt.logging.debug(
+                        f"Skipping PR #{pr_raw['number']} in {repository_full_name} - ineligible repo"
+                    )
+                    continue
+
+                # Parse merge date and filter by time window
```
LandynDev: We do this check in the _should_skip_merged_pr function right below that gets called, so this is redundant. It's the first step in that function below

LandynDev (approved):

PR #109, opened 2025-12-18
Fix: remove duplicate function and update timezone to Chicago TZ

eureka0928:
## Move the constant

  `VALIDATOR_WAIT` and `VALIDATOR_STEPS_INTERVAL` in config.py are validator-specific runtime timing constants used only by validator code.
  Problem: `MERGED_PR_LOOKBACK_DAYS` is a scoring/scoping constant used by gittensor/utils/github_api_tools.py (non-validator code), creating an incorrect dependency.
  Changed: Move `MERGED_PR_LOOKBACK_DAYS` to constants.py alongside similar scoring/scoping constants.


LandynDev (changes requested):

leonace924:
@LandynDev pushed the updates and let me know if this looks good to you

PR #111, opened 2025-12-19
fix: validate GitHub user data before caching

Achieve3318:
Prevent caching incomplete user data that's missing required fields (id, created_at). This prevents crashes when later code tries to access these fields.

Changes:
- Added validation before caching user data in get_github_user()
- Returns data without caching if required fields are missing
- Added descriptive warning log for debugging


anderdc:
have you observed any times when the github api has returned incomplete responses? this seems like an unnecessary edge case solution

PR #138, opened 2026-01-20
Fix null pointer exception in open PR collateral calculation

frozenflux2:
## Summary
Adds a null guard for `repository_tier_configuration` in `calculate_open_pr_collateral_score()` to prevent `AttributeError` crashes when processing open PRs without tier configuration.

## Problem
The `calculate_open_pr_collateral_score()` function accessed `pr.repository_tier_configuration.open_pr_collateral_percentage` without checking if `repository_tier_configuration` was `None`. Since `repository_tier_configuration` is an optional field (defined as `Optional[TierConfig] = None` in `classes.py`), this would cause an `AttributeError` when processing open PRs from repositories that don't have a tier configuration assigned.

## Solution
- Added a null check at the start of the function
- Returns `0.0` collateral score if tier config is missing (safe default - no penalty applied)
- Logs a warning with the PR number and repository name for debugging purposes

## Testing
- Manually verified the fix handles `None` case gracefully
- Existing tests should continue to pass as the change is additive


frozenflux2:
@LandynDev @anderdc 
This PR is ready for review.
It fixes a null pointer exception in open PR collateral calculation.
Thanks!

PR #257, opened 2026-03-02
fix: exclude github organization member role

claytonlin1110:
## Summary
Currently there are some repositories which has temporarily creating a PR from the collaborators. Thus they invite to their organization who is willing to contribute to their repository.
We need to remove "Member" role so that PRs for such kind of repositories could still be in tier calculation so not to skip them.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)


claytonlin1110:
@anderdc Please review

PR #349, opened 2026-04-06
Revamp repo list

anderdc:
- pruning


MkDev11 (commented):
could we make the `sbt/sbt` repo active and increase its weight?

PR #365, opened 2026-04-10
fix: refresh stale repo weights on cached evaluations (#364)

### LINKED ISSUE #364: Incorrect repository weight shown for some PR pages
## Description
The repository weight shown on the PR details page is inconsistent between two PRs.

PR B shows the correct repository weight based on the updated value, but PR A still shows an incorrect value and does not appear to reflect the latest repository weight.

## Steps to Reproduce
1. Open PR A: https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=231
2. Open PR B: https://gittensor.io/miners/pr?repo=janhq%2Fjan&number=7744
3. Compare the repository weight displayed on each PR page.
But for the A is not correct as expected updated value

## Expected Behavior
Both PR pages should show the correct repository weight based on the latest updated value.

## Actual Behavior
PR B displays the correct updated repository weight.

jonathanchang31:
## Summary
- Cached miner evaluations retained `repo_weight_multiplier` from the round they were originally scored. When `master_repositories.json` was updated between rounds, cached evaluations used stale weights for scoring, causing incorrect repository weights.
- Added `_refresh_cached_repo_weights()` in `reward.py` that updates `repo_weight_multiplier` on all PRs in cached evaluations using the current `master_repositories` before `finalize_miner_scores()` runs.
- DB storage still skips cached UIDs (to avoid overwriting file patches with NULL from lightweight cache copies) and self-corrects on the next successful evaluation.

Fixes #364

## Test plan
- [x] 10 new unit tests covering: merged/open/closed PRs, multi-repo scenarios, rounding, fallback default for removed repos, non-cached UIDs untouched, mixed cached+fresh evaluations
- [x] Full test suite (282 tests) — all pass, no regressions
- [x] Ruff lint — all checks passed
- [x] Ruff format — all files formatted
- [x] Pyright type check — 0 errors, 0 warnings

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -70,10 +70,8 @@ async def forward(self: 'Validator') -> None:
         # 3. Score issue discovery
         issue_rewards = await issue_discovery(miner_evaluations, master_repositories, miner_uids)
 
-        # 4. Store all evaluations to DB (includes issue discovery fields).
-        # Cached evaluations are no longer skipped because their repo weights
-        # have been refreshed to current values and must be persisted.
-        await self.bulk_store_evaluation(miner_evaluations)
+        # 4. Store all evaluations to DB (includes issue discovery fields)
+        await self.bulk_store_evaluation(miner_evaluations, skip_uids=cached_uids)
 
         # 5. Blend 4 emission pools into final rewards
         rewards = blend_emission_pools(oss_rewards, issue_rewards, miner_uids)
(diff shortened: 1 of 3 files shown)


anderdc:
if master repositories is updated mid round, it's from a push from test -> main, which causes the validator to restart and thus all cached evaluations are cleared/reset. 

The only case this happens is if a validator is trying to make their own repository weights, which would then cause them to lose vtrust, so this scenario you're fixing doesn't happen in production.

PR #389, opened 2026-04-11
Enhance `gitt` help CLI output in small terminal size

### LINKED ISSUE #388: Keep the plain text output for narrow terminal size
## Description

gitt help CLI uses modern TUI - box + column style for the command output.
but this has one weakness, it's just broken when terminal has a small size.
So the output is invisible competely.
It's the best way to keep plain text output in a small certain terminal size.

Image

Image

## Steps to Reproduce

1. Clone gittensor
2. Run `gitt --help`, `gitt admin --help` and etc
3. Observe that TUI is broken completely so it's invisible.

## Expected Behavior

It should use plain text output in small terminal size.

## Actual Behavior

gitt help CLI is broken completely when terminal has a small certain size.

## Environment

- Browser: Terminal
- OS: Linux

dataCenter430:
## Summary

The `gitt --help` output uses Rich-powered panels and ratio-based columns which break visually on narrow or split-pane terminals. This PR adds a plain-text fallback that activates automatically when the terminal is too narrow to render the box layout cleanly.

## Root Cause

Two bugs combined to make the existing width-check ineffective:

1. **`ctx.terminal_width` is always `None`** — Click does not populate this field on its own. The `or 120` fallback meant the console was always created at 120 columns regardless of the real terminal width, so the width guard never fired.
2. **Threshold was 80** — even if detection had worked, 80 cols is the classic standard terminal width. The 4-column ratio layout (`4:1:1:7`) is already cramped at that size.

### Width detection fix
Replaced `ctx.terminal_width or 120` with a `_terminal_width(ctx)` helper that falls back to `shutil.get_terminal_size()`, which correctly reads the OS terminal dimensions.

### Threshold raised to 100
Terminals under 100 columns get plain text. The 4-column options panel needs real room; at 80 cols (standard terminal) it was always visually broken.

### Plain-text rendering path
Three new helpers render help output without any Rich boxes when width is below the threshold:

| Helper | Output |
|---|---|
| `_plain_usage` | Bare `Usage:` line |
| `_plain_options` | `Options:` section, two-column aligned |
| `_plain_section` | `Commands:` section, two-column aligned |

`_plain_options` also handles long flag names gracefully: if a label would push the description off-screen, it uses a **stacked layout** (flag on its own line, description indented below) instead of an inline one. The label column is capped at half the usable terminal width.

## Behaviour

| Terminal width | Output |
|---|---|
| ≥ 100 cols | Rich panels + boxes (unchanged) |
| 60 – 99 cols | Plain text, inline where it fits |
| ≤ 59 cols | Plain text, stacked for long flags |

## Testing

Verified manually at widths 40, 60, 79, 80, 99, 100, and 120 columns with both `StyledCommand` and `StyledGroup` help output. No regressions on the wide-terminal path.

## Screenshot

Before:

image

After:

image

image

### DIFF
--- a/gittensor/cli/issue_commands/help.py
+++ b/gittensor/cli/issue_commands/help.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 
 from inspect import cleandoc
 
+import shutil
+
 import click
 from rich import box
 from rich.console import Console
@@ -14,6 +16,18 @@ from rich.markup import escape
 from rich.padding import Padding
 from rich.panel import Panel
 from rich.table import Table
+
+# Terminal width below which plain-text output is used instead of Rich boxes.
+# 100 cols gives the 4-column options panel enough room to breathe; anything
+# narrower (including the classic 80-col standard terminal) gets plain text.
+_PLAIN_WIDTH = 100
+
+
+def _terminal_width(ctx: click.Context) -> int:
+    """Reliably resolve terminal width from Click context or the OS."""
+    if ctx.terminal_width:
+        return ctx.terminal_width
+    return shutil.get_terminal_size(fallback=(120, 24)).columns
 
 
 def _single_paragraph(text: str) -> str:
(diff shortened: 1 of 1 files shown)


dataCenter430:
Hi, @anderdc 
Would you review this PR and share me your thoughts, please?

anderdc:
why would someone not just rerun the command and expand their terminal size? It seems you're adding unnecessary complexity for an issue that doesn't really matter?

dataCenter430:
Hi, @anderdc 
Thanks for your consideration.
Fair point, but resizing isn't always possible:
- SSH sessions and container shells (often fixed at 80 cols)
- CI logs (GitHub Actions captures output at fixed width)
- Split panes in tmux/VSCode — can't expand without shrinking another
- Mobile terminals (Termux, Blink)

For a new user, broken box rendering on first `gitt --help` is a poor first impression.

The change is ~50 lines behind a single `width < 100` guard — no new deps, no state, wide-terminal path unchanged. Tools like `git`, `kubectl`, `cargo`, and the `af` CLI in the screenshot all degrade gracefully on narrow terminals; we were the outlier.

Happy to tune the threshold if you'd prefer.

wdeveloper16:
Good, I also think so, @dataCenter430 
Thanks for your PR.

syndrome4002-leo:
Good enhancement, @dataCenter430 
It's so helpful for many cases.
cc: @anderdc

anderdc:
suspicious engagement from accounts that barely have activity, and I don't like the PR that much, closing

PR #406, opened 2026-04-13
fix: ignore stale closed PRs in lookback window

### LINKED ISSUE #277: Dead weight PRs
### The problem

As a miner I encounter situation when you do PRs in repositories, but maintainers ignore them.

There are a lot of cases when your PR could be ignored and will be just a dead weight for you. Besides cases when project is obviously inactive and you should pay attention to not contribute there, there are many other reasons why your PR could be ignored, everything depends on maintainer. Sometimes they just ignore contributions from non-members, sometimes review-loop takes too long (I mean months or even years), sometimes they are half-active and merge only simple fixes.

You can say that I can ask before contributing, but it doesn't guarantee anything. I mean, you should ask, of course, but it doesn't protect you from being ignored after you open PR. You can also say that I just can skip this project if it seems suspicious. Well, yes, but you never know whether you would be ignored in other project too.

All what I want to say that current system restricts you too much:
- You're not protected against ignoring. I see a lot of great projects, but then I look at their average PR lifetime, feedback loop, maintainers involvement and so on, and skip them, because of a risk that merging a PR would take forever. I would be happy to try contributing into some projects, but such risk often stops me.
- If you already opened a PR in such project, it will hang there for months and will take score from each consequence PRs to other projects. Closing it will punish you with credibility penalty.
- At the end, it looks not fair to enforce contributors to face credibility loss because they opened a PR in a wrong project.

### What I suggest

There are quite a lot of possible ways to improve this system and it requires quite a big discussion as I think. But the very first thing comes to my mind is the following approach: 

- When opening a PR, it's "closing" credibility penalty factor starts declining with time (from 1 to 0). For example, after a month of "opened" state, I can safely close PR without credibility loss. If it was closed after a week, then let's say its' penalty coefficient is something around `0.7`.
- We can also consider additional penalty if closed by a maintainer or/and the opposite, consider slightly decreased penalty if closed by PR's owner.
- `closed_prs` variable in the credibility formula (`credibility = merged_prs / (merged_prs + closed_prs)`) now not just a plain count of closed PRs, but sum of their's coefficients.

Something like that looks fairer than what we have today. It will negate the fear contributing to projects without previous contributions and enforce people to try more instead of sticking to the well-known "tested" projects.

MkDev11:
## Summary

Fixes closed-PR lookback filtering so stale PRs are not counted against miner credibility.

Previously, CLOSED PRs were counted when `closedAt` was within the lookback window, even if the PR had been created before the window. This could penalize miners for closing old backlog PRs.

## Related Issues

Fixes #277 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [ ] Tests added/updated
- [x] Manually tested

Validated behavior with existing test suite/lint/type checks in local dev environment.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -881,7 +881,19 @@ def try_add_open_or_closed_pr(
             bt.logging.warning(f'PR #{pr_raw["number"]} is CLOSED but missing closedAt timestamp.')
             return
 
+        created_at = pr_raw.get('createdAt')
+        if not created_at:
+            bt.logging.warning(f'PR #{pr_raw["number"]} is CLOSED but missing createdAt timestamp.')
+            return
+
         closed_dt = datetime.fromisoformat(closed_at.rstrip('Z')).replace(tzinfo=timezone.utc)
+        created_dt = datetime.fromisoformat(created_at.rstrip('Z')).replace(tzinfo=timezone.utc)
+
+        # Ignore stale PRs that were created before the scoring lookback window.
+        # This allows users to close old PRs without receiving a fresh credibility penalty.
+        if created_dt < lookback_date_filter:
+            return
+
         if closed_dt >= lookback_date_filter:
             miner_eval.add_closed_pull_request(pr_raw)
 


anderdc (approved):

anderdc:
please link issue #277 to this before I merge

PR #408, opened 2026-04-14
refactor: extract shared normalization logic from duplicate normalize functions

bittoby:
### Summary
- Create a generic `_normalize_scores()` function in `oss_contributions/normalize.py` that accepts a score getter and label
- Rewrite both `normalize_rewards_linear()` and `normalize_issue_discovery_rewards()` as thin wrappers around it

### Why
Both normalization functions had the same algorithm: extract scores from evaluations, sum them, divide each by the total. The only difference was which field they read (`total_score` vs `issue_discovery_score`) and the log messages.

### Test plan
- [x] All 301 tests pass
- [x] Pyright clean
- [x] Ruff lint and format clean

### DIFF
--- a/gittensor/validator/issue_discovery/normalize.py
+++ b/gittensor/validator/issue_discovery/normalize.py
@@ -3,32 +3,10 @@
 
 from typing import Dict
 
-import bittensor as bt
-
 from gittensor.classes import MinerEvaluation
+from gittensor.validator.oss_contributions.normalize import _normalize_scores
 
 
 def normalize_issue_discovery_rewards(miner_evaluations: Dict[int, MinerEvaluation]) -> Dict[int, float]:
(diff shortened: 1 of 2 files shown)


bittoby:
@anderdc why you closed this PR? It is not different from #409 and #407.
pls let me know. thanks

anderdc:
#407 and #409 remove genuinely duplicated code — the same function copy-pasted in two places. #408 is different: the two normalize functions aren't duplicates, they operate on different fields and have different log messages. They share a simple pattern (sum-and-divide), but abstracting that into a generic function with callbacks adds indirection without real benefit. If the normalization logic for one scoring type needs to diverge later, having them as separate straightforward functions is easier to work with than unwinding a shared abstraction.

PR #412, opened 2026-04-14
Move duplicated ISO timestamp parser to shared datetime_utils

statxc:
### Description
- Extract `_parse_iso()` from `validator/issue_discovery/repo_scan.py` into `validator/utils/datetime_utils.py` as `parse_iso()`
- Import the shared function in `repo_scan.py`

### DIFF
--- a/gittensor/validator/issue_discovery/repo_scan.py
+++ b/gittensor/validator/issue_discovery/repo_scan.py
@@ -18,6 +18,7 @@ import bittensor as bt
 import requests
 
 from gittensor.classes import Issue, MinerEvaluation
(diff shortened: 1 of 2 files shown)


anderdc:
Closing this one out:

- **`_parse_iso` isn't duplicated anywhere**, it only exists in `repo_scan.py`
- `parse_github_timestamp_to_cst` in `datetime_utils.py` is a different function (converts to CST, not just parsing)
- Moving a single-use private helper to a shared module is premature, adds an import hop for no benefit
- If we need `parse_iso` elsewhere later, that's the time to extract it

PR #431, opened 2026-04-14
feat: Shell completion command

seroperson:
This PR introduces the `completion` command for enabling shell completion. With click [it's available out-of-box](https://click.palletsprojects.com/en/stable/shell-completion/), just needed a handy command to use it easily. Tools like `gh`, `pip`, `kubectl` etc have such `completion` command too, so it's a well-known adopted pattern.

```sh
eval "$(gitt completion bash)" # bash
eval "$(gitt completion zsh)"  # zsh
gitt completion fish | source  # fish
```

https://github.com/user-attachments/assets/13dd2e8b-ce52-461b-9892-bf4e7f0ba1b5


anderdc:
curious, so I don't need to do anything to set this up? just install gitt like normal and it should just autocomplete?

seroperson:
> curious, so I don't need to do anything to set this up? just install gitt like normal and it should just autocomplete?

If you're an end user and you just have the `gitt` in your `$PATH`, you only need a one-time setup by adding `eval "$(gitt completion zsh)"` to your `.zshrc`. Or you may want add this to your `.venv/bin/activate` like `echo 'eval "$(gitt completion)"' >> .venv/bin/activate`, if you work with `gitt` only in virtual env.

Various setup scenarios depend on how exactly `gitt` is installed, but it's usually an end user's and/or setup scripts responsibility, the tool itself just provides a `completion` command.

anderdc (approved):

PR #433, opened 2026-04-14
refactor: extract duplicate contract client init to _make_contract_client helper

dataCenter430:
The same 3-line block — `bt.Wallet`, `bt.Subtensor`, `IssueCompetitionContractClient(...)` — was copy-pasted identically across 6 admin commands and 2 vote commands (8 sites in total). Extracting it as `_make_contract_client()` in `helpers.py` gives a single source of truth.

The lazy imports (`bittensor`, `IssueCompetitionContractClient`) are preserved inside the helper, so `ImportError` still propagates to the existing `except ImportError` handlers in each caller unchanged.

- Pyright clean
- Ruff lint and format clean


dataCenter430:
Hi, @anderdc 
`admin.py` has 6 commands and `vote.py` has 2 commands that all start with the same 8-line block — create a wallet, a subtensor, and an `IssueCompetitionContractClient.` 
This PR extracts that block into `_make_contract_client()` in `issue_commands/helpers.py` and replaces all 8 call sites with a single line.
Really hope your review with this.

anderdc (approved):
Clean mechanical refactor. All 8 call sites updated, lazy imports correctly preserved, and the two sites in `mutations.py` that look similar were reasonably excluded (extra steps that don't fit the helper signature).

LGTM

PR #436, opened 2026-04-15
Refactor/deduplicate config path constants

wdeveloper16:
## Summary

`GITTENSOR_DIR` and `CONFIG_FILE` were defined identically in two places:
- `cli/main.py:28-29`
- `cli/issue_commands/helpers.py:48-49`

Moves the canonical definition to `gittensor/constants.py`. Both files now import from there.
Also removes the now-unused `from pathlib import Path` import from `helpers.py`.

> `constants.py` is used as the shared home to avoid a circular import -
> `main.py` already imports from `issue_commands/__init__.py` → `helpers.py`.

## Test plan
- [x] `pytest tests/cli/ tests/utils/`, no new failures

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -15,7 +15,6 @@ import urllib.error
 import urllib.request
 from contextlib import nullcontext
 from decimal import Decimal, InvalidOperation
-from pathlib import Path
 from typing import Any, Callable, ContextManager, Dict, List, Optional, Tuple, TypeVar
 
 import click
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
just move it and export it from helpers, doesn't need to be moved all the way up to `constants.py`.

Or just keep it in cli/main.py and have helper import it from there. 

Either way these are just for the CLI and not needed across the codebase

wdeveloper16:
Hi, @anderdc 
Good point! these are CLI-only and don't belong in `constants.py.`

One thing worth noting: option 2 (keep in main.py, import from there in helpers.py) would create a circular import - `main.py` already imports from `issue_commands/__init__.py,` which imports helpers.py, so importing back from main.py would close the loop.

Option 1 (define in `helpers.py,` have main.py import from there) doesn't have that issue. 
Happy to update the PR to do that instead.

anderdc (dismissed):
I see what you mean, we'll just leave it as is, simple dedup

wdeveloper16:
And sorry for my CI failure.
I will fix it right now.

wdeveloper16:
Hi, @anderdc 
Would you please review it again? 🙏 
I'm sure CI will be passed.

anderdc (approved):

PR #445, opened 2026-04-15
refactor: extract duplicate bulk upsert logic to _bulk_upsert helper

bittoby:
## Summary

Four methods in `repository.py` repeated the same 12–18 line block: open a cursor, call execute_values, commit on success, rollback on failure. Extracted that into a single _bulk_upsert() helper. Each method now delegates to it in one line. No behavior change.

### Changes

- `repository.py` - added `_bulk_upsert(query, values, error_context)` to `BaseRepository`, replaced all 4 duplicated blocks with calls to it

### Checks

- Ruff lint: clean
- Ruff format: clean
- Pyright: 0 errors, 0 warnings
- Pytest: 300 passed, 0 failed

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -51,6 +51,23 @@ class BaseRepository:
         finally:
             cursor.close()
 
+    def _bulk_upsert(self, query: str, values: list, error_context: str) -> int:
+        """Execute a bulk upsert via ``psycopg2.extras.execute_values``.
+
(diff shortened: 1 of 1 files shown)


anderdc:
Closing this — we're building a GitHub mirror that will replace the validator's direct PR/issue/file-change storage entirely. Three of the four methods this deduplicates will be removed once the mirror is serving data, so cleaning them up now would just be churn on code that's going away soon.

PR #446, opened 2026-04-15
refactor: refactor issue competition contract storage decoding

claytonlin1110:
## Summary
Refactored duplicate issue-competition contract storage parsing into a shared storage_utils module and switched both the validator contract client and CLI helpers to use it. Added fixture-based tests to lock in lazy key generation and byte-level decode behavior (packed storage and issue entries).

This cleans up duplicate contract storage parsing code that existed in both the CLI helpers and the issue competition contract client.

I pulled the shared child-storage key + decode logic into a new module:
- gittensor/validator/issue_competitions/storage_utils.py

Then updated both callers to use it:

- gittensor/validator/issue_competitions/contract_client.py
- gittensor/cli/issue_commands/helpers.py
No behavior changes intended here — this is mainly consolidation so the decode path lives in one place.

I also added fixture-style tests for the new utility module:

- tests/utils/test_issue_competitions_storage_utils.py

Those tests cover:

- lazy key generation
- packed storage decoding
- issue entry decoding (including compact-length string cases)
- child storage key extraction from different trie_id formats


claytonlin1110:
@anderdc Would you please review and give me feedback on this?

anderdc (commented):
## Review: Refactor Issue Competition Storage Decoding

**Verdict: Not junk, but needs changes before merge.**

This is a reasonable refactor that consolidates duplicate contract storage parsing from `contract_client.py` and `cli/issue_commands/helpers.py` into a shared `storage_utils.py` module. The dataclass approach (`PackedContractStorage`, `DecodedIssueStorage`) is cleaner than the raw dict returns. Tests are solid.

### Issues to fix

- **`codec.py` and `test_issue_codec.py` are not removed.** The PR replaces all imports from `codec.py` but leaves the old module and its tests in the tree. These should be deleted since `storage_utils.py` fully supersedes them. Without this, you have two modules doing the same thing.
- **`_compute_ink5_lazy_key` wrapper in `helpers.py` is pointless.** The PR adds a private function `_compute_ink5_lazy_key` (line ~124 in the diff) that just calls `compute_ink5_lazy_key` from `storage_utils`. It's a one-line passthrough. Either call the shared function directly at the call sites or remove the wrapper entirely.
- **Broad `except Exception` in `decode_issue_from_storage`.** The function catches all exceptions and returns `None`, which will silently swallow real bugs (e.g. a changed struct layout). At minimum, log the exception or let `struct.error` propagate so callers know **why** a decode failed.
- **Typo in test fake.** `_FakeSubstrate.rpc_request` raises `AssertionError` (missing an 'r') instead of `AssertionError`. This won't cause test failures since that branch isn't hit, but it should be `AssertionError`.

### Minor nits

- The `compute_ink5_lazy_key` method added to the `IssueCompetitionContractClient` class is only a passthrough to the module-level function. Consider whether it's needed on the class at all, or if callers should just use the import directly.
- `_read_contract_packed_storage` in `helpers.py` now wraps everything in a single broad `try/except` where the original had more granular error paths. The verbose debug messages inside the try block reference data that may not exist yet if an earlier step failed.

claytonlin1110:
@anderdc Updated, please review again

anderdc (approved):

PR #447, opened 2026-04-15
fix: assign evaluation_timestamp before stale cleanup

### LINKED ISSUE #378: fix: evaluation_timestamp is never assigned, stale cleanup SQL always evaluates to NULL
## Bug Description

`MinerEvaluation.evaluation_timestamp` is declared in `gittensor/classes.py` as:

```python
evaluation_timestamp: Optional[datetime] = None
```

It is **never assigned** anywhere in the codebase. However, in `gittensor/validator/storage/repository.py`, `cleanup_stale_miner_data()` uses `evaluation.evaluation_timestamp` directly as a SQL parameter:

```python
eval_params = params + (evaluation.evaluation_timestamp,)
# ...
self.execute_command(CLEANUP_STALE_MINER_EVALUATIONS, eval_params)
```

The SQL queries (`CLEANUP_STALE_MINER_EVALUATIONS` and `CLEANUP_STALE_MINER_EVALUATIONS_BY_HOTKEY`) both contain the clause:

```sql
AND created_at <= %s
```

In SQL, `ANY_VALUE <= NULL` evaluates to `NULL` (not `TRUE`), so the entire `WHERE` predicate is never satisfied. The `DELETE` statements silently delete **zero rows** on every call.

## Impact

Stale `miner_evaluations` records — those belonging to a miner that has re-registered under a new `uid`/`hotkey` or re-linked to a new `github_id` — are **never cleaned up**. These records accumulate indefinitely, causing:

- Incorrect historical score data associated with the wrong miner identity
- Unbounded growth of the `miner_evaluations` table

## Fix

Assign `datetime.now(timezone.utc)` to `evaluation.evaluation_timestamp` at the start of `cleanup_stale_miner_data()`, before it is used to build the SQL parameters.

```python
def cleanup_stale_miner_data(self, evaluation: MinerEvaluation) -> None:
    if not evaluation.github_id or evaluation.github_id == '0':
        return

    evaluation.evaluation_timestamp = datetime.now(timezone.utc)  # <-- assign here

    params = (evaluation.github_id, evaluation.uid, evaluation.hotkey)
    eval_params = params + (evaluation.evaluation_timestamp,)
    ...
```

## Files Affected

- `gittensor/validator/storage/repository.py` — fix location
- `gittensor/classes.py` — `MinerEvaluation.evaluation_timestamp` field (line ~312)

plind-junior:
Fixes #378 
All the previous PRs for this issue closed and didn't reopened so I'm opening a new PR with timestamp fix

cc @anderdc

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -8,6 +8,7 @@ and miner evaluations.
 
 import logging
 from contextlib import contextmanager
+from datetime import datetime, timezone
 from typing import List, TypeVar
 
 import numpy as np
@@ -122,6 +123,8 @@ class Repository(BaseRepository):
         """
         if not evaluation.github_id or evaluation.github_id == '0':
             return
+
+        evaluation.evaluation_timestamp = datetime.now(timezone.utc)
 
         params = (evaluation.github_id, evaluation.uid, evaluation.hotkey)
         eval_params = params + (evaluation.evaluation_timestamp,)
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -8,7 +8,6 @@ and miner evaluations.
 
 import logging
 from contextlib import contextmanager
-from datetime import datetime, timezone
 from typing import List, TypeVar
 
 import numpy as np
@@ -122,8 +121,6 @@ class Repository(BaseRepository):
         """
         if not evaluation.github_id or evaluation.github_id == '0':
             return
-
-        evaluation.evaluation_timestamp = datetime.now(timezone.utc)
 
         params = (evaluation.github_id, evaluation.uid, evaluation.hotkey)
         eval_params = params + (evaluation.evaluation_timestamp,)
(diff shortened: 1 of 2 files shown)


plind-junior:
@anderdc I saw your comment on the closed PRs and opening duplicated PR as nobody reopens it

anderdc (commented):
_(superseded by changes-requested review below)_

anderdc (changes requested):
- **`evaluation_timestamp` is set inside `cleanup_stale_miner_data`, which is a hidden side effect.** A cleanup helper shouldn't silently mutate the evaluation object.
- **Move the timestamp assignment to the top of `store_evaluation_data` in `storage.py`** (before the repo calls). That way the setup is visible at the entry point, not buried in a sub-call.
- The fix itself is correct, `evaluation_timestamp` defaults to `None` and is never set anywhere else, so this solves the bug. Just needs to live in the right place.

plind-junior:
@anderdc Thanks for the review. I made an update accordingly

anderdc (approved):

PR #452, opened 2026-04-15
Fix/issue discovery closed at window

plind-junior:
## Summary
- `_fetch_closed_issues` in `gittensor/validator/issue_discovery/repo_scan.py` passes `since=<PR_LOOKBACK_DAYS ago>` to `GET /repos/{repo}/issues`, expecting a `closed_at` lookback
- GitHub REST actually filters `since` by `updated_at`, so ancient closed issues with a recent comment / label / reaction were flowing into Phase 2 repo scan as Case 2 / Case 3, polluting credibility scoring and wasting `REPO_SCAN_*_CAP` solver-lookup budget on stale threads
- Because the trigger depended on comment timing, validators could disagree on the same miner in the same round
- Fix: parse `since` once, then drop any response row with missing or out-of-window `closed_at`. If `since` cannot be parsed, fall back to raw results so the scan is never silently wiped

## Details
GitHub docs describe `since` as filtering by last-updated time, not closed time. REST has no server-side `closed_at` filter, so the fix is client-side after pagination.

### Reproduction
```bash
gh api "/repos/ggml-org/llama.cpp/issues?state=closed&since=$(date -u -d '35 days ago' +%Y-%m-%dT%H:%M:%SZ)&per_page=100" \
  --jq '.[] | select(.closed_at < "2025-01-01") | {number, closed_at, updated_at}'

### DIFF
--- a/tests/validator/test_repo_scan_closed_at_window.py
+++ b/tests/validator/test_repo_scan_closed_at_window.py
@@ -1,0 +1,108 @@
+"""Tests for `_scan_repo` closed_at lookback window enforcement.
+
+GitHub REST `/repos/{repo}/issues?since=...` filters by `updated_at`, not
+`closed_at`, so response pages can contain ancient issues that were merely
+re-touched (comment, label, reaction) within the window. `_scan_repo` must
+drop these client-side before spending solver-lookup budget on them.
+"""
+
+import asyncio
+from datetime import datetime, timedelta, timezone
+from unittest.mock import patch
+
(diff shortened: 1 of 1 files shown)


anderdc:
Closing this PR:

- **Incomplete** - adds tests but no actual fix to `repo_scan.py`, so the tests would fail against current code
- **Not necessary** - the GitHub mirror will replace direct GitHub PAT/REST calls for issue discovery, making this client-side workaround obsolete

PR #453, opened 2026-04-15
refactor: extract duplicated normalization logic into shared utility

MkDev11:
## Summary

`normalize_rewards_linear` (OSS contributions) and `normalize_issue_discovery_rewards` (issue discovery) implemented the same "normalize scores to sum to 1.0" algorithm independently. Extracted the shared math into `normalize_scores()` in `validator/utils/normalize.py`. Both callers now delegate to it while preserving their domain-specific field extraction and logging.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #454, opened 2026-04-15
refactor: extract shared normalize_scores utility from duplicate normalize functions

dataCenter430:
## Summary

`normalize_rewards_linear` (oss_contributions) and `normalize_issue_discovery_rewards` (issue_discovery) implement the same algorithm: build a uid→score map, check if total > 0, divide each score by total. The only difference is the field each pulls from (`evaluation.total_score` vs `evaluation.issue_discovery_score`).

This PR extracts the shared division logic into `gittensor/validator/utils/normalize.py` as `normalize_scores(scores: Dict[int, float]) -> Dict[int, float]` and has both functions delegate to it. All per-domain logic (field selection, logging) stays in each function.

## Test plan
- [ ] All 310 tests pass
- [ ] Pyright clean
- [ ] Ruff lint and format clean

### DIFF
--- a/gittensor/validator/issue_discovery/normalize.py
+++ b/gittensor/validator/issue_discovery/normalize.py
@@ -6,6 +6,7 @@ from typing import Dict
 import bittensor as bt
 
 from gittensor.classes import MinerEvaluation
+from gittensor.validator.utils.normalize import normalize_scores
 
 
 def normalize_issue_discovery_rewards(miner_evaluations: Dict[int, MinerEvaluation]) -> Dict[int, float]:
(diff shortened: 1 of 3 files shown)


anderdc:
- **Closing this as duplicate.** This is the same refactor already attempted in #408 and #453, both closed for the same reason.
- **The shared logic is a one-line dict comprehension** (`{uid: score / total ...}`). That's not enough duplication to justify a new module and file.
- Both normalize functions still keep all their own domain logic (field selection, logging, zero-count tracking). The only "shared" part is trivial division.

PR #459, opened 2026-04-15
refactor: extract duplicate _connect bittensor setup to shared helper

dataCenter430:
## Summary

- `check.py` and `post.py` both defined an identical inner `_connect()` closure that set up `bt.Wallet`, `bt.Subtensor`, `bt.Metagraph`, and `bt.Dendrite`
- Extracted to `_connect_bittensor(wallet_name, wallet_hotkey, ws_endpoint, netuid)` in `miner_commands/helpers.py`
- Both commands now import and call the shared helper, removing the duplicated closure and the redundant `import bittensor as bt` in each command

## Test plan

- [ ] `gitt miner check` connects and queries validators as before
- [ ] `gitt miner post` connects and broadcasts PAT as before
- [ ] Ruff lint/format clean
- [ ] Pyright passes
- [ ] Pytest passes


dataCenter430:
Hi, @anderdc 
Would you please review this PR?

anderdc (approved):

PR #460, opened 2026-04-15
refactor: deduplicate get_contract_address into shared utils

dataCenter430:
## Summary

- `validator/utils/issue_competitions.py` and `cli/issue_commands/helpers.py` both implemented the same `os.environ.get('CONTRACT_ADDRESS') or CONTRACT_ADDRESS` lookup independently
- Extracted the shared base to `gittensor/utils/utils.py` as `get_contract_address() -> str`
- `issue_competitions.py` removes its local definition and re-exports from `utils.utils`
- `helpers.py` drops the duplicated lookup and delegates to the shared function when no CLI override is provided

## Test plan

- [ ] `get_contract_address()` returns env var override when `CONTRACT_ADDRESS` is set
- [ ] Falls back to `constants.py` default when env var is absent
- [ ] CLI `--contract` flag override still takes priority
- [ ] Ruff lint/format clean
- [ ] Pyright passes
- [ ] Pytest passes


gittensor/utils/utils.py
```diff
@@ -2,9 +2,17 @@
...
 from typing import Dict
 
 
 def parse_repo_name(repo_data: Dict):
     """Normalizes and converts repository name from dict"""
     return f'{repo_data["owner"]["login"]}/{repo_data["name"]}'.lower()
+
+
+def get_contract_address() -> str:
+    """Get contract address. Override via CONTRACT_ADDRESS env var for dev/testing."""
```
anderdc: please don't delete the original function doc string, needs 

```
    Returns:
        Contract address string (env var override or constants.py default)
```

anderdc (changes requested):
One small fix needed: restore the full docstring on `get_contract_address()` in `utils.py` (the `Returns:` section was dropped during the move). Otherwise LGTM.

gittensor/utils/utils.py
dataCenter430: Resolved

anderdc (approved):

PR #465, opened 2026-04-16
test → main sync

anderdc:
## Summary

Sync test branch to main.

## Test plan
- [ ] All tests passing on test branch


seroperson:
> refactor (1.05x)

@anderdc shouldn't it be heavily penalized instead? To discourage nearly risk-free refactoring PRs which can be spammed indefinitely to any repo: "duplicated code", "make one thing to become other thing" etc etc etc. 

They require almost zero efforts to implement, provide zero end-user benefits, and it's impossible to compete with them, if not spam refactoring too. Implementing features or something requires you to think and take a risk of being declined, while refactoring likely is not. 

Existing feature 0.45x gap, or even less 0.2x fix and 0.1x enh gaps are far not enough to compete with dozens of refactoring PRs.

Average top-N miner currently looks like:
image image

LandynDev:
Closing — reopening fresh after reconciliation PR #500 landed so mergeability check re-runs.

PR #472, opened 2026-04-16
refactor(cli): centralize Subtensor endpoint resolution

jsdevninja:
## Summary
Moves load_config, resolve_network, and shared config path constants into gittensor/cli/endpoint_resolution.py so issue and miner CLIs use one implementation for resolving the Subtensor WebSocket endpoint (--rpc-url -> --network -> ~/.gittensor/config.json -> default finney), including URL->network-name handling for explicit RPC URLs.

## Changes

- New: gittensor/cli/endpoint_resolution.py — CONFIG_FILE, GITTENSOR_DIR, load_config(), resolve_network(), resolve_ws_endpoint().
- Updated: gittensor/cli/issue_commands/helpers.py — imports and re-exports shared symbols; removes duplicated logic; drops unused Path import.
- Updated: gittensor/cli/miner_commands/helpers.py — _resolve_endpoint() delegates to resolve_ws_endpoint().

## Motivation
Avoid duplicated precedence rules between miner and issue commands so behavior cannot drift when config handling changes.

### DIFF
--- a/gittensor/cli/endpoint_resolution.py
+++ b/gittensor/cli/endpoint_resolution.py
@@ -1,0 +1,91 @@
+# The MIT License (MIT)
+# Copyright © 2025 Entrius
+
+"""Shared Subtensor WS endpoint resolution for CLI commands (issues, miner, etc.)."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any, Dict, Optional, Tuple
+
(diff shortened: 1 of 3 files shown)


anderdc:
**Closing -- the dedup doesn't justify the churn here.** It's +100/-99, essentially moving code between files without a meaningful reduction. The two implementations are close enough that drift isn't a real concern.

PR #469, opened 2026-04-16
refactor: remove dead resolve_netuid_from_contract and verify_miner_registration functions

dataCenter430:
## Summary

- `resolve_netuid_from_contract` and `verify_miner_registration` were defined in `gittensor/cli/issue_commands/helpers.py` but never called or imported anywhere outside their own file
- Neither is exported from `__init__.py`
- They appear to be a speculative registration-check feature that was never wired up to any CLI command

## Test plan

- [ ] Ruff lint/format clean
- [ ] Pyright passes
- [ ] Pytest passes

PR #476, opened 2026-04-16
refactor: remove dead verify_miner_registration and resolve_netuid_from_contract

t7929375-eng:
## Summary
Remove two dead functions from `gittensor/cli/issue_commands/helpers.py`:

- **`resolve_netuid_from_contract`** (line 290) — never called from any module. Its only caller was `verify_miner_registration` which is also dead.
- **`verify_miner_registration`** (line 302) — never imported or called anywhere in the codebase. Grep confirms zero references outside the definition.

Both functions were likely part of an earlier CLI workflow that has since been replaced.

## Related Issues
N/A (code cleanup)

## Type of Change

## Testing
- [x] `ruff check` — 0 errors
- [x] `pyright` — 0 errors

## Checklist
- [x] Self-review completed

cc @anderdc @landyndev

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -287,34 +287,6 @@ def print_issue_submission_table(
     console.print(f'Showing {len(pull_requests)} submissions{suffix}')
 
 
-def resolve_netuid_from_contract(ws_endpoint: str, contract_addr: str) -> Optional[int]:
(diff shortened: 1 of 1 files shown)


anderdc:
**Closing -- duplicate of #469**

This removes the same two functions (`resolve_netuid_from_contract` and `verify_miner_registration`) that dataCenter430 already submitted in #469, which was opened first.

PR #468, opened 2026-04-16
refactor: remove dead code across neurons and cli modules

dataCenter430:
## Summary

- **`Validator.store_evaluation`** (`neurons/validator.py`) — never called; `bulk_store_evaluation` is the only storage path used from `forward.py`
- **`resolve_netuid_from_contract` + `verify_miner_registration`** (`cli/issue_commands/helpers.py`) — never called or imported anywhere; speculative feature that was never wired up to any CLI command
- **`Repository` dataclass** (`classes.py`) — never imported anywhere in production code or tests; superseded by the separate `Repository` class in `gittensor/validator/storage/repository.py`
- **`MinerEvaluationCache.invalidate`** (`classes.py`) — never called anywhere; cache is cleared via `clear()` or natural eviction in `get()`

## Test plan

- [ ] Ruff lint/format clean
- [ ] Pyright passes
- [ ] Pytest passes

PR #479, opened 2026-04-16
refactor: remove dead store_evaluation and unused WANDB_API_KEY

t7929375-eng:
## Summary
Remove two pieces of dead code from the validator module:

- **`Validator.store_evaluation`** (`neurons/validator.py`, line 142) — async method never called anywhere. The codebase uses `bulk_store_evaluation` (line 101, called from `forward.py:74`) instead. The only `store_evaluation` calls in the codebase target `self.db_storage.store_evaluation()` (a different class: `DatabaseStorage`), not this Validator method.

- **`WANDB_API_KEY`** (`gittensor/validator/utils/config.py`, line 10) — constant assigned from `os.getenv('WANDB_API_KEY')` but never imported or referenced anywhere else. The sibling constants `WANDB_PROJECT` and `WANDB_VALIDATOR_NAME` are both imported and used in `neurons/validator.py`, but `WANDB_API_KEY` is not. The wandb library reads the environment variable directly.

## Related Issues
N/A (code cleanup)

## Type of Change

## Testing
- [x] `ruff check` — 0 errors
- [x] `pyright` — 0 errors

## Checklist
- [x] Self-review completed

cc @anderdc @landyndev

### DIFF
--- a/gittensor/validator/utils/config.py
+++ b/gittensor/validator/utils/config.py
@@ -7,7 +7,6 @@ VALIDATOR_STEPS_INTERVAL = 120  # 2 hours, every time a scoring round happens
 
 # required env vars
 GITTENSOR_VALIDATOR_PAT = os.getenv('GITTENSOR_VALIDATOR_PAT')
-WANDB_API_KEY = os.getenv('WANDB_API_KEY')
 WANDB_PROJECT = os.getenv('WANDB_PROJECT', 'gittensor-validators')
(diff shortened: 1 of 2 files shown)


anderdc:
**Closing -- overlaps with #468**

The `store_evaluation` removal is already covered by dataCenter430's #468, which was opened first. The extra `WANDB_API_KEY` removal can be submitted separately if needed.

PR #495, opened 2026-04-16
refactor: extract shared validator shutdown and storage error logging

jsdevninja:
## Summary
- Extract duplicated background-thread stop logic in `BaseValidatorNeuron` into `_stop_background_thread()`.
- Extract duplicated partial DB storage failure logging in `Validator` into `_log_storage_partial_failure()`.

## Why
- Reduces repeated logic across validator lifecycle and DB storage paths.
- Makes future updates less error-prone by centralizing shared behavior.
- Improves readability.

## Files changed
- `neurons/base/validator.py`
- `neurons/validator.py`

### DIFF
--- a/neurons/base/validator.py
+++ b/neurons/base/validator.py
@@ -184,12 +184,7 @@ class BaseValidatorNeuron(BaseNeuron):
         """
         Stops the validator's operations that are running in the background thread.
         """
-        if self.is_running:
-            bt.logging.debug('Stopping validator in background thread.')
-            self.should_exit = True
-            self.thread.join(5)
-            self.is_running = False
-            bt.logging.debug('Stopped')
+        self._stop_background_thread()
 
     def __enter__(self):
         self.run_in_background_thread()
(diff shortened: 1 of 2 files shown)


jsdevninja:
@anderdc Please review when you have a chance to take a look at.

anderdc:
**Closing -- we're not accepting changes to BaseValidatorNeuron.**

Beyond that, neither extraction improves the code:

- **`_stop_background_thread()`** -- `stop_run_thread()` and `__exit__` aren't "duplicated logic." `stop_run_thread()` is the public API, `__exit__` is the context manager teardown. They both need to do the same thing, but extracting a private method just adds indirection for 4 lines of code. Worse, `stop_run_thread()` now delegates to a private method, making the call path harder to follow for no benefit.
- **`_log_storage_partial_failure()`** -- this extracts 3 lines of logging (a warning + a loop) into a static method. That's not a meaningful abstraction. The logging is already clear inline, and the "dedup" only saves 2 lines per call site while adding a method and a level of indirection.

Also note `store_evaluation` (one of the two call sites for `_log_storage_partial_failure`) is dead code that's being removed in #468.

PR #498, opened 2026-04-16
Extract require_valid_issue_id and require_valid_ss58 CLI helpers

### LINKED ISSUE #497: Extract require_valid_issue_id and require_valid_ss58 helpers to deduplicate CLI validation
## Summary

Extract two helper functions (`require_valid_issue_id` and `require_valid_ss58`) into `gittensor/cli/issue_commands/helpers.py` to replace 7 identical try/except/re-raise blocks across `admin.py` and `vote.py`.

## Motivation

The same validation-and-reraise pattern appears 7 times across CLI commands (3 for `validate_issue_id`, 4 for `validate_ss58_address`). Each block is identical: call the validator, catch `click.BadParameter`, re-raise as `click.ClickException`. Extracting these into one-liner helpers reduces boilerplate and keeps validation behavior consistent.

statxc:
## Summary

Extract `require_valid_issue_id` and `require_valid_ss58` helpers into `gittensor/cli/issue_commands/helpers.py`, replacing 7 identical try/except/re-raise blocks across `admin.py` and `vote.py`.

### DIFF
--- a/gittensor/cli/issue_commands/admin.py
+++ b/gittensor/cli/issue_commands/admin.py
@@ -26,8 +26,8 @@ from .helpers import (
     print_error,
     print_network_header,
     print_success,
-    validate_issue_id,
-    validate_ss58_address,
+    require_valid_issue_id,
+    require_valid_ss58,
     with_network_contract_options,
     with_wallet_options,
 )
@@ -62,10 +62,7 @@ def admin_cancel(issue_id: int, network: str, rpc_url: str, contract: str, walle
     """
     contract_addr, ws_endpoint, network_name = _resolve_contract_and_network(contract, network, rpc_url)
 
-    try:
-        validate_issue_id(issue_id)
-    except click.BadParameter as e:
-        raise click.ClickException(str(e))
+    require_valid_issue_id(issue_id)
 
     print_network_header(network_name, contract_addr)
 
@@ -120,10 +117,7 @@ def admin_payout(issue_id: int, network: str, rpc_url: str, contract: str, walle
     """
     contract_addr, ws_endpoint, network_name = _resolve_contract_and_network(contract, network, rpc_url)
 
-    try:
-        validate_issue_id(issue_id)
-    except click.BadParameter as e:
-        raise click.ClickException(str(e))
+    require_valid_issue_id(issue_id)
 
     print_network_header(network_name, contract_addr)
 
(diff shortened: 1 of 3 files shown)


anderdc:
`val_vote_solution` in vote.py has an outer `try` that catches `click.BadParameter`:

```python
try:
    validate_issue_id(issue_id)
    validate_ss58_address(solver_hotkey, ...)
    ...
except click.BadParameter:
    raise
except ValueError as e:
    raise click.BadParameter(str(e), param_hint='pr_number_or_url')
```

Swapping to `require_valid_issue_id` changes the exception type to `click.ClickException`, which isn't caught by `except click.BadParameter`. Revert that one call site — leave `validate_issue_id` since the outer `try` already provides the same wrap.

Other 6 call sites look good.

statxc:
@anderdc I've updated. Please check again

anderdc (changes requested):
Clean dedup. One small thing -- fill out the docstrings on `require_valid_issue_id` and `require_valid_ss58` to match the existing `validate_issue_id` / `validate_ss58_address` pattern in the same file (Args, Returns sections). They're right next to each other so they should be consistent.

statxc:
> Clean dedup. One small thing -- fill out the docstrings on `require_valid_issue_id` and `require_valid_ss58` to match the existing `validate_issue_id` / `validate_ss58_address` pattern in the same file (Args, Returns sections). They're right next to each other so they should be consistent.

Thanks! Just checked - `validate_issue_id` is a one-liner and `validate_ss58_address` is a short paragraph, neither has Args/Returns sections. The wrappers already match that local pattern, so I'd prefer to keep them as-is.

anderdc (changes requested):
The `val_vote_solution` fix and dedup are correct. One remaining thing on the docstrings:

Both wrappers have a `return` statement and a `-> int` / `-> str` return type, but the docstrings don't mention what they return. At minimum add a `Returns:` line. For example:

```python
def require_valid_issue_id(value: int, param_name: str = 'issue_id') -> int:
    """Validate an issue ID, raising ClickException on failure.

    Returns:
        The validated issue ID.
    """
```

Same for `require_valid_ss58` -- `Returns: The validated SS58 address string.`

This matches the level of detail in `validate_ss58_address` right above them, which has a multi-line docstring explaining what it does.

statxc:
@anderdc I've updated. Please review again.

anderdc (approved):

PR #504, opened 2026-04-16
refactor: deduplicate ImportError/Exception handler across issue_commands CLI

dataCenter430:
## Summary

The same 6-line `except ImportError / except Exception` block appeared 10 times across `admin.py` (×6), `vote.py` (×3), and `view.py` (×1).

Extracted into `_handle_command_error(e)` in `helpers.py` and replaced all occurrences with a one-liner call.


anderdc:
Refactor is valid — helper doesn't exist anywhere on `test` yet, so introducing it in `issue_commands/helpers.py` is the right move. Behavior-equivalent with the original 6-line pattern.

One gap: `mutations.py` has two more `ImportError` + `Exception` pairs (lines 243/247 and 386/390) that fit the same helper. Extend the PR to cover those and it wraps up the pattern cleanly in this module.

FYI the referenced issue was closed as invalid (it claimed the helper already existed).

dataCenter430:
Very good catch! I will got it

dataCenter430:
Hi, @anderdc 
I've just found two pairs from the codebase.
The two pairs in mutations.py look similar on the surface but both have custom `except Exception` logic that doesn't fit `_handle_command_error`:

- Lines 243–257: the `Exception` handler branches on `ContractReverted` to print specific failure reasons — silently replacing it would drop that behaviour.
- Lines 386–398: the `Exception` handler prints `type(e).__name__` and conditionally shows a full traceback when `--verbose` is passed.

The 10 occurrences already covered (admin.py ×6, vote.py ×3, view.py ×1) are the ones that are truly identical - generic `print_error(str(e)) + SystemExit(1)` with no branching. 
Happy to add a follow-up PR for mutations.py with a more targeted helper if that's preferred, but that felt out of scope for this one.

dataCenter430:
Hi, @anderdc 
Would you please consider the comment?

anderdc (approved):

PR #512, opened 2026-04-16
feat: Implementing pre-commit and pre-push git hooks

seroperson:
## Summary

Pre-commit hooks catch formatting and lint issues locally on every commit (takes <1s). Pyright and pytest run before push. Weight config files are validated against JSON Schemas (structure, types, value ranges) and extended Python checks (duplicates, case collisions, sorted keys, weight bounds, tree-sitter availability). CI uses a single pre-commit-based workflow that mirrors the local hook configuration exactly.

- Add pre-commit framework with hooks for ruff lint/format, trailing whitespace, end-of-file, YAML/JSON checks, mixed line endings, and large file detection
- Add pre-push hooks for pyright type checking and pytest
- Add JSON Schema validation for all three weight config files (`master_repositories.json`, `programming_languages.json`, `token_weights.json`)
- Add Python-based extended validation that catches issues schemas cannot express: duplicate JSON keys, case-insensitive repo name collisions, unsorted keys, out-of-range weights, and unavailable tree-sitter languages
- Consolidate CI lint and typecheck workflows into a single pre-commit-based workflow, removing the separate `typecheck.yml`
- Update `CONTRIBUTING.md` with development setup instructions. Manual step is required to enable hooks.

## Type of Change

- [x] New feature

https://github.com/user-attachments/assets/ffd90fbe-d356-46a1-9d19-f415e06a1185


seroperson:
The thing I also changed is deleting auto-fix + auto-commit for member's PRs. I can revert it if necessary.
Just it was a little bit hard to keep in line with introduced checks.

anderdc:
Core pre-commit + CI consolidation work is good. Two changes to drop before this can merge:

**Remove the `pretty-format-json` hook.** Existing tests already enforce format/validity on the weight configs, so auto-rewriting them buys nothing and produces ~520 lines of noise in this PR (and in every future weight PR).

**Drop everything touching the weight configs.** That means:
- Revert `programming_languages.json` and `token_weights.json` (they should be untouched)
- Remove all three files in `scripts/schemas/`
- Remove `scripts/validate_weights.py` and the `validate-weights` pre-commit hook entry

The three weight files (`master_repositories.json`, `programming_languages.json`, `token_weights.json`) should not be touched in this PR, nor do we care too much about enforcing a formatting schema.

Once those are pulled, the PR lands at roughly 100 lines of real config + 214 lines of `uv.lock`, which is what the stated scope should be. Everything else stays: ruff/pyright/pytest hooks, trailing-whitespace/EOF/check-yaml/check-json/large-file/mixed-line-ending, CI consolidation, CONTRIBUTING.md.

seroperson:
@anderdc done 👍

anderdc:
One thing to restore: please bring back the auto-commit-format path for internal PRs.

In the old `lint.yml` there were two jobs — internal PRs had `ruff --fix` + `ruff format` + `stefanzweifel/git-auto-commit-action` auto-committing the fixes back, while fork PRs failed loud. We'd like to keep that split using the new pre-commit setup:

- **Internal PR:** run `pre-commit run --all-files` with auto-fixes allowed (it exits non-zero when it makes changes), then `git-auto-commit-action` pushes any formatting tweaks back, then re-run pre-commit to confirm the branch is clean.
- **Fork PR:** run `pre-commit run --all-files` and fail on any diff (no auto-commit possible anyway, no write access).

Everything else (pyright via `--hook-stage pre-push`, pre-commit config, typecheck.yml removal, CONTRIBUTING.md) looks good — keep as-is. Also keep the `permissions: contents: write` since the internal path will use it.

seroperson:
@anderdc done 🫡

anderdc (approved):

PR #513, opened 2026-04-16
refactor: extract should_exclude_maintainer_pr helper

fansilas:
## Summary

The maintainer-author exclusion predicate (`MAINTAINER_ASSOCIATIONS` membership plus the `DEV_MODE` bypass) was inlined identically in `try_add_open_or_closed_pr` and `should_skip_merged_pr`. Pulling it into `should_exclude_maintainer_pr(author_association)` collapses the duplication and gives the rule a single source of truth, so future changes to the bypass policy or the association list only need to touch one place.

The reviewer-side maintainer check in `get_pull_request_maintainer_changes_requested_count` is intentionally left untouched: it reads the REST `author_association` field, has no `DEV_MODE` bypass, and consolidating it would change behavior.

## Related Issues

N/A

## Type of Change

- [x] Refactor

## Testing

- [x] Tests added/updated
- [x] Manually tested

New `TestShouldExcludeMaintainerPr` class in `tests/utils/test_github_api_tools.py` covers:

- All three maintainer associations (`OWNER`, `MEMBER`, `COLLABORATOR`) return `True`
- Non-maintainer values (`CONTRIBUTOR`, `FIRST_TIME_CONTRIBUTOR`, `NONE`, `MANNEQUIN`, `''`, `None`) return `False`
- `DEV_MODE` set to a truthy value bypasses the check for every maintainer association
- Empty `DEV_MODE` still applies the check (Python truthiness behavior preserved)

Local verification:

- `ruff check`: clean
- `ruff format --check`: clean
- `pyright`: 0 errors
- `pytest tests/`: 322 passed

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -828,6 +828,18 @@ def _has_resource_limit_errors(data: Dict) -> bool:
     return any(isinstance(e, dict) and e.get('type') == 'RESOURCE_LIMITS_EXCEEDED' for e in errors)
 
 
+def should_exclude_maintainer_pr(author_association: Optional[str]) -> bool:
+    """Return True when a PR should be excluded from scoring due to a maintainer author.
+
+    Maintainers can merge directly, so their PRs aren't valid scoring targets.
+    DEV_MODE bypasses the check so local testing works against repos where
+    contributors are predominantly maintainers.
+    """
+    if os.environ.get('DEV_MODE'):
+        return False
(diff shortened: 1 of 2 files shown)


anderdc:
**Closing -- the inline check is clear enough as-is.** It's a one-line condition used in two places. Extracting it into a named function, adding a docstring, and writing 36 lines of tests adds more code than it saves. The existing code is readable and self-explanatory.

PR #514, opened 2026-04-16
refactor: migrate PR datetime fields to parse_github_iso_to_utc and deprecate cst wrapper

fansilas:
## Summary

PR #474 introduced `parse_github_iso_to_utc` and rewrote `parse_github_timestamp_to_cst` as a thin wrapper around it (`parse_github_iso_to_utc(...).astimezone(CHICAGO_TZ)`). That left the nine callsites in `gittensor/classes.py` still on the older name. This change migrates those callsites and marks the wrapper with a `DeprecationWarning`, so a follow-up release can drop both the wrapper and the `pytz` dependency once any out-of-tree consumers have had a cycle to migrate.

The migration is behavior-preserving for the downstream arithmetic actually performed on these fields in this codebase (subtraction, comparison, `min`/`max`), all of which are timezone-independent for tz-aware datetimes. No code in the codebase formats the affected fields as zone-localized strings, and no test asserts on Chicago timezone output. The wrapper, `CHICAGO_TZ`, and the `pytz` dependency are intentionally retained this cycle so any external caller is given a deprecation cycle before removal.

This PR is deliberately narrow. The scope here is a single focused step: migrate the nine remaining callsites, attach a deprecation marker, and leave wrapper removal plus the `pytz` drop to a follow-up.

## Related Issues

N/A. Builds on #474.

## Type of Change

- [x] Refactor

## Testing

- [x] Tests added/updated
- [x] Manually tested

New `tests/validator/test_datetime_utils.py` covers `parse_github_iso_to_utc` directly (the helper had no prior tests), with cases for the `Z` suffix, numeric UTC offsets, naive input, equivalent-moment equality across zones, isoformat round-trip idempotency, and surrounding whitespace. The CST wrapper is also covered by a deprecation test and a moment-equality test against the UTC helper.

Local verification:

- `ruff check`: clean
- `ruff format --check`: clean
- `pyright` on `gittensor/`: 0 errors
- `pytest tests/`: 317 passed
- `pytest tests/ -W "error::DeprecationWarning:gittensor.validator.utils.datetime_utils"`: 317 passed (confirms no remaining internal caller of the deprecated wrapper)

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -224,7 +224,7 @@ class PullRequest:
     @classmethod
     def from_graphql_response(cls, pr_data: dict, uid: int, hotkey: str, github_id: Optional[str]) -> 'PullRequest':
         """Create PullRequest from GraphQL API response for any PR state."""
-        from gittensor.validator.utils.datetime_utils import parse_github_timestamp_to_cst
+        from gittensor.validator.utils.datetime_utils import parse_github_iso_to_utc
 
         repository_full_name = parse_repo_name(pr_data['repository'])
         pr_state = PRState(pr_data['state'])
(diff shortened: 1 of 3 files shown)


anderdc:
**Closing -- we use CST intentionally.** The team and server are CST, so `parse_github_timestamp_to_cst` is the correct function for our use case. It doesn't need to be deprecated or migrated away from.

PR #518, opened 2026-04-17
fix: avoid cancel votes when solver lookup fails in issue bounties

### LINKED ISSUE #517: Issue bounties cast cancel votes when solver lookup fails (API/rate-limit path)
## Description

When GitHub solver lookup fails (for example due to GraphQL/API failure or rate limiting), the issue-bounties flow treats it as "no solver found" and casts a cancel vote.  
This makes transient API failures policy-indistinguishable from genuine no-solver closures.

## Steps to Reproduce

1. Make `check_github_issue_closed` see the issue as closed, but force solver lookup failure (e.g. GraphQL failure/`None` result in `find_solver_from_cross_references` path).
2. Run `issue_competitions(...)` with an active contract issue.
3. Observe `vote_cancel_issue(...)` is called with reason `"Issue closed without identifiable solver"`.

## Expected Behavior

If solver lookup fails due to API/GraphQL/rate-limit errors, validator should **not** cast a cancel vote.  
It should skip voting for that issue and retry on a later round.

## Actual Behavior

Solver lookup failure is collapsed into `(None, None)`, and forward logic treats `not solver_github_id` as true no-solver, then casts cancel.

## Environment

- OS: Linux
- Python version: 3.12.3
- Commit/Version: `test` branch before fix (e.g. `9424da7`)

## Additional Context

Impacted paths:
- `gittensor/utils/github_api_tools.py` (`find_solver_from_cross_references`, `check_github_issue_closed`)
- `gittensor/validator/issue_competitions/forward.py`

Root cause:
- Solver lookup failures return `(None, None)` and are not explicitly marked as failure state.
- Forward logic uses falsy solver ID to trigger cancel voting.

MkDev11:
## Summary

- Distinguish GitHub solver lookup failures from true no-solver outcomes in issue bounty resolution.
- Add explicit lookup-failure signaling in `check_github_issue_closed` via `solver_lookup_failed`.
- Update issue bounties forward flow to skip voting when solver lookup fails, instead of casting cancel.
- Preserve existing cancel behavior for genuinely closed issues without an identifiable solver.
  - GraphQL/API lookup failure vs true no-solver behavior in `github_api_tools`.
  - Forward voting policy (skip on lookup failure, cancel on true no-solver).

## Related Issues

Closes #517

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Commands run:
- `uv run pytest tests/utils/test_github_api_tools.py -q`
- `uv run pytest tests/validator/test_issue_competitions_forward.py -q`
- `uv run pytest tests/ -v`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pyright`

## Checklist

- [x] Self-review completed


anderdc (changes requested):
Thanks for catching this — the bug is real, the policy fix is correct, and the two forward-flow tests land in the right spots. Requesting changes on the implementation shape before merge.

**1. Replace `raise_on_api_failure` + `RuntimeError` with a sentinel return.**

Using exceptions for an expected alternative outcome is control flow wearing an exception costume. Catching a generic `RuntimeError` at `github_api_tools.py:1199` will silently swallow any unrelated `RuntimeError` that ever bubbles up from `requests`, `bt.logging`, or future refactors — latent bug. The flag also doubles the surface area of three helpers and produces the readability-hostile bare positional `True` at the call site.

Suggested shape: `find_solver_from_timeline` / `find_solver_from_cross_references` return `Optional[Tuple[Optional[int], Optional[int]]]` where

- `None` means **lookup failed** (retry later)
- `(None, None)` means **closed, no solver identified**
- `(id, pr)` means **solver found**

`check_github_issue_closed` then becomes a clean `if result is None` branch. No flag, no exception hop, no positional-bool call site. `_search_issue_referencing_prs_graphql` should return `Optional[List[PRInfo]]` on the same convention.

**2. Treat any non-empty GraphQL `errors` as a lookup failure.**

The current new block logs a warning and continues to parse `result['data']`, which may be partial or null when errors are present. Make it decisive: if `result.get('errors')` is non-empty, return the failure sentinel (`None` per (1)). Do not log-and-fall-through into partial data.

**3. Update existing tests rather than adding more.**

`test_graphql_query_failure_returns_none` at `tests/utils/test_github_api_tools.py:1007` currently asserts the old `(None, None)` contract — update it to assert the new failure sentinel (`None`) instead. Parametrize or extend it to cover the `errors`-in-response case from (2) in the same test, since both are the same failure policy. `TestFindSolverFromTimeline::test_delegates_to_cross_references` at line 1022 needs the call-site assertion updated to match the simplified signature. The new `TestCheckGithubIssueClosed` class stays but its two tests should be rewritten against the sentinel contract. Net should be fewer tests, not more.

**Nice-to-have (not blocking):** `TypedDict` for the `check_github_issue_closed` return value now that it has up to 4 optional keys — would catch call-site typos.

MkDev11:
> Thanks for catching this — the bug is real, the policy fix is correct, and the two forward-flow tests land in the right spots. Requesting changes on the implementation shape before merge.
> 
> **1. Replace `raise_on_api_failure` + `RuntimeError` with a sentinel return.**
> 
> Using exceptions for an expected alternative outcome is control flow wearing an exception costume. Catching a generic `RuntimeError` at `github_api_tools.py:1199` will silently swallow any unrelated `RuntimeError` that ever bubbles up from `requests`, `bt.logging`, or future refactors — latent bug. The flag also doubles the surface area of three helpers and produces the readability-hostile bare positional `True` at the call site.
> 
> Suggested shape: `find_solver_from_timeline` / `find_solver_from_cross_references` return `Optional[Tuple[Optional[int], Optional[int]]]` where
> 
> * `None` means **lookup failed** (retry later)
> * `(None, None)` means **closed, no solver identified**
> * `(id, pr)` means **solver found**
> 
> `check_github_issue_closed` then becomes a clean `if result is None` branch. No flag, no exception hop, no positional-bool call site. `_search_issue_referencing_prs_graphql` should return `Optional[List[PRInfo]]` on the same convention.
> 
> **2. Treat any non-empty GraphQL `errors` as a lookup failure.**
> 
> The current new block logs a warning and continues to parse `result['data']`, which may be partial or null when errors are present. Make it decisive: if `result.get('errors')` is non-empty, return the failure sentinel (`None` per (1)). Do not log-and-fall-through into partial data.
> 
> **3. Update existing tests rather than adding more.**
> 
> `test_graphql_query_failure_returns_none` at `tests/utils/test_github_api_tools.py:1007` currently asserts the old `(None, None)` contract — update it to assert the new failure sentinel (`None`) instead. Parametrize or extend it to cover the `errors`-in-response case from (2) in the same test, since both are the same failure policy. `TestFindSolverFromTimeline::test_delegates_to_cross_references` at line 1022 needs the call-site assertion updated to match the simplified signature. The new `TestCheckGithubIssueClosed` class stays but its two tests should be rewritten against the sentinel contract. Net should be fewer tests, not more.
> 
> **Nice-to-have (not blocking):** `TypedDict` for the `check_github_issue_closed` return value now that it has up to 4 optional keys — would catch call-site typos.

thanks for the review, just pushed the updates

anderdc (approved):

PR #519, opened 2026-04-17
perf: eliminate a separate API call for requested_reviews

seroperson:
## Summary

- Eliminates 1 REST API call per merged PR by fetching `CHANGES_REQUESTED` reviews inline via the existing GraphQL PR query
- Amount of elements to fetch now calculated dynamically depending on penalty constant: `ceil(1 / REVIEW_PENALTY_RATE)` (no reason to fetch more, as it'll result in 0.0 multiplier anyway)
- Remove the now-unused `get_pull_request_maintainer_changes_requested_count` function and its tests

## Tests

Added several tests to ensure it still fills correctly. 

You can also manually verify that this data is available using GraphQL:

```
# Single PR with known CHANGES_REQUESTED review
gh api graphql -f query='
  query {
    repository(owner: "Entrius", name: "gittensor") {
      pullRequest(number: 487) {
        number
        title
        state
        reviews(first: 3, states: APPROVED) {
          nodes { author { login } }
        }
        changesRequestedReviews: reviews(first: 7, states: CHANGES_REQUESTED) {
          nodes { authorAssociation }
        }
      }
    }
  }'
```

And compare it with REST:

```
gh api repos/Entrius/gittensor/pulls/487/reviews --jq '[.[] | select(.state == "CHANGES_REQUESTED") | {state, author_association}]'
```

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -8,7 +8,7 @@ from typing import DefaultDict, Dict, List, Optional, Set
 
 import bittensor as bt
 
-from gittensor.constants import MAX_CODE_DENSITY_MULTIPLIER, MIN_TOKEN_SCORE_FOR_BASE_SCORE
+from gittensor.constants import MAINTAINER_ASSOCIATIONS, MAX_CODE_DENSITY_MULTIPLIER, MIN_TOKEN_SCORE_FOR_BASE_SCORE
 from gittensor.utils.utils import parse_repo_name
 
 GITHUB_DOMAIN = 'https://github.com/'
@@ -290,6 +290,13 @@ class PullRequest:
         last_edited_at = parse_github_timestamp_to_cst(raw_edited_at) if isinstance(raw_edited_at, str) else None
         merged_at = parse_github_timestamp_to_cst(pr_data['mergedAt']) if is_merged else None
 
+        changes_requested_count = 0
+        if is_merged:
+            cr_reviews = pr_data.get('changesRequestedReviews', {}).get('nodes', [])
+            changes_requested_count = sum(
+                1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS
+            )
+
         # Extract last label from timeline events
         timeline_nodes = pr_data.get('timelineItems', {}).get('nodes', [])
         label = timeline_nodes[0]['label']['name'].lower() if timeline_nodes else None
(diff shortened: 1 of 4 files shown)


anderdc (changes requested):
Great optimization — eliminates a REST round-trip per merged PR, reuses the existing GraphQL retry path, and the `ceil(1 / REVIEW_PENALTY_RATE)` bound is a nice touch since the multiplier already clamps to 0 beyond that. Requesting changes on style consistency and framing before merge.

**1. Use a GraphQL variable instead of Python `%d` interpolation.**

Every other GraphQL query in this file parameterizes via variables, not string formatting:

- `QUERY` at `github_api_tools.py:36-37` declares `query($userId: ID!, $limit: Int!, $cursor: String)` and uses `first: $limit` at line 41, with values passed via `variables={...}` at line 757-761.
- `_PR_TIMELINE_QUERY` at `github_api_tools.py:453-454` follows the same pattern.

This PR introduces the only `QUERY % _MAX_CHANGES_REQUESTED_REVIEWS` in the file. Suggested changes to match the existing convention:

- Append to the signature: `query($userId: ID!, $limit: Int!, $cursor: String, $maxChangesRequestedReviews: Int!)`
- In the body: `reviews(first: $maxChangesRequestedReviews, states: CHANGES_REQUESTED) { nodes { authorAssociation } }`
- At the call site (`line 757`): add `'maxChangesRequestedReviews': _MAX_CHANGES_REQUESTED_REVIEWS` to the variables dict

The query string stays a static triple-quoted constant, and the parameter flows through the same channel as `$limit` and `$cursor`.

**2. Commit prefix should be `perf:`, not `fix:`.**

There's no bug here — miner scoring output is identical before and after this PR (the formula, inputs, and gating are unchanged). The gains are operational: fewer API round-trips, less REST quota, simpler error surface. That is textbook `perf:` per conventional commits. Mislabeling it as `fix:` confuses release-note generators and downstream tooling that categorize by prefix.

**3. Add an invariant test for `_MAX_CHANGES_REQUESTED_REVIEWS`.**

One line asserting `_MAX_CHANGES_REQUESTED_REVIEWS == ceil(1 / REVIEW_PENALTY_RATE)`. If `REVIEW_PENALTY_RATE` is ever tuned, this surfaces the coupling between the rate and the fetch cap so nobody silently breaks the "beyond cap, multiplier is already 0" guarantee.

anderdc (approved):

PR #522, opened 2026-04-17
fix: add retry logic for transient DB failures in repository

### LINKED ISSUE #520: Add retry logic for transient database failures in repository.py
## Summary
Database operations in `gittensor/validator/storage/repository.py` have no retry mechanism.
Any transient failure (connection reset, pool exhaustion, lock timeout) immediately logs and returns `None` or `False`, with no attempt to recover.

## Motivation
Transient database errors are a normal operational reality — connection pools get exhausted, network blips cause brief disconnects, and Postgres occasionally rejects connections under load.
Without retries, a single transient failure can silently corrupt a validator cycle result (returning "no data" instead of the real value). 
Adding exponential backoff retries for `OperationalError` and `InterfaceError` would make the validator resilient to these common failure modes without masking genuine errors.

**Affected file:**
- `gittensor/validator/storage/repository.py`

dataCenter430:
## Summary

- Adds `_execute_with_retry` helper to `BaseRepository` that retries on `psycopg2.OperationalError` and `InterfaceError` (connection resets, pool exhaustion) up to 3 times with 1 / 2 / 4 s exponential backoff
- Refactors all 5 DB methods (`execute_command`, `store_pull_requests_bulk`, `store_issues_bulk`, `store_file_changes_bulk`, `set_miner_evaluation`) to use it
- Non-transient errors (e.g. `ProgrammingError`, `IntegrityError`) are not retried and still propagate immediately

## Errors found during review

**1. Unused `typing.Any` import (ruff F401)**
```
F401 `typing.Any` imported but unused
  --> repository.py:12:20
12 | from typing import Any, Callable, List, TypeVar
   |                    ^^^
```

**2. `from psycopg2.extras import execute_values` was placed outside the `_run()` closures in all 4 bulk methods — an `ImportError` would escape `_execute_with_retry` entirely instead of being caught as a non-transient error**

Both fixed in this PR.

## Test plan

- [x] `_execute_with_retry` returns value on first success
- [x] Non-transient error returns default immediately without retrying
- [x] `OperationalError` retried with success on attempt 3
- [x] Backoff delays are exactly `(1, 2, 4)` seconds
- [x] Default returned after all 4 attempts exhausted
- [x] `db.rollback()` called on every transient failure
- [x] `psycopg2.InterfaceError` treated as transient
- [x] `execute_command` integration: retries on `OperationalError`, returns `False` on persistent error

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -9,7 +9,7 @@ and miner evaluations.
 import logging
 import time
 from contextlib import contextmanager
-from typing import Any, Callable, List, TypeVar
+from typing import Callable, List, TypeVar
 
 try:
     import psycopg2
@@ -226,9 +226,8 @@ class Repository(BaseRepository):
                 )
             )
 
-        from psycopg2.extras import execute_values
-
         def _run() -> int:
+            from psycopg2.extras import execute_values
             with self.get_cursor() as cursor:
                 execute_values(
                     cursor,
@@ -282,9 +281,8 @@ class Repository(BaseRepository):
                 )
             )
 
-        from psycopg2.extras import execute_values
-
         def _run() -> int:
+            from psycopg2.extras import execute_values
             with self.get_cursor() as cursor:
                 execute_values(
                     cursor, BULK_UPSERT_ISSUES.replace('VALUES %s', 'VALUES %s'), values, template=None, page_size=100
@@ -324,10 +322,10 @@ class Repository(BaseRepository):
                 )
             )
 
-        from psycopg2.extras import execute_values
         prs = {(fc.pr_number, fc.repository_full_name) for fc in file_changes}
 
         def _run() -> int:
+            from psycopg2.extras import execute_values
             with self.get_cursor() as cursor:
                 execute_values(
                     cursor,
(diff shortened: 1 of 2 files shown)


anderdc:
**Closing -- we haven't experienced transient DB failures in production.** The connection pool and DB infrastructure have been stable. Adding retry logic with exponential backoff for a problem that doesn't exist adds complexity and masks potential real issues that we'd want to surface immediately rather than silently retry.

If we start seeing OperationalError/InterfaceError in the logs, we'll revisit.

dataCenter430:
Hi, @anderdc 
Thanks for the detailed feedback -- fair point on YAGNI.

A couple of things worth considering before closing:

It doesn't mask failures - it surfaces them more clearly. The retry only catches OperationalError/InterfaceError (connection-level errors), logs a WARNING on each attempt, and logs an ERROR after all retries are exhausted. 
Silent failures aren't possible - every retry attempt is logged with the attempt count and delay.

Transient failures are common during routine ops. Connection pool exhaustion under load, DB restarts during deploys, brief network blips - these are well-known psycopg2 failure modes. 
The fact that we haven't noticed them doesn't mean they aren't happening; it may mean they're silently dropping operations today.

The change is small and isolated. It adds one private method (_execute_with_retry) to BaseRepository with 10 tests. 
Removing it later is trivial if it proves unnecessary.

That said, I understand the team's preference to keep things simple until there's a concrete need. If you're open to a middle ground- merging with the retry count reduced to 1 (a single immediate retry) - that would catch the vast majority of transient blips with almost no added complexity. Otherwise, happy to keep this branch open and revisit when we see OperationalError in the logs.

dataCenter430:
Hi, @anderdc 
Would you please consider the PR?
I think it's really needed for gittensor.

PR #523, opened 2026-04-17
feat: add GitHub API rate limit observability

### LINKED ISSUE #521: Add GitHub API rate limit observability to the validator
## Summary
The validator makes frequent GitHub API calls but has no instrumentation for remaining rate limit quota. 
There is no logging, no metric, and no early-warning mechanism before the quota is exhausted.

## Motivation
GitHub's REST API enforces a hard per-token rate limit. When the validator hits the limit mid-cycle, requests silently fail or return errors that are caught by generic exception handlers. 
There is currently no way to know how close to the limit the validator is, which tokens are being exhausted fastest, or when a cycle was affected by rate limiting.

Adding rate limit tracking — by reading `X-RateLimit-Remaining` / `X-RateLimit-Reset` headers from every GitHub API response — would enable proactive throttling, better error messages, and the foundation for quota-aware scheduling.

**Affected file:**
- `gittensor/utils/github_api_tools.py`

dataCenter430:
## Summary

- Adds `_log_rate_limit(response, context)` helper that reads `X-RateLimit-Remaining`, `X-RateLimit-Limit`, and `X-RateLimit-Reset` headers from every GitHub API response
- Logs at `debug` level normally; escalates to `warning` when remaining quota drops below 200
- Wired into 7 call sites: `/user`, `compare`, PR files, PR reviews, `check_github_issue_closed`, `graphql`, and `graphql/prs`

## Error found during review

**`TypeError` crash breaking all 67 existing tests**

`response.headers.get(...)` returns a `Mock` object in the existing test fixtures. `int(Mock())` raises `TypeError`, which the original code only caught as `ValueError` — causing every test that exercises a 200-response path to fail immediately.

```
gittensor/utils/github_api_tools.py in get_github_graphql_query
    _log_rate_limit(response, 'graphql/prs')
gittensor/utils/github_api_tools.py in _log_rate_limit
    remaining_int = int(remaining)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Mock'
```

Fixed by catching `(ValueError, TypeError)` instead of `ValueError` alone.

## Test plan

- [x] Debug log emitted when remaining is above threshold
- [x] Warning log emitted when remaining is below 200
- [x] Boundary: 199 → warning, 200 → debug
- [x] No log when `X-RateLimit-Remaining` header is absent
- [x] No crash when header value is a `Mock` (guards existing test fixtures)
- [x] Context label appears in log message
- [x] Reset timestamp included in log message
- [x] Invalid reset header handled gracefully without raising
- [x] All 67 pre-existing tests continue to pass


dataCenter430:
This adds lightweight rate limit logging to every GitHub API call — DEBUG level normally, WARNING when remaining drops below 200. No behaviour changes, just visibility. 
Happy to adjust the threshold or log level if needed. Would appreciate a review when you get a chance.

dataCenter430:
Hi, @anderdc 
Would you please review the PR carefully?
If you really won't it, please let's discuss with me before close. 🙏 
I can reproduce the edge-cases for this PR to prove why it's needed..
Anyway, hope your feedback.
Thanks

anderdc:
Hi visibility is not needed, we're working on a github mirror so api usage isn't particularly helpful thanks!

PR #529, opened 2026-04-17
fix: skip inline test downweight when tests pre-exist in old content

### LINKED ISSUE #528: [BUG] Inline test detection penalizes production code changes in files with pre-existing tests
## Description

When a Rust, Zig, or D source file already contains inline tests (e.g. `#[cfg(test)]` in Rust) **before** the PR, and the PR only modifies production code in that file, the entire file is incorrectly scored with the 0.05x test weight instead of 1.0x.

The inline test detection only looks at the file's new version. It sees the pre-existing test block and marks the whole file as a test file - even though the tests were already there and the PR didn't touch them.

This causes a **95% score loss** on production code contributions to any file that happens to contain inline tests.

## Steps to Reproduce

1. Take any Rust file that has `#[cfg(test)]` at the bottom (standard Rust convention)
2. Submit a PR that only adds or modifies production code above the test module
3. The file is scored at 0.05x instead of 1.0x

## Expected Behavior

If the file already had inline tests before the PR, the test module is pre-existing and should not trigger the test penalty. The file should keep the full 1.0x weight.

## Actual Behavior

The file is marked as a test file and scored at 0.05x weight, regardless of whether the tests existed before the PR or were newly introduced.

Verified locally:
Expected score (1.0x weight): 1.78
Actual score   (0.05x weight): 0.09
Score loss: 95%

## Environment

- OS: Linux
- Python version: 3.12

## Additional Context

Affected languages: Rust, Zig, D - any language with idiomatic inline tests.

Affected repos include `opentensor/subtensor`, `bevyengine/bevy`, `zed-industries/zed`, and any other Rust project where `#[cfg(test)]` is standard practice.

The fix is to also check the old version of the file. If both old and new contain inline tests, keep the full weight. Only apply the test penalty when the PR itself introduces the tests.

HadesHappy:
## Summary

- Fix inline test detection to check both old and new file content
- Only apply 0.05x test weight when the PR itself introduces inline tests
- If the file already had inline tests before the PR, keep full 1.0x weight
- Add 4 tests covering pre-existing, newly added, new file, and removed scenarios

Closes #528 

## Problem

Inline test detection only checked the new version of the file. Any Rust/Zig/D file with pre-existing inline tests (e.g. `#[cfg(test)]`) was penalized at 0.05x weight even when the PR only changed production code. This caused a 95% score loss on production contributions to files with existing tests.

## Test plan

- [ ] `test_preexisting_inline_tests_keep_full_weight` — old and new both have tests, prod-only changes → full weight
- [ ] `test_new_inline_tests_get_test_weight` — old has no tests, PR adds tests → 0.05x
- [ ] `test_new_file_with_inline_tests_gets_test_weight` — brand new file with tests → 0.05x
- [ ] `test_removed_inline_tests_keep_full_weight` — old had tests, new doesn't → full weight
- [ ] All 17 existing inline test detection tests still pass
- [ ] Full test suite passes (325 tests)


HadesHappy:
Hi @anderdc 
Could you please review this PR please?

anderdc:
What happens if the file already had tests and the change introduces more tests in the file?

PR #530, opened 2026-04-17
enh: Cleaver label matching

seroperson:
Currently label multiplier only applies if there is exact match, which is not always the case considering the variety of labels in the wild.

The real-world example is: https://gittensor.io/miners/pr?repo=infiniflow%2Fragflow&number=14112

## Summary

- Now labels are matched by substring search.
- Added more labels to check, like `crash`, `perf`, `chore`, `feat` etc
- Added some tests to ensure everything works as expected.

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -166,7 +166,7 @@ class PullRequest:
     time_decay_multiplier: float = 1.0
     credibility_multiplier: float = 1.0
     review_quality_multiplier: float = 1.0  # Penalty for CHANGES_REQUESTED reviews from maintainers
-    label_multiplier: float = 1.0  # Multiplier based on PR label (feature, bug, enhancement, refactor)
+    label_multiplier: float = 1.0  # Multiplier based on PR label (exact match against known labels)
     label: Optional[str] = None  # Last label set on the PR
     changes_requested_count: int = 0  # Number of maintainer CHANGES_REQUESTED reviews
     earned_score: float = 0.0
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -74,12 +74,27 @@ CONTRIBUTION_SCORE_FOR_FULL_BONUS = 1500
 # Boosts
 MAX_CODE_DENSITY_MULTIPLIER = 1.5
 
-# Label multipliers — applied based on the last label set on the PR (requires triage+ access)
+# Label multipliers - applied based on the last label set on the PR (requires triage+ access)
 LABEL_MULTIPLIERS: dict[str, float] = {
+    # features
     'feature': 1.50,
+    'feat': 1.50,
+    # bug fixes
     'bug': 1.25,
+    'fix': 1.25,
+    'crash': 1.25,
+    'regression': 1.25,
+    'security': 1.25,
+    # enhancements
     'enhancement': 1.10,
+    'improve': 1.10,
+    'perf': 1.10,
+    # refactors
     'refactor': 1.00,
+    'cleanup': 1.00,
+    'polish': 1.00,
+    'debt': 1.00,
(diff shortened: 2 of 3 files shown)


anderdc (changes requested):
Good motivation — the real-world label variance is a legitimate problem — but this solution overcomplicates the data structure, introduces a latent bug, and creates false-positive risk that we're better off avoiding entirely. Requesting changes to simplify.

**1. Revert `LABEL_MULTIPLIERS` to a flat `dict[str, float]`. Just extend it with the standard entries we were missing, no substring matching.**

Keep exact-match semantics and expand the keyword set explicitly:

```python
LABEL_MULTIPLIERS: dict[str, float] = {
    # features
    'feature': 1.50,
    'feat': 1.50,
    # bug fixes
    'bug': 1.25,
    'fix': 1.25,
    'crash': 1.25,
    'regression': 1.25,
    'security': 1.25,
    # enhancements
    'enhancement': 1.10,
    'improve': 1.10,
    'perf': 1.10,
    # refactors
    'refactor': 1.00,
    'cleanup': 1.00,
    'polish': 1.00,
    'debt': 1.00,
    'chore': 1.00,
}
```

This is simpler, eliminates the triple-nested-tuple type, and **structurally prevents the `'perf'` duplication bug** in the current diff (`'perf'` appears in both the 1.25 tier and the 1.10 tier — since first-match wins in the current implementation, the 1.10 entry is unreachable dead code; dict keys can't collide, so the type itself rejects this class of error).

Tradeoff acknowledged: `kind/feature` and `🐞 bug` won't match under exact lookup. That's a deliberate policy — we reward conventional labels, not every label in the wild. If a maintainer labels a PR `kind/feature` instead of `feature`, they don't get the bonus. Simpler, predictable, no false positives.

**2. Remove `_get_label_multiplier` from `scoring.py`.**

With a flat dict and exact match, the helper isn't needed. Restore the one-liner:

```python
pr.label_multiplier = LABEL_MULTIPLIERS.get(pr.label, 1.0) if pr.label else 1.0
```

This is what the file had before the PR, and it's the simplest possible expression of the lookup.

**3. Trim tests significantly.**

Most of the 18 parametrized cases are tautological — they enumerate the dict and assert each key maps to its value, which is just re-asserting the dict literal. Once substring matching is gone, the false-positive tests (`wontfix`, `kind/feature`, emoji prefix) are moot.

A minimal test suite:

```python
@pytest.mark.parametrize('label,expected', list(LABEL_MULTIPLIERS.items()))
def test_known_labels(label, expected):
    assert LABEL_MULTIPLIERS.get(label, 1.0) == expected

@pytest.mark.parametrize('label', ['docs', 'question', 'wontfix', 'kind/feature'])
def test_unknown_labels_return_default(label):
    assert LABEL_MULTIPLIERS.get(label, 1.0) == 1.0
```

Two tests that actually test something meaningful: one that every entry is reachable, one that anything outside the set falls through to the default. Done.

seroperson:
I've addressed all comments.

> Tradeoff acknowledged: `kind/feature` and `🐞 bug` won't match under exact lookup. That's a deliberate policy — we reward conventional labels, not every label in the wild. If a maintainer labels a PR `kind/feature` instead of `feature`, they don't get the bonus. Simpler, predictable, no false positives.

Maybe with such flatten structure we can just define this false-positive `wontfix` at the top of dict (like `wontfix: 1.0`) and restore the substring search? `wontfix` actually the only false-positive I can imagine.

anderdc:
PR looks good now, let's add this and see how it goes, I imagine (further down the line), repositories would set their specific set of labels and multipliers at the per repo level when they register. Since we're a ways away from that, this is a good start.

anderdc (approved):

PR #545, opened 2026-04-18
Scoring deleted code with a lower weight

seroperson:
## Summary

Currently deleting an unused function gives you the same score as if you would write it from scratch. There is no difference between added and deleted code, both are scored the same.

It results in PRs like:
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=468 (28.76 base score)
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=448 (22.32 base score)
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=449 (14.92 base score)
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=450 (33.73 base score)
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=487 (35.82 base score)
https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=468 (28.76 base score)

(and counting)

All these PRs only delete dead code, but scored like an average solid feature.

We already give zero score for completely deleted files and lowering the weight of deleted code is just getting in line with existing logic, so maybe we should lower weight even more. But I guess this value is something in the middle to still reward deleting somehow.

Changes introduced in this PR:

- Introduced a lowered weight for deleted code, so deleting code won't scored as like it was added.
- Calculating density is tweaked just a little, so deleted lines contribute with its' own weight. It prevents situations when deleting code reduces your source density by contributing lowered score.

## Testing

- Added a test which ensures that the same code snippet added / deleted results in different score.


ClaudeCenter:
Hi @seroperson 
I hear you, but I actually disagree. You're assuming deleting code is easier than adding it, but I see it the other way around.
Think about it: even a kid can learn to code and slap together a basic new feature, but they’d have no idea how to review or understand an entire project. They can't just delete code because they’d likely break something important. In my opinion, the skill required to maintain a project by safely removing dead code is actually much harder than just adding something new. We should view PRs that delete code as a sign of senior-level expertise. The high score is totally justified, if anything, we should be rewarding it even more. Also, regarding your point, if everyone tries to lower each other's scores just because they're worried about their own ranking, that kind of behavior goes against the whole spirit of OSS.

codege333:
> ## Summary
> Currently deleting an unused function gives you the same score as if you would write it from scratch. There is no difference between added and deleted code, both are scored the same.
> 
> It results in PRs like: https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=468 (28.76 base score) https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=448 (22.32 base score) https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=449 (14.92 base score) https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=450 (33.73 base score) https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=487 (35.82 base score) https://gittensor.io/miners/pr?repo=entrius%2Fgittensor&number=468 (28.76 base score)
> 
> (and counting)
> 
> All these PRs only delete dead code, but scored like an average solid feature.
> 
> We already give zero score for completely deleted files, so lowering the weight of deleted code is just getting in line with existing logic, so maybe we should lower weight even more.

Honestly, don’t try to downplay other people's efforts. You shouldn't worry about what others are doing. Like @ClaudeCenter  said, deleting code safely is very difficult. I think you are saying this only because you don't have the skill to do it yourself.

seroperson:
It's funny how much attention it receives.

> Honestly, don’t try to downplay other people's efforts. You shouldn't worry about what others are doing. Like @ClaudeCenter said, deleting code safely is very difficult. I think you are saying this only because you don't have the skill to do it yourself.

🤡

> We should view PRs that delete code as a sign of senior-level expertise. The high score is totally justified, if anything, we should be rewarding it even more.

Finding code references is the one click action in IDE / well-configured editor.

> if everyone tries to lower each other's scores just because they're worried about their own ranking, that kind of behavior goes against the whole spirit of OSS.

Abusing the obvious bug is actually the kind of behavior which goes against the whatever

anderdc:
to add intuition, the label multiplier helps features stand out, I would argue this is more of a base score calculation issue and the density. However I also think that code deletion is a fairly finite 'resource'. Once deletions happen they won't occur again, or they get cleaned up quick. I almost like it as a quick sweep for cleaning up a repository.

Anyways, the score will decay to 0 and no more dead code will be found, I find codebase clean-up good for repo health.

Finally it's not a bug, this was known, the current scoring setup rewards small, targeted PRs more than anything else (they're easier to review and mitigate massive PRs), perhaps think on somehow tuning the contribution bonus to further balance a 'bigger' PR, but again remember we don't want people just spamming big un-reviewable PRs.

PS:
to the bots in this thread please don't harass people or you will be reported

seroperson:
> Finally it's not a bug, this was known, the current scoring setup rewards small, targeted PRs more than anything else (they're easier to review and mitigate massive PRs), perhaps think on somehow tuning the contribution bonus to further balance a 'bigger' PR, but again remember we don't want people just spamming big un-reviewable PRs.

Is it intended to give the newly-introduced code the same score as deleted? Deleting the code is far not the same work as introducing it, and it seems strange to score deleting a function the same as the function itself.

> Anyways, the score will decay to 0 and no more dead code will be found, I find codebase clean-up good for repo health.

Well, it still gives something and all interested parties still can clean it with some score received (although it doesn't match with the fact that we don't score deleted files). Just it likely won't be a first priority. Currently scored in 35.82 PR will result in 8.95, which is actually still high enough.

And although it's finite in a concrete repo, there are still many other repos with (I guess) a lot of dead code and refactoring opportunities. Finding another refactoring-friendly repo will result in another score pump using dead-code elimination, as it's the easiest and likely risk-free option than fixing bugs and/or implementing features, which give the same score (*), but must be researched, reproduced, tested, discussed and so on.

(*) Note regarding the same score: label multipliers make things just a little bit better: 
- not every repo uses labels
- feature label is kind of good, but fix/enh are x1.25 / x1.1 and it doesn't make you to choose them over the risk-free refactoring (especially dead-code elimination), considering all things mentioned above (researching, reproducing, testing etc). At the end, fixing a bug or implementing an enh usually results in not so much tokens (and score), so multiplying it by x1.25/x1.1 doesn't make it much more attractive.

bittoby:
> I almost like it as a quick sweep for cleaning up a repository.

Totally agree. Removing dead code is like curing cancer. 
Writing codes is easy even with AI but keeping code simple and clean is the main point of code quality.
I wish some miners focus on their contribution rather than blaming others.

anderdc:
I would then think the best solution is setting `refactor` label mulitplier to 0.5 no? it's the simplest. Yes I know not all repositories use labels, but the goal is to make Gittensor useful to US, when we can see clear utility trying various labels, multipliers, etc, then I think we'd be clearly set to make some product with Gittensor. (to give you an idea on what we're doing now and where we're trying to go with this)

I'd prefer a simpler solution like that ^ rather than going and tuning up the base mechanics of scoring. Deleted code to me is about on par of value for adding code. If this was pre-AI, that'd have more weight, but now it's very easy to generate code, lowering the threshold by magnitudes.

So yes I understand your qualms, I'd argue though we should use the label multiplier to tune this up rather than the core scoring calculation.

seroperson:
> I would then think the best solution is setting `refactor` label mulitplier to 0.5 no? it's the simplest.

It's actually what I proposed at day 0 of label multipliers: https://github.com/entrius/gittensor/pull/465#issuecomment-4257462359 (maybe went unseen)
I'm up for this solution and I can re-write this PR to introduce this instead.

_Although still I think that deleting code should not be scored like newly introduced code, it just seems wrong even logically. Scoring process is not ideal and polishing it would be more generalized solution, rather than tuning it with label multipliers. Still label multipliers are project-specific and usually require manual intervention, while scoring is fully-automatic and ecosystem-wide. Well, it's up to you, of course._

> Deleted code to me is about on par of value for adding code. If this was pre-AI, that'd have more weight, but now it's very easy to generate code, lowering the threshold by magnitudes.

It was one-click action pre-AI and now it's even simpler, as AI can analyze false-positives for you. Some can argue that it's some exceptionally hard senior-level work, but it always was one-click, as well as many other refactoring PRs submitted here (or in recently inactivated dify). https://github.com/jendrikseipp/vulture even can be integrated into pre-push and/or CI process (with whitelisting necessary false-positives and exceptions). I can try to integrate this as well if you would like.

So, what should I do here then?

anderdc:
Sorry yea  we didn't see that comment we were having some deployment issues. It's ok to disagree, I do have some more points for you to try to sway to my point of view:

>A flat deletion discount can't distinguish the work it's penalizing. The AST diff treats all deleted lines as equivalent, but deletions come from very different places:
>- Refactor PR that only removes dead code (the gaming vector you flagged)
>- Bug fix that rewrites a buggy function (30 old lines removed, 25 corrected ones added)
>- Feature that consolidates three helpers into one (200 removed, 80 added)
>- Optimization that replaces a complex algorithm with a simpler one (100 removed, 30 added)

>Three of those four are high-value work where the deletion IS the feature. There's no AST-level signal that separates "pure cleanup" from "feature-with-deletions". Intent lives with the label not necessarily the diff.

>The asymmetry also creates a new, worse gaming vector. Rewarding additions at 100% and deletions at 25% makes the rational strategy: add more code, delete less. Write new abstractions beside old ones instead of replacing them. Add workarounds instead of fixing root causes. Leave cruft in place because nobody wants a 75% score cut for cleaning it up. That directly contradicts every software engineering principle about keeping codebases lean, and the resulting bloat is harder to spot in review than the current 'cleanup-gaming' is.

With this, I lean more towards adjusting label multipliers, however, I encourage you to open up another issue or even discussion thread to continue thinking upon this issue. Realistically, we don't want every type of PR to get the same penalty, we're just trying to reel in score for `refactor` PRs in the end right?

tl;dr
- let's change this PR to JUST readjust label multiplier weights. Give your opinion on what the `refactor` label multiplier should be and we'll go from there.
- Open up another issue (or discussion) to further talk on this. I'm still of the opinion "we want to reduce score of just `refactor` type PRs"

anderdc:
conflicts from your previous PR

anderdc (approved):

PR #546, opened 2026-04-18
fix: escape file paths in GraphQL object expressions to prevent query breakage

### LINKED ISSUE #532: [BUG] GraphQL file-content fetch interpolates SHAs and filenames via f-string, enabling query-breakage
## Description

`_fetch_file_contents_with_base_batch` builds a GraphQL query by f-string interpolation of `base_sha:base_path` and `head_sha:head_path` directly into the query body rather than passing them as `$variables`. Filenames can legitimately contain `"`, `\`, and newlines, which will break out of the GraphQL string literal and either cause a malformed query (dropping the file's score) or, in a hostile case, produce unexpected query
structure. SHAs are currently safe but the pattern is fragile.

See `gittensor/utils/github_api_tools.py`, lines 1205–1234.

## Steps to Reproduce

1. A merged PR adds a file whose name contains `"` (rare but legal on case-sensitive  filesystems).
2. Scoring runs `_fetch_file_contents_with_base_batch` for that PR.
3. GraphQL returns a parse error; the PR's token_score silently drops.

## Expected Behavior

Each `object(expression: …)` field is populated from a GraphQL variable, e.g. `object(expression: $base0)` with `$base0: String!` declared in the query header.

## Actual Behavior

f-string interpolation into the query body; a malformed filename produces a GraphQL error and `results[fc.filename] = FileContentPair(None, None)` for the entire batch.

## Additional Context

Suggested fix: generate `$base0: String, $head0: String, …` in the query header, populate a `variables` dict keyed by those names, and reference them from each field selector. Refactor the query builder into a helper that returns
`(query_string, variables_dict)`.

HadesHappy:
Fixes #532 

## Summary

- Extract `_escape_graphql_string()` to sanitize special characters in file paths
- Extract `_build_blob_field()` to build safe GraphQL `object(expression: ...)` fields
- Replace raw f-string interpolation with escaped helper calls
- Add 11 tests covering escaping edge cases and field construction

## Problem

`_fetch_file_contents_with_base_batch` interpolates `sha:path` directly into a GraphQL
string literal via f-string. File paths containing `"`, `\`, or newlines break the query,
causing the file to silently get zero score. The `_build_blob_field` helper escapes the
expression before interpolation.

## Test plan

- [ ] `TestEscapeGraphQLString` — 7 tests: plain path, quotes, backslashes, newlines, carriage returns, combined, empty
- [ ] `TestBuildBlobField` — 4 tests: normal path, quotes in path, backslash in path, alias preserved
- [ ] Existing `TestFetchFileContentsWithBase` renamed-file test still passes (verifies backward compatibility)
- [ ] Full test suite passes (337 tests)

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -1198,6 +1198,25 @@ class FileContentPair:
     new_content: Optional[str]  # None for deleted files
 
 
+def _escape_graphql_string(value: str) -> str:
+    """Escape special characters for safe interpolation inside a GraphQL double-quoted string.
+
+    File paths may legitimately contain backslashes, double quotes, or newlines,
+    all of which break the ``object(expression: "...")`` syntax if unescaped.
+    """
+    return value.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n').replace('\r', '\\r')
+
+
+def _build_blob_field(alias: str, sha: str, path: str) -> str:
+    """Build a GraphQL ``object(expression: ...)`` field for fetching blob content.
+
+    Escapes the ``sha:path`` expression so that special characters in file paths
+    cannot break out of the GraphQL string literal or corrupt the query.
+    """
+    escaped_expr = _escape_graphql_string(f'{sha}:{path}')
+    return f'{alias}: object(expression: "{escaped_expr}") {{ ... on Blob {{ text byteSize isBinary }} }}'
+
+
 def _fetch_file_contents_with_base_batch(
     repo_owner: str,
     repo_name: str,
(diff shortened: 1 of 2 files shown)


HadesHappy:
Hi @anderdc 
Could you review my PR please?

HadesHappy:
Hi @anderdc
Could you review my PR please?

anderdc:
Can you point to a PR where we've had this problem?

HadesHappy:
Hi @anderdc 
What I mean is this would be the best practice.  No existing PR has hit this so far. Currently Tracked repos use clean paths.
The fix is preventive and I think it is worth to keep for reliability.

PR #548, opened 2026-04-18
refactor: extract per-file scoring branches from calculate_token_score_from_file_changes

### LINKED ISSUE #547: Refactor: extract per-file scoring branches from calculate_token_score_from_file_changes
## Description

`calculate_token_score_from_file_changes` in `tree_sitter_scoring.py` is 167 lines with four distinct scoring branches inlined into a single for-loop:

1. Removed files (line 278-286)
2. Non-code files via line-count (line 287-298)
3. Skipped files — binary, oversized, unsupported (lines 300-331)
4. Code files via tree-diff (lines 332-360)

Each branch builds a `FileScoreResult` with different logic, but they all live inside one function. This makes the function hard to read, hard to test individually, and hard to extend.

Extracting each branch into a named helper keeps the main loop as a dispatcher and lets each scoring path be understood and tested in isolation. 

## Suggested approach

Extract three helpers (removed + skipped can stay inline since they just return zero-score results):

- `_score_non_code_file(file, ext, programming_languages, file_weight)` — line-count scoring
- `_score_code_file(file, ext, content_pair, weights, programming_languages, file_weight)` — tree-diff + inline test detection
- `_build_category_results(cat_files, cat_score, cat_nodes, cat_lines, cat_breakdowns)` — category sub-result assembly

The main function becomes a dispatcher loop + accumulator.

edwin-rivera-dev:
## Summary

- Extract `_score_non_code_file()` for line-count scoring path
- Extract `_score_code_file()` for tree-diff scoring path with inline test detection
- Extract `_build_category_results()` for per-category sub-result assembly
- Main function reduced from 167 to ~60 lines as a clean dispatcher loop

## Problem

`calculate_token_score_from_file_changes` is 167 lines with four inlined scoring branches.
Each branch has independent logic that is hard to read, test, or extend as a single block.

## Test plan

- [ ] All existing tree-diff and inline test detection tests pass unchanged
- [ ] Full test suite passes
- [ ] `ruff check` and `ruff format` clean

Closes #547

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -1,6 +1,7 @@
 # The MIT License (MIT)
 # Copyright © 2025 Entrius
 from collections import Counter
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import bittensor as bt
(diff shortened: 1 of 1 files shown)


anderdc (changes requested):
**Scope this down to just `_score_code_file` and come back.**

The four-branch `else` block (binary / too-large / unsupported / tree-diff + inline-test mutation) is the one extraction worth doing — that's where readability actually improves. Everything else in this PR is noise:

- **Drop `_score_non_code_file`** — 10 lines of straight-line code, fine inline.
- **Drop `_build_category_results`** — trivial 10-line loop, fine inline.
- **Revert the cosmetic reformatting** (single-line `PrScoringResult(...)` early return, blank-line removals). Unrelated to the refactor and makes the diff harder to read.
- **Replace the 7-arg signature on `_score_code_file`** with a small context dataclass (`file`, `ext`, `content_pair`, `weights`, `programming_languages` as one `ScoringContext`; `is_test_file` / `file_weight` stay as individual args since they're mutated). 7 positional params is the signal the extraction is carrying too much shared state.
- **Run CI** before the next push — the branch currently has zero checks, and the test checklist in the PR body is unchecked.

As it stands the PR grows the file by +18 lines net and the main function "reduction from 167 to ~60 lines" is just relocation. A tight version of this refactor is ~20–30 lines of diff. Please come back with that.

edwin-rivera-dev:
@anderdc 
Could you review again please? 👍

anderdc:
unnecessary abstraction. No other function uses that code.

PR #563, opened 2026-04-18
refactor: use emit_json for JSON output in view commands

### LINKED ISSUE #562: [Refactor] use emit_json helper for JSON output in view commands
## Description

The `emit_json` helper in `helpers.py` uses `click.echo` for clean machine-readable JSON output
without Rich styling. However, `view.py` bypasses it and uses `console.print(json_mod.dumps(...))`
directly in three commands:

- `issues_list` (lines 86, 88)
- `issues_pending_harvest` (lines 267-279)
- `admin_info` (line 314)

`console.print` goes through Rich, which can inject ANSI escape codes or modify output encoding
depending on terminal settings. This breaks JSON parsers consuming stdout (e.g. `gitt issues list --json | jq`).

The miner commands (check.py, post.py) and `emit_error_json` already use `click.echo` for JSON.
The view commands should use the existing `emit_json` helper for consistency.

HadesHappy:
## Summary

- Replace `console.print(json_mod.dumps(...))` with `emit_json(...)` in view.py
- Covers `issues_list`, `issues_pending_harvest`, and `admin_info` JSON paths
- Ensures JSON output goes through `click.echo` (no Rich styling) for clean pipe consumption

## Problem

`view.py` uses `console.print(json_mod.dumps(...))` for JSON output while the codebase already
has an `emit_json` helper that uses `click.echo` for clean, machine-readable output. Rich's
`console.print` can inject ANSI escape codes in some terminal configurations, breaking downstream
JSON parsers.

## Test plan

- [ ] `gitt issues list --json` outputs valid JSON without ANSI codes
- [ ] `gitt issues list --json | jq .` parses correctly
- [ ] All existing tests pass

Fixes #562

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -204,12 +204,12 @@ def emit_error_json(message: str, error_type: str = 'cli_error') -> None:
     click.echo(json.dumps(payload), err=False)
 
 
-def emit_json(payload: Any, pretty: bool = True) -> None:
+def emit_json(payload: Any, pretty: bool = True, default: Optional[Callable] = str) -> None:
     """Emit a machine-readable JSON payload to stdout."""
     if pretty:
-        output = json.dumps(payload, indent=2, ensure_ascii=False)
+        output = json.dumps(payload, indent=2, ensure_ascii=False, default=default)
     else:
-        output = json.dumps(payload, separators=(',', ':'), ensure_ascii=False)
+        output = json.dumps(payload, separators=(',', ':'), ensure_ascii=False, default=default)
     click.echo(output, err=False)
 
 
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -11,7 +11,6 @@ Commands:
     gitt admin info
 """
 
-import json as json_mod
 from decimal import Decimal
 
 import click
@@ -27,6 +26,7 @@ from .helpers import (
     colorize_status,
     console,
     emit_error_json,
+    emit_json,
     format_alpha,
     print_error,
     print_network_header,
@@ -83,9 +83,9 @@ def issues_list(issue_id: int, network: str, rpc_url: str, contract: str, verbos
             if issue is None:
                 emit_error_json(f'Issue {issue_id} not found on-chain.', error_type='not_found')
                 raise SystemExit(1)
-            console.print(json_mod.dumps(issue, indent=2, default=str))
+            emit_json(issue)
         else:
-            console.print(json_mod.dumps(issues, indent=2, default=str))
+            emit_json(issues)
         return
 
     # Single issue detail view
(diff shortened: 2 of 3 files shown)


anderdc (changes requested):
Good catch on the consistency/ANSI issue — issue #562 is valid and this PR's direction is right. Requesting changes on one regression risk and some verification before merge.

**1. `default=str` is lost in the conversion — latent `TypeError` risk in 3 of the 4 call sites.**

The original calls at `view.py` passed `default=str` to `json.dumps`, which allows serializing non-native types (Decimal, datetime, UUID, etc.) via `str()` coercion:

- `console.print(json_mod.dumps(issue, indent=2, default=str))` (single-issue path)
- `console.print(json_mod.dumps(issues, indent=2, default=str))` (list path)
- `console.print(json_mod.dumps(packed, indent=2, default=str))` (admin_info)

`emit_json` in `helpers.py:207-213` does **not** accept or apply `default=str`, so after this PR any Decimal/datetime/UUID in `issue`, `issues`, or `packed` will raise `TypeError: Object of type X is not JSON serializable`. The `from decimal import Decimal` import at the top of `view.py` is a strong signal that Decimals actually flow through these paths.

The two call sites that originally omitted `default=str` (`issues_bounty_pool`, `issues_pending_harvest`) are safe — they build dicts from ints and pre-formatted strings. No type issue there.

**Fix: extend `emit_json` signature to accept a `default` callable**, defaulting to `str` to match the prior behavior at the risky call sites:

```python
def emit_json(payload: Any, pretty: bool = True, default: Optional[Callable] = str) -> None:
    """Emit a machine-readable JSON payload to stdout."""
    if pretty:
        output = json.dumps(payload, indent=2, ensure_ascii=False, default=default)
    else:
        output = json.dumps(payload, separators=(',', ':'), ensure_ascii=False, default=default)
    click.echo(output, err=False)
```

Two lines in `helpers.py`, no call-site changes needed in this PR — the default covers all four migrated sites.

**2. Secondary behavior change worth flagging: `ensure_ascii`.**

Original calls used `json.dumps` default (`ensure_ascii=True`, escapes non-ASCII as `\uXXXX`). `emit_json` passes `ensure_ascii=False`, so Unicode in issue titles/descriptions now emits as raw UTF-8. Strictly better for machine consumers (smaller, still valid JSON), but it's an observable output diff. Worth noting in the PR description so downstream consumers aren't surprised. Not blocking.

**3. Actually run the test plan.**

The PR's own checklist is all unchecked:

- [ ] `gitt issues list --json` outputs valid JSON without ANSI codes
- [ ] `gitt issues list --json | jq .` parses correctly
- [ ] All existing tests pass

For a CLI change with a serialization regression risk, the Decimal path specifically needs exercising — run `gitt issues list --json` against a real or fixtured contract that returns Decimal values and confirm it doesn't raise. Also please run the existing pytest suite before marking this ready.

**4. Add a regression test.**

One-line parametrized test in `tests/cli/` covering `emit_json` with Decimal and datetime inputs, asserting it produces valid JSON without raising. Small addition, catches the exact failure mode (1) prevents.

HadesHappy:
Hi, @anderdc 

Addressed all review feedback:

1. **`default=str` restored** - `emit_json` now accepts a `default` parameter (defaults to `str`) so Decimal, datetime, and other non-native types serialize correctly. No call-site changes needed.

2. **`ensure_ascii` behavior noted** - understood, output now emits raw UTF-8 instead of `\uXXXX` escapes. Strictly better for machine consumers.

3. **Regression test added** - parametrized test in `test_cli_helpers.py` covering Decimal, datetime, and date inputs through `emit_json`.

4. **CI executed:**
   - pytest: 366 passed
   - ruff check: all passed
   - ruff format: all formatted
   - pyright: 0 errors

Could you review again please?

anderdc (approved):

PR #569, opened 2026-04-18
refactor: deduplicate identical blacklist and priority handlers in pat_handler

### LINKED ISSUE #568: [Refactor] deduplicate identical blacklist and priority handlers in pat_handler
## Description

`pat_handler.py` has two pairs of completely identical functions:

### blacklist handlers (lines 89-93 and 121-125)

    async def blacklist_pat_broadcast(validator, synapse):
        hotkey = _get_hotkey(synapse)
            return True, f'Hotkey {hotkey[:16]}... not registered'
        return False, 'Hotkey recognized'

    async def blacklist_pat_check(validator, synapse):
        # exact same body

### priority handlers (lines 96-101 and 128-133)

    async def priority_pat_broadcast(validator, synapse):
        hotkey = _get_hotkey(synapse)
            return 0.0
        return float(validator.metagraph.S[uid])

    async def priority_pat_check(validator, synapse):
        # exact same body

The bodies don't use the synapse type — they only call `_get_hotkey(synapse)` which
accepts any `bt.Synapse`. Extract shared helpers and have the typed wrappers delegate.

HadesHappy:
## Summary

- Extract `_blacklist_unregistered()` from identical blacklist_pat_broadcast / blacklist_pat_check
- Extract `_priority_by_stake()` from identical priority_pat_broadcast / priority_pat_check
- Typed wrappers delegate to shared helpers

## Problem

`blacklist_pat_broadcast` and `blacklist_pat_check` have identical bodies.
`priority_pat_broadcast` and `priority_pat_check` have identical bodies.
Both pairs only use `_get_hotkey(synapse)` which accepts any `bt.Synapse`.

## Test plan

- [ ] All existing PAT handler tests pass
- [ ] Full test suite passes
- [ ] `ruff check` and `ruff format` clean

Fixes #568

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -78,19 +78,12 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
 
 async def blacklist_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSynapse) -> Tuple[bool, str]:
     """Reject PAT broadcasts from unregistered hotkeys."""
-    hotkey = _get_hotkey(synapse)
-    if hotkey not in validator.metagraph.hotkeys:
-        return True, f'Hotkey {hotkey[:16]}... not registered'
-    return False, 'Hotkey recognized'
+    return await _blacklist_unregistered(validator, synapse)
 
 
 async def priority_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSynapse) -> float:
     """Prioritize PAT broadcasts by stake."""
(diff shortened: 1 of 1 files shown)


anderdc:
Closing along with #568. Full reasoning in the issue comment.

TL;DR: the axon framework's signature-equality check forces two typed wrappers regardless of what we do internally, so extracting a shared helper only shuffles duplication around rather than eliminating it. Each body is 3-4 lines of stable code, and adding indirection for that is hairsplitting. Not worth the review cost.

Appreciate the work — the analysis was useful even though we're not taking the change.

PR #597, opened 2026-04-19
fix: remove penalized UIDs from cached_uids so duplicate penalty persists to DB

### LINKED ISSUE #596: [BUG] Duplicate-account penalty skipped for cached miners - zeroed score never persisted to DB
## Description

`get_rewards` in `reward.py` runs cache restoration before duplicate detection:

    cached_uids = self.store_or_use_cached_evaluation(miner_evaluations)  # line 122
    detect_and_penalize_miners_sharing_github(miner_evaluations)           # line 125

When a miner's GitHub API fails AND they share a GitHub account with another UID:

1. `store_or_use_cached_evaluation` restores their old valid score from cache, adds UID to `cached_uids`
2. `detect_and_penalize_miners_sharing_github` zeroes the evaluation in-memory
3. `bulk_store_evaluation` sees UID in `cached_uids` → skips DB write

The penalty (zeroed score) is never written to DB. Next cycle, if GitHub API fails again,
the old positive score is restored from cache. The duplicate miner keeps earning indefinitely.

## Steps to Reproduce

1. Two UIDs register the same GitHub account
2. GitHub API is intermittently unavailable for one of them
3. Duplicate detection zeroes their in-memory evaluation
4. DB write is skipped because UID is in cached_uids
5. Next cycle: cache returns the old positive score again

## Expected Behavior

Duplicate penalty should always be persisted to DB. Penalized UIDs should be removed from cached_uids so the zeroed score is written.

## Actual Behavior

Penalized UIDs remain in cached_uids. The zero score is never written to DB. The old
positive score persists in both cache and DB across cycles.

HadesHappy:
## Summary

- After duplicate detection, remove penalized UIDs from `cached_uids` so the zeroed
  evaluation is written to DB by `bulk_store_evaluation`
- Prevents duplicate miners from retaining old positive scores across cycles when
  GitHub API is intermittently unavailable

Closes #596

## Problem

`store_or_use_cached_evaluation` adds UIDs to `cached_uids` (skip DB write) before
`detect_and_penalize_miners_sharing_github` zeroes their scores. The penalty is never
persisted - the old positive score stays in DB and cache indefinitely.

## Test plan

- [ ] Verify penalized UIDs are removed from cached_uids
- [ ] Verify zeroed evaluation is written to DB after duplicate detection
- [ ] All existing tests pass

### DIFF
--- a/gittensor/validator/oss_contributions/inspections.py
+++ b/gittensor/validator/oss_contributions/inspections.py
@@ -11,7 +11,7 @@ from gittensor.constants import RECYCLE_UID
 from gittensor.validator.utils.github_validation import validate_github_credentials
 
 
-def detect_and_penalize_miners_sharing_github(miner_evaluations: Dict[int, MinerEvaluation]):
+def detect_and_penalize_miners_sharing_github(miner_evaluations: Dict[int, MinerEvaluation]) -> set[int]:
     """
     Detects miners that used the same github, duplicated across multiple uids.
     Will then penalize detected 'duplicate miners' with a score of 0.0.
@@ -19,6 +19,9 @@ def detect_and_penalize_miners_sharing_github(miner_evaluations: Dict[int, Miner
 
     Args:
         miner_evaluations (Dict[int, MinerEvaluation]): Mapping of miner UID to their MinerEvaluation.
+
+    Returns:
+        Set of penalized UIDs.
     """
 
     bt.logging.info('Now checking for duplicate users across miners...')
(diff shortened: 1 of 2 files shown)


HadesHappy:
Hi, @anderdc The Conflicts has been resolved.
Could you review the PR please?

anderdc:
Closing in favor of #600, which takes the same explicit-signal approach but fits the codebase slightly better on two points:

- **Style consistency with the file.** `inspections.py` already uses `from typing import Dict, List, Optional`. #600 extends that same import line with `Set`, keeping the file on a single typing convention. This PR introduces `set[int]` (PEP 585 builtin) alongside the existing `typing` imports, which mixes two conventions in the same file.
- **Better caller-side comment.** #600 updates the `reward.py` comment to explain *why* the subtraction happens (`returns penalized UIDs so they are removed from cached_uids and their zeroed scores are written to DB`). This PR keeps the original terse comment.

Both approaches are otherwise structurally identical and equally correct. Genuinely close call — picking one to avoid parallel review churn. Appreciate the work.

PR #598, opened 2026-04-19
fix: remove penalized UIDs from cached_uids so zeroed score is written to DB

### LINKED ISSUE #596: [BUG] Duplicate-account penalty skipped for cached miners - zeroed score never persisted to DB

Tet-9:
Fixes #596

## Problem
`get_rewards` in `reward.py` runs cache restoration before duplicate
detection:

1. `store_or_use_cached_evaluation` restores cached eval for a miner
   whose GitHub API failed, and adds their UID to `cached_uids`
2. `detect_and_penalize_miners_sharing_github` then zeroes their score
   in-memory
3. `bulk_store_evaluation` sees the UID in `cached_uids` and skips the
   DB write

The zeroed penalty score is never persisted. Next cycle, cache returns
the old positive score and the duplicate miner keeps earning.

## Fix
After `detect_and_penalize_miners_sharing_github` runs, remove any
penalized UIDs from `cached_uids` so `bulk_store_evaluation` writes
their zeroed score to DB.

A UID is considered penalized when `total_score == 0.0` and
`failed_reason is None` (genuine penalty, not a validation failure).

## Changes
- `gittensor/validator/oss_contributions/reward.py` — remove penalized
  UIDs from `cached_uids` after duplicate detection

### DIFF
--- a/gittensor/validator/oss_contributions/reward.py
+++ b/gittensor/validator/oss_contributions/reward.py
@@ -126,7 +126,8 @@ async def get_rewards(
 
     # Remove penalized UIDs from cached_uids so their zeroed score is written to DB
     penalized_uids = {
-        uid for uid, evaluation in miner_evaluations.items()
+        uid
+        for uid, evaluation in miner_evaluations.items()
         if evaluation.total_score == 0.0 and evaluation.failed_reason is None and uid in cached_uids
     }
     cached_uids -= penalized_uids
--- a/gittensor/validator/oss_contributions/reward.py
+++ b/gittensor/validator/oss_contributions/reward.py
@@ -124,6 +124,13 @@ async def get_rewards(
     # Adjust scores for duplicate accounts
     detect_and_penalize_miners_sharing_github(miner_evaluations)
 
+    # Remove penalized UIDs from cached_uids so their zeroed score is written to DB
+    penalized_uids = {
+        uid for uid, evaluation in miner_evaluations.items()
+        if evaluation.total_score == 0.0 and evaluation.failed_reason is None and uid in cached_uids
+    }
+    cached_uids -= penalized_uids
+
     # Finalize scores: apply eligibility gate, credibility, pioneer dividends, collateral
     finalize_miner_scores(miner_evaluations)
 


anderdc:
Closing in favor of #600.

The core concern with this PR's approach is the **heuristic inference** of penalized UIDs:

```python
penalized_uids = {
    uid
    for uid, evaluation in miner_evaluations.items()
    if evaluation.total_score == 0.0 and evaluation.failed_reason is None and uid in cached_uids
}
```

This infers "penalized" from `total_score == 0.0` rather than receiving an explicit signal from the penalizer. Problems:

- **Fragile semantic coupling.** A cached eval that happens to have `total_score == 0.0` from a prior cycle would get flagged as "penalized this cycle" even when duplicate detection didn't touch it. Harmless today but relies on "penalized implies zero" being the only zero-producing code path, which isn't guaranteed as the scoring logic evolves.
- **Order-dependent.** Has to run exactly between duplicate detection and DB write. If `finalize_miner_scores` is ever moved before this inference (or applies its own zero-out conditions), the detection breaks silently.
- **Discards available information.** `detect_and_penalize_miners_sharing_github` already tracks which UIDs it zeroed internally (line 44 logs `PENALTY: Zeroing score for duplicate uid {uid}`). Having it return that set directly is strictly better than reconstructing it from side effects.

#600 (and #597) take the explicit-signal approach: modify the penalizer to return the set of UIDs it zeroed, caller subtracts that set from `cached_uids`. Can't desynchronize from the penalizer's behavior. #600 is the one getting merged.

Appreciate the work — the diagnosis of the bug was correct, just preferring the cleaner propagation shape.

PR #599, opened 2026-04-19
refactor: deduplicate blacklist and priority handlers in pat_handler

### LINKED ISSUE #568: [Refactor] deduplicate identical blacklist and priority handlers in pat_handler

Tet-9:
Fixes #568

## Problem
`pat_handler.py` had two pairs of completely identical functions:
- `blacklist_pat_broadcast` and `blacklist_pat_check` — exact same body
- `priority_pat_broadcast` and `priority_pat_check` — exact same body

Both pairs only call `_get_hotkey(synapse)` which accepts any `bt.Synapse`,
so the synapse type was irrelevant to the logic.

## Solution
Extracted two shared async helpers:
- `_blacklist_handler(validator, synapse)` — shared blacklist logic
- `_priority_handler(validator, synapse)` — shared priority logic

The four typed wrappers now delegate to these helpers, eliminating
the duplicated logic while preserving the public API.

## Changes
- `gittensor/validator/pat_handler.py` — extracted 2 helpers, simplified 4 functions

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -78,19 +78,12 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
 
 async def blacklist_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSynapse) -> Tuple[bool, str]:
     """Reject PAT broadcasts from unregistered hotkeys."""
-    hotkey = _get_hotkey(synapse)
-    if hotkey not in validator.metagraph.hotkeys:
-        return True, f'Hotkey {hotkey[:16]}... not registered'
-    return False, 'Hotkey recognized'
+    return await _blacklist_handler(validator, synapse)
 
 
 async def priority_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSynapse) -> float:
     """Prioritize PAT broadcasts by stake."""
-    hotkey = _get_hotkey(synapse)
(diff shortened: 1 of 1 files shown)


anderdc:
Closing along with #568. Full reasoning in the issue comment.

TL;DR: the axon framework's signature-equality check forces two typed wrappers regardless of what we do internally, so extracting a shared helper only shuffles duplication around rather than eliminating it. Each body is 3-4 lines of stable code, and adding indirection for that is hairsplitting. Not worth the review cost.

Appreciate the work — the analysis was useful even though we're not taking the change.

PR #613, opened 2026-04-19
fix: include label_multiplier in open PR collateral score

### LINKED ISSUE #612: Open PR collateral score ignores label_multiplier
## Description

When a PR is open, the validator computes a "collateral score" using a few multipliers (`repo_weight`, `issue`). But `label_multiplier` is missing from that formula, even though it's already set on the PR and gets applied correctly once the PR merges.

This means two open PRs in the same repo with the same base score get the same collateral - even if one is labeled `feature` (1.5×) and the other has no label. The label only starts counting the moment the PR merges, which seems wrong. Labels are available from the start.

## Steps to Reproduce

1. Open two PRs in the same tracked repo with equal base scores, repo weight, and issue links.
2. Label one `feature`, leave the other unlabeled.
3. Run the validator and check the collateral scores assigned to each PR.

## Expected Behavior

The `feature` PR's collateral should be higher than the unlabeled PR's collateral, mirroring what happens after merge (where `feature` gets a 1.5× boost).

## Actual Behavior

Both PRs get identical collateral scores. The `label_multiplier` is silently dropped from the collateral formula.

## Environment

- OS:
- Python version:
- Commit/Version:

nightmare0329:
Fixes #612

`calculate_open_pr_collateral_score` applied `repo_weight` and `issue` multipliers but omitted `label_multiplier`. Labels are already set on PRs before merge, so a `feature`-labeled PR should carry higher collateral than an unlabeled one with the same base score, mirroring what happens after merge.

Changes:
- `gittensor/validator/oss_contributions/scoring.py` — add `label` to the multipliers dict in `calculate_open_pr_collateral_score`

### DIFF
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -506,7 +506,7 @@ def calculate_open_pr_collateral_score(pr: PullRequest) -> float:
 
     Collateral = base_score * applicable_multipliers * OPEN_PR_COLLATERAL_PERCENT
 
-    Applicable multipliers: repo_weight, issue
+    Applicable multipliers: repo_weight, issue, label
     NOT applicable: time_decay (merge-based), credibility_multiplier (merge-based),
                     open_pr_spam (not for collateral)
     """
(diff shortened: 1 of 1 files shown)


nightmare0329:
@anderdc ready for review.

anderdc (approved):

PR #570, opened 2026-04-18
fix: guard handle_pat_check against unregistered hotkeys and use descriptive errors

nightmare0329:
## Summary

- `handle_pat_check` now checks metagraph membership before calling `.index()`, preventing a `ValueError` crash when a hotkey disappears between the blacklist check and handler invocation (race condition on deregistration)
- `_get_hotkey` replaces a bare `assert` with a `ValueError` carrying a meaningful message — `AssertionError` is opaque and is silently stripped when Python runs with `-O`
- Fix `gitt --version` displaying `3.2.0` instead of the actual package version `5.0.0`

## Related Issues

N/A

## Type of Change

- [x] Bug fix

## Testing

- Added `TestHandlePatCheck::test_unregistered_hotkey_rejected` covering the new guard
- All 16 PAT handler tests pass

## Checklist

- [x] Code follows existing style conventions
- [x] Self-reviewed the diff
- [x] No documentation changes required

PR #614, opened 2026-04-19
fix: guard handle_pat_check against unregistered hotkey

### LINKED ISSUE #574: [BUG] handle_pat_check crashes with ValueError on deregistered hotkey (missing membership guard)
## Description

`handle_pat_check` in `pat_handler.py` line 104 calls:

without first verifying the hotkey is in the list. If the metagraph resyncs between
the blacklist check and handler execution (race condition), a previously-registered
hotkey becomes unregistered and `.index()` raises `ValueError`.

`handle_pat_broadcast` (line 46-49) correctly guards this:

        return _reject('Hotkey not registered on subnet')

`handle_pat_check` is missing the same guard.

## Steps to Reproduce

1. A miner sends a PAT check request
2. `blacklist_pat_check` passes (hotkey is registered)
3. Between blacklist and handler, the validator resyncs the metagraph (new epoch)
4. The miner's hotkey is no longer in the metagraph (deregistered or UID changed)
5. `handle_pat_check` calls `.index(hotkey)` → `ValueError`

## Expected Behavior

`handle_pat_check` should check hotkey membership before calling `.index()`,
matching the pattern in `handle_pat_broadcast`.

nightmare0329:
Fixes #574

`handle_pat_check` called `.index(hotkey)` without first checking membership, raising `ValueError` if the metagraph resyncs between the blacklist check and handler execution. `handle_pat_broadcast` already guards this correctly; apply the same pattern.

Changes:
- `gittensor/validator/pat_handler.py` — add membership check before `.index(hotkey)` in `handle_pat_check`

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -101,6 +101,12 @@ async def priority_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSy
 async def handle_pat_check(validator: 'Validator', synapse: PatCheckSynapse) -> PatCheckSynapse:
     """Check if the validator has the miner's PAT stored and re-validate it."""
     hotkey = _get_hotkey(synapse)
+    if hotkey not in validator.metagraph.hotkeys:
+        synapse.has_pat = False
+        synapse.pat_valid = False
+        synapse.rejection_reason = 'Hotkey not registered on subnet'
+        bt.logging.warning(f'PAT check rejected — hotkey {hotkey[:16]}... not in metagraph')
+        return synapse
     uid = validator.metagraph.hotkeys.index(hotkey)
     entry = pat_storage.get_pat_by_uid(uid)
 


nightmare0329:
@anderdc ready for review.

anderdc:
Closing as a duplicate of prior closed work and because the underlying concern is already handled by the existing architecture.

**Prior art (all closed):**

- Issue #574 was closed as `not planned` earlier today
- PR #570 (bundled variant) was closed
- PR #575 (focused variant of the same fix) was closed

**Why the guard is redundant:**

- `blacklist_pat_check` at `pat_handler.py:139-144` already checks `hotkey in validator.metagraph.hotkeys` before any handler runs. The bittensor axon pipeline (`axon.py:1162-1172`) only invokes the handler after the blacklist passes, so by design the handler runs only for registered hotkeys.
- If a race did somehow occur, the axon middleware has a blanket exception handler (`axon.py:1189-1191`) that converts any handler exception into an error response. The validator does **not** crash on a `ValueError` from `.index()`.
- The axon dispatch for a single request is one coroutine with no `await` point between `blacklist()` and `run()` that yields control to the validator's metagraph resync loop in a way that would swap `metagraph.hotkeys` mid-request.

**Stylistic parity with `handle_pat_broadcast`:**

- The guard there is not defensive against a race — it is required because that handler performs a PAT validation flow that costs GitHub API calls. Short-circuiting unregistered hotkeys early avoids wasted work. `handle_pat_check` has the same early filtering already done at the blacklist layer.

No functional bug, no crash risk, and the fix was already declined twice. Closing.

PR #615, opened 2026-04-19
fix: sort repo_contributions iteration in calculate_pioneer_dividends

### LINKED ISSUE #538: Pioneer dividend is non-deterministic — dict iteration order changes earned_score across repos
## Description

`calculate_pioneer_dividends` in `scoring.py:297-334` iterates over `repo_contributions.items()` and mutates `pioneer_pr.earned_score` in-place on line 334:

```python
pioneer_pr.earned_score = round(pioneer_pr.earned_score + pioneer_pr.pioneer_dividend, 2)
```

If the same miner is a pioneer in repo A and a follower in repo B, the processing order matters. When repo A is processed first, the pioneer's `earned_score` gets inflated. When repo B is processed later, that inflated `earned_score` feeds into the follower's `total_score` calculation at line 302:

```python
new_total = total_score + pr.earned_score
```

Dict iteration order in Python is insertion-order since 3.7 but differs across processes if hash randomization is enabled (default). Two validators processing the same data could compute different dividends.

## Impact

Consensus divergence — validators may disagree on final miner scores depending on which repo gets processed first in the pioneer dividend loop.

## Suggested Fix

Sort `repo_contributions.items()` by repo name before iterating, or compute dividends in a separate pass without mutating `earned_score` in-place.

nightmare0329:
Fixes #538

Dict iteration order depends on insertion order, which varies across validator runs (it follows the order miners's PRs are processed). When the same miner is a pioneer in repo A and a follower in repo B, processing A first inflates their `earned_score` before it feeds into B's follower tally. Sort by repo name to make the output deterministic.

Changes:
- `gittensor/validator/oss_contributions/scoring.py` — `sorted(repo_contributions.items())` in `calculate_pioneer_dividends`

### DIFF
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -305,7 +305,7 @@ def calculate_pioneer_dividends(
                 else:
                     repo_contributions[repo][pr.uid] = (earliest_at, earliest_num, new_total)
 
-    for repo, uid_entries in repo_contributions.items():
+    for repo, uid_entries in sorted(repo_contributions.items()):
         sorted_uids = sorted(uid_entries.items(), key=lambda x: (x[1][0], x[1][1]))
 
         for rank_pos, (uid, _) in enumerate(sorted_uids):


nightmare0329:
@anderdc ready for review.

anderdc:
Closing along with #538 — the underlying bug does not exist, so this fix is not needed.

`calculate_pioneer_dividends` runs in **two sequential passes**:

- **Pass 1** builds `repo_contributions` by reading `pr.earned_score` once per PR and storing an **immutable snapshot tuple** `(merged_at, number, total_score)`.
- **Pass 2** iterates `repo_contributions.items()` and mutates `pioneer_pr.earned_score`. The follower score used for dividend math is `follower_earned = entry[2]`, read from the **Pass 1 snapshot tuple**, not from the live `pr.earned_score`.

As a result:

- Pass 1 completes entirely before any mutation happens, so iteration order there cannot affect `total_score` accumulation.
- In Pass 2, cross-iteration reads only touch the frozen tuple values — mutating repo A's pioneer cannot influence repo B's follower tally.
- Sorting `repo_contributions.items()` is a no-op in terms of computed output; it only changes log ordering.

No consensus divergence risk, so the sort is unnecessary. Thanks for the investigation, but closing as invalid.

PR #211, opened 2026-02-18
feat(cli): harden and polish issue commands

MkDev11:
## Summary

Hardens and polishes the `gitt` CLI across all issue commands to make it production-grade. This PR addresses the issue: input validation that catches bad data before it hits the chain, a precision fix for bounty amounts (`--bounty` is parsed as a string and converted via `Decimal`, so no IEEE 754 float is ever created), consistent formatting with a shared `format_alpha()` helper (Decimal-based), non-interactive register support, and correct exit codes so scripts and CI can detect failures. Invalid input and missing contract config now raise `click.ClickException` and exit non-zero instead of printing and returning 0.

The goal is to make `gitt` solid and scriptable — no silent precision bugs, no wasted gas on invalid addresses, and reliable exit codes for automation.

This PR ensures **non-zero exit codes** on all validation and config errors (so CI and scripts can fail fast) and adds **66 automated tests**. SS58 uses the **existing substrate stack** (no extra deps). **Closed-issue handling** follows the issue spec: warn with "Issue #{number} is already closed." and continue (no reject).

---

## Task-by-task (issue #210)

| # | Task | Done |
|---|------|------|
| 1 | **Bounty precision** — `--bounty` as string → `Decimal` only. Min 10, max 100M ALPHA (u128-safe), max 9 decimals, `isfinite`. | ✅ |
| 2 | **Repo validation** — Regex `owner/repo`; optional GitHub API existence check before register. | ✅ |
| 3 | **GitHub issue** — Exists, is issue not PR; closed → warning. Issue number capped at u32 max (2³²−1). | ✅ |
| 4 | **On-chain ID & SS58** — Issue ID 1–999999 (vote/admin). SS58 via `substrateinterface.utils.ss58` + regex fallback. PR ≥ 1, number or URL; invalid → `param_hint`. | ✅ |
| 5 | **Shared formatting** — `format_alpha(raw, decimals)` + constants. All validators set `param_hint`. | ✅ |
| 6 | **Deprecated helper cleanup** — Remove `get_ws_endpoint()` and unused code. | ✅ N/A — `get_ws_endpoint()` is not present in the codebase; only `resolve_network()` is used. |
[...]

PR #626, opened 2026-04-20
refactor: extract _github_api_get helper to dedupe urllib plumbing

plind-junior:
## Summary
- Extract shared urllib + error-handling plumbing from `validate_repository` and `validate_github_issue` into a private `_github_api_get(url, check_name) -> Optional[dict]` helper.
- Use a private `_GitHubNotFound` sentinel so each caller keeps its own `click.BadParameter` message and `param_hint` on 404.
- Behavior unchanged: 404 → context-specific `BadParameter`; other HTTP/network errors → warn and return `None`.

## Test plan
- [x] `uv run pytest tests/cli/test_cli_helpers.py` (69 passed)
- [ ] Manual: `gitt issue register --repo <invalid/repo>` → still rejects with "not found" message
- [ ] Manual: `gitt issue register --issue <missing_number>` → still rejects with "Issue #N not found" message

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -347,6 +347,30 @@ def validate_bounty_amount(bounty: str) -> int:
     return raw
 
 
+class _GitHubNotFound(Exception):
+    """Signals a 404 from _github_api_get so callers can raise context-specific errors."""
+
+
+def _github_api_get(url: str, check_name: str) -> Optional[Dict[str, Any]]:
+    """GET a GitHub API URL and return parsed JSON.
+
+    Raises _GitHubNotFound on 404. Warns and returns None on other HTTP errors
+    or network failures.
+    """
+    try:
+        req = urllib.request.Request(url, headers={'User-Agent': 'gittensor-cli'})
+        resp = urllib.request.urlopen(req, timeout=GITHUB_API_TIMEOUT)
+        return json.loads(resp.read().decode())
+    except urllib.error.HTTPError as e:
+        if e.code == 404:
+            raise _GitHubNotFound() from e
+        console.print(f'[yellow]Warning: GitHub API returned {e.code} — skipping {check_name}[/yellow]')
+        return None
+    except (urllib.error.URLError, OSError):
+        console.print(f'[yellow]Warning: Could not reach GitHub API — skipping {check_name}[/yellow]')
+        return None
+
+
 def validate_repository(repo: str, verify_exists: bool = True) -> Tuple[str, str]:
     """Validate owner/repo format and optionally verify it exists on GitHub.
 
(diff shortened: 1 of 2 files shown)


anderdc (changes requested):
Appreciate the effort here — the duplication is real at a narrow level. But zooming out reveals the dedup is targeting the wrong pattern, so requesting a pivot rather than landing this as-is.

**The underlying issue: the codebase doesn't actually have a "urllib pattern" worth consolidating.**

Every other GitHub API caller in the codebase uses `requests`:

- `gittensor/utils/github_api_tools.py` (validator PR fetching, GraphQL)
- `gittensor/validator/pat_handler.py` (PAT validation)
- `gittensor/validator/issue_discovery/repo_scan.py`
- `gittensor/cli/miner_commands/post.py`

The `urllib` usage in `cli/issue_commands/helpers.py` is a single-file holdover from PR #211 when `validate_repository` and `validate_github_issue` were added. No principled reason for it — `requests` is already a required dependency of the project, so there's no dep-minimization argument, and the two urllib callers here have no special requirements.

Adding `_github_api_get` as a urllib-wrapping helper:

- Dedupes two call sites that were never going to grow
- Introduces a **third pattern** for GitHub API access (urllib direct, requests direct, and now `_github_api_get` wrapping urllib) instead of unifying
- Keeps the urllib island permanently quarantined from the rest of the codebase's HTTP calls

**Requested pivot: convert `validate_repository` and `validate_github_issue` to use `requests`.**

This is a slightly different scope but substantially more valuable — it removes the last urllib-based GitHub API caller in the codebase, unifying the HTTP pattern. Rough shape for each caller:

```python
import requests  # replace urllib imports

try:
    resp = requests.get(
        f'https://api.github.com/repos/{owner}/{repo_name}',
        headers={'User-Agent': 'gittensor-cli'},
        timeout=GITHUB_API_TIMEOUT,
    )
    if resp.status_code == 404:
        raise click.BadParameter(
            f"Repository '{owner}/{repo_name}' not found on GitHub",
            param_hint='--repo',
        )
    if not resp.ok:
        console.print(f'[yellow]Warning: GitHub API returned {resp.status_code} — skipping existence check[/yellow]')
except requests.RequestException:
    console.print('[yellow]Warning: Could not reach GitHub API — skipping existence check[/yellow]')
```

**And drop the `_github_api_get` / `_GitHubNotFound` helper entirely.** At 2 callers on a now-consistent HTTP library, a few lines of inline `requests` code is more readable than a local wrapper, and there's no ecosystem-wide dedup to accomplish without a much bigger refactor. Better to leave the small amount of duplication than add a local abstraction that doesn't address the real fragmentation.

This way:

- urllib is removed from the codebase entirely (one less pattern to maintain)
- Both callers match the HTTP convention used everywhere else
- If someone later wants real dedup, it can happen on a consistent `requests` substrate — possibly sharing with `github_api_tools.py`

Reuse the same branch and PR, no need to open a new one. Thanks again for the attention to duplication; redirecting the effort toward the bigger consistency win.

plind-junior:
Dropped `_github_api_get` and `_GitHubNotFound`
validate_repository / validate_github_issue now call requests.get inline (404 → BadParameter, other errors → warn + None)
urllib fully removed from the file
Updated the one test that patched urllib.request.urlopen → requests.get
pytest tests/cli/test_cli_helpers.py → 69 passed. Happy to reword the PR title away from "urllib plumbing" if you'd like.

@anderdc

anderdc:
Hi @plind-junior — thanks for running the test plan and the detailed write-up of what changed. One issue though: **I'm not seeing the changes you described in the PR diff.**

Looking at the commit history on this branch:

- `afcda00` (Apr 20 13:35) — original commit, "refactor: extract _github_api_get helper to dedupe urllib plumbing"
- Every commit since then is a "Merge branch 'test' into refactor/..." merge commit, not new code

The current diff (verified via both `gh pr diff 626` and the raw `/pulls/626/files` API) is byte-identical to the original:

- `_GitHubNotFound` exception class is still present
- `_github_api_get` helper is still present
- urllib imports are still there, and the helper still uses `urllib.request.Request` / `urllib.error.HTTPError`

My guess: you made the changes locally and ran the tests (the "69 passed" output is real from your local run), but the commit containing those changes didn't get pushed to the remote. Could you push it and confirm? Once I see the requests.get conversion and the dropped helper in the diff, I'll re-review right away.

plind-junior:
You're right @anderdc . The commit was not pushed

anderdc (approved):

PR #639, opened 2026-04-21
Unify _read_contract_u128/u32 in contract_client.py

### LINKED ISSUE #642: [BUG] Issue discovery credibility uses valid_solved_count instead of total solved_count - lowers credibility unfairly
## Description

`check_issue_eligibility` at `issue_discovery/scoring.py:62` receives `data.valid_solved_count`
as its first parameter (line 131) and uses it for BOTH:

1. The eligibility minimum check (correct — should use valid count)
2. The credibility formula via `calculate_issue_credibility` (incorrect — should use total solved)

The credibility formula per the docs:

    issue_credibility = solved_issues / (solved_issues + closed_issues - mulligan)

"Solved issues" means ALL issues resolved via merged PRs, not just those where the solving
PR has token_score >= 5. The OSS credibility implementation does this correctly — it passes
all merged PRs to credibility and separately counts valid ones for the minimum check.

## Impact

A miner with 10 solved issues (7 valid, 3 with low token_score PRs) and 2 closed issues:
- **Correct credibility**: 10 / (10 + max(0, 2-1)) = 10/11 = 0.91
- **Current credibility**: 7 / (7 + max(0, 2-1)) = 7/8 = 0.875

The miner's credibility multiplier is 0.875 instead of 0.91 — a 4% scoring penalty
for having solved issues that happened to involve small PRs.

## Root cause

Line 131 passes `data.valid_solved_count` where `data.solved_count` should be used for
the credibility calculation:

    is_eligible, credibility, reason = check_issue_eligibility(data.valid_solved_count, data.closed_count)

The OSS pattern separates these correctly (credibility.py:49-52):

    credibility = calculate_credibility(merged_prs, closed_prs)
    valid_merged_count = sum(1 for pr in merged_prs if pr.token_score >= MIN_TOKEN_SCORE_FOR_BASE_SCORE)

## Evidence

Image

statxc:
## Summary

`_read_contract_u128` and `_read_contract_u32` in `contract_client.py` were 8-line methods that differed only in which extractor they called. Replaced both with a single `_read_contract_numeric(method_name, extractor)` helper and updated the two call sites (`get_alpha_pool`, `get_last_harvest_block`) to pass their extractor explicitly. 
## Related Issues

None.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- [x] Tests added/updated
- [x] Manually tested

New tests in `tests/validator/test_contract_client_reads.py` cover the three helper paths (None response, extractor returns None, success) and assert that each public caller is wired to the correct method name **and** extractor — so a swap regression fails fast. All 417 tests pass; ruff/pyright clean.

## Checklist

- [x] Self-review completed

## Behavior equivalence

| Path | Before | After |
|---|---|---|
| `_raw_contract_read` returns `None` | `return 0` | `return 0` |
| Extractor returns `None` | `return 0` | `return 0` |
| Extractor returns `int` | `return int` | `return int` |

Public signatures (`get_alpha_pool`, `get_last_harvest_block`), RPC selector lookup, SCALE decoding, and `try/except` surface are all unchanged.

### DIFF
--- a/gittensor/validator/issue_competitions/contract_client.py
+++ b/gittensor/validator/issue_competitions/contract_client.py
@@ -9,7 +9,7 @@ import struct
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 import bittensor as bt
 from substrateinterface import Keypair
@@ -218,28 +218,17 @@ class IssueCompetitionContractClient:
     def get_alpha_pool(self) -> int:
         """Get the current alpha pool balance."""
         try:
-            value = self._read_contract_u128('get_alpha_pool')
-            return value
+            return self._read_contract_numeric('get_alpha_pool', self._extract_u128_from_response)
         except Exception as e:
             bt.logging.error(f'Error fetching alpha pool: {e}')
             return 0
 
-    def _read_contract_u128(self, method_name: str) -> int:
-        """Read a u128 value from a no-arg contract method."""
+    def _read_contract_numeric(self, method_name: str, extractor: Callable[[bytes], Optional[int]]) -> int:
+        """Read a numeric value from a no-arg contract method via ``extractor``."""
         response = self._raw_contract_read(method_name)
         if response is None:
             return 0
-
-        value = self._extract_u128_from_response(response)
-        return value if value is not None else 0
-
-    def _read_contract_u32(self, method_name: str) -> int:
-        """Read a u32 value from a no-arg contract method."""
-        response = self._raw_contract_read(method_name)
-        if response is None:
-            return 0
-
-        value = self._extract_u32_from_response(response)
+        value = extractor(response)
         return value if value is not None else 0
 
     def _raw_contract_read(self, method_name: str, args: dict = None) -> Optional[bytes]:  # type: ignore[assignment]
@@ -612,8 +601,7 @@ class IssueCompetitionContractClient:
     def get_last_harvest_block(self) -> int:
         """Query the block number of the last harvest."""
         try:
-            value = self._read_contract_u32('get_last_harvest_block')
-            return value
+            return self._read_contract_numeric('get_last_harvest_block', self._extract_u32_from_response)
         except Exception as e:
             bt.logging.error(f'Error fetching last harvest block: {e}')
             return 0
--- a/tests/validator/test_contract_client_reads.py
+++ b/tests/validator/test_contract_client_reads.py
@@ -1,0 +1,58 @@
+# Entrius 2025
+
+"""Tests for IssueCompetitionContractClient numeric read methods."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from gittensor.validator.issue_competitions.contract_client import (
+    IssueCompetitionContractClient,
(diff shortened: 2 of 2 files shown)


anderdc (changes requested):
Thanks for catching the inconsistency — the motivation is solid and the testing is thorough. Requesting a change on the envelope shape before merge.

**The issue: this PR adds `success: True` but the success-vs-error envelope becomes asymmetric.**

Compare:

- **Error shape** (pre-existing, via `emit_error_json`): `{success: False, error: {type, message}}` — 2 top-level keys; data nested under `error`
- **Success shape** (this PR): `{success: True, owner, treasury_hotkey, netuid, ...}` — many top-level keys; data flat alongside `success`

Consumers that want to uniformly check `payload.get('success')` then navigate to the actual data have to know the shape diverges: for errors they go to `payload['error']`, for success they go to the top level. That's the exact inconsistency this PR is trying to reduce, just shifted to a different axis.

**Request: nest the success data under a `data` key to mirror the error envelope.**

```python
emit_json({'success': True, 'data': packed})
```

Then consumers access `payload['data']['owner']` on success and `payload['error']['type']` on error. Envelope shape is uniform; only the payload key differs.

**Applies to these 4 call sites:**

- `issues list --json --id` (single-issue path)
- `issues bounty-pool --json`
- `issues pending-harvest --json`
- `admin info --json`

Tests need to be updated to assert `payload['data'][key]` instead of `payload[key]`, including the parametrized schema-lockdown in `test_view_json_success.py`.

**On the `issues list --json` (no --id) list path:**

Your existing test documents that it deliberately stays a bare list for backward compat. Two options:

1. **Leave it bare** — accept the asymmetry (list path has no envelope; single-issue and other paths do). Add a comment in the docstring noting this is an intentional backward-compat pin.
2. **Also wrap** it as `{success: True, data: [...]}` — cleanest long-term shape but is a breaking change for any existing consumer iterating the bare list.

I'd lean (1) for this PR to keep scope tight, and leave (2) as a potential follow-up when you're ready to do breaking changes across the CLI JSON surface intentionally. Your call.

**Everything else looks good:**

- Motivation is real (error paths already envelope via `success`; success paths should too)
- Testing is solid (parametrized schema-lockdown is exactly the right pattern for CLI JSON contract tests)
- Validation was actually run (382 passed, ruff + pyright clean) — appreciate the thoroughness

Once the envelope shape is consistent, happy to approve.

anderdc:
any reason this is a draft?

statxc:
Hi @anderdc could you review this please?

anderdc:
I don't review draft PRs

statxc:
@anderdc Ready for review. Please review again.

anderdc:
Closing as churn.

- **Marginal value**: collapsing two 8-line methods into one helper doesn't pay for itself when call sites become more verbose. `_read_contract_u128('get_alpha_pool')` was self-documenting; the new form pushes the encoding info onto every caller.
- **Only 2 call sites**, so the abstraction isn't earning its keep. Revisit when a third numeric reader (e.g. u64) actually lands.
- **Behavior-preserving refactors with no real simplification win** are churn, not improvement.

Tests were well-written, appreciate the effort, but this isn't worth merging as-is.

PR #645, opened 2026-04-21
fix: skip PR scoring on file-content GraphQL fetch failure

### LINKED ISSUE #644: [Bug] _fetch_file_contents_with_base_batch silently swallows GraphQL failures as base_score=0
## Description

`_fetch_file_contents_with_base_batch` is the only fetch path that silently converts GraphQL failures into a valid-looking result. When `execute_graphql_query` returns `None` (all 8 retries exhausted), the function returns `FileContentPair(None, None)` for every file in the batch. Downstream scoring treats this identically to "binary/unsupported file" and records `base_score = 0`.

The result: a transient API failure produces the same recorded score as a PR with no scorable code. The failure is invisible to the rest of the system.

Every other fetch failure path in the scoring pipeline — PR-list GraphQL fetch, JSON parse failure, missing user data, GraphQL errors — signals failure via `github_pr_fetch_failed`. This is the gap.

## Current Behavior

1. `execute_graphql_query` returns `None` after 8 retries
2. `_fetch_file_contents_with_base_batch` returns `{filename: FileContentPair(None, None)}` — no error signal
3. Scoring records `base_score = 0` for the PR
4. `github_pr_fetch_failed` remains `False`

## Expected Behavior

The failure should be distinguishable from a legitimate zero-score result. The affected PR should be skipped (contribute nothing) rather than record a false zero.

The behavior remains deterministic in the same sense the existing `if not file_changes: return` path is — a failed fetch always produces the same outcome (skip), not a random one. The only difference from the current behavior is that "skip" replaces "false zero."

## Proposed Fix Scope

**Per-PR only** — no miner-level abort, no cache fallback trigger, no effect on other PRs in the same cycle:

- `_fetch_file_contents_with_base_batch` / `fetch_file_contents_with_base` return a `fetch_failed` flag alongside results
- `score_pull_request` early-returns when file-content fetch failed (PR contributes nothing instead of a false zero)
- Other PRs continue scoring normally
- `github_pr_fetch_failed` is **not set** — no miner-level side effects, no cache rollback

This does not set a precedent for "skip on failure" proliferation: it mirrors the existing pattern where `get_pull_request_file_changes` returning empty already causes an early return. The fix simply extends that to distinguish "empty" from "failed."

## Environment

- Python: 3.12
- Branch: `test` at `83ca7f9`

mkdev5:
## Summary

Surfaces file-content batch fetch failures so scoring can distinguish "transient API failure" from "no scorable content." Per-PR scoped only: the affected PR is skipped (contributes nothing) instead of recording a false `base_score=0`. Other PRs in the same cycle continue scoring normally.

This addresses the maintainer's feedback on the previous iteration — no miner-level abort, no `github_pr_fetch_failed` signaling, no cache fallback trigger.

## Changes

- [_fetch_file_contents_with_base_batch](cci:1://file:///root/mkdev11/gold/gittensor/gittensor/utils/github_api_tools.py:1142:0-1230:25) returns `(results, fetch_failed)` tuple; flags `True` on GraphQL failure or missing `data.repository`
- [fetch_file_contents_with_base](cci:1://file:///root/mkdev11/gold/gittensor/gittensor/utils/github_api_tools.py:1233:0-1268:32) propagates `fetch_failed` across batches
- [fetch_file_contents_for_pr](cci:1://file:///root/mkdev11/gold/gittensor/gittensor/validator/oss_contributions/scoring.py:118:0-154:38) returns `(dict, fetch_failed)` tuple (no `miner_eval` param)
- [score_pull_request](cci:1://file:///root/mkdev11/gold/gittensor/gittensor/validator/oss_contributions/scoring.py:76:0-113:75) early-returns when file-content fetch failed (mirrors existing `if not file_changes: return` pattern)
- [get_pull_request_file_changes](cci:1://file:///root/mkdev11/gold/gittensor/gittensor/utils/github_api_tools.py:296:0-379:15) returns `None` on final failure (was `[]`) so callers can distinguish "no files" from "fetch failed"

## What this does NOT do

- Does **not** set `github_pr_fetch_failed`
- Does **not** abort scoring for other PRs
- Does **not** trigger cache fallback
- Does **not** add "skip on failure" branches to other fetch paths (issue, review, label, credibility)

## Related Issues

Closes #644

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -377,7 +377,7 @@ def get_pull_request_file_changes(repository: str, pr_number: int, token: str) -
     bt.logging.error(
         f'File changes request for PR #{pr_number} in {repository} failed after {max_attempts} attempts: {last_error}'
     )
-    return []
+    return None
 
 
 # GraphQL fragment used by both issue submissions and solver detection.
@@ -1136,6 +1136,10 @@ class FileContentPair:
     new_content: Optional[str]  # None for deleted files
 
 
+def _empty_file_content_pairs(batch_changes: List['FileChangeType']) -> Dict[str, FileContentPair]:
+    return {fc.filename: FileContentPair(None, None) for fc in batch_changes}
+
+
 def _fetch_file_contents_with_base_batch(
     repo_owner: str,
     repo_name: str,
@@ -1143,7 +1147,7 @@ def _fetch_file_contents_with_base_batch(
     head_sha: str,
     batch_changes: List['FileChangeType'],
     token: str,
-) -> Dict[str, FileContentPair]:
+) -> tuple[Dict[str, FileContentPair], bool]:
     """Fetch base and head file contents for a single batch of file changes.
 
     Args:
@@ -1155,7 +1159,7 @@ def _fetch_file_contents_with_base_batch(
         token: GitHub PAT for authentication
 
     Returns:
-        Dict mapping file paths to FileContentPair (old_content, new_content)
+        Tuple of (file-content dict, fetch_failed flag).
     """
     file_fields = []
     for i, fc in enumerate(batch_changes):
@@ -1178,7 +1182,7 @@ def _fetch_file_contents_with_base_batch(
             )
 
     if not file_fields:
-        return {}
+        return {}, False
 
     query = f"""
         query($owner: String!, $name: String!) {{
@@ -1193,12 +1197,17 @@ def _fetch_file_contents_with_base_batch(
     data = execute_graphql_query(query, variables, token)
     if data is None:
         bt.logging.warning(f'Failed to fetch file contents for {repo_owner}/{repo_name}')
-        return {fc.filename: FileContentPair(None, None) for fc in batch_changes}
+        return _empty_file_content_pairs(batch_changes), True
 
     if 'errors' in data:
         bt.logging.warning(f'GraphQL errors fetching files: {data["errors"]}')
 
-    repo_data = data.get('data', {}).get('repository', {})
+    data_payload = data.get('data')
+    repo_data = data_payload.get('repository') if isinstance(data_payload, dict) else None
+    if repo_data is None:
+        bt.logging.warning(f'GraphQL response missing repository data for {repo_owner}/{repo_name}')
+        return _empty_file_content_pairs(batch_changes), True
+
     results: Dict[str, FileContentPair] = {}
 
     for i, fc in enumerate(batch_changes):
@@ -1219,7 +1228,7 @@ def _fetch_file_contents_with_base_batch(
 
         results[fc.filename] = FileContentPair(old_content=old_content, new_content=new_content)
 
-    return results
+    return results, False
 
 
 def fetch_file_contents_with_base(
@@ -1229,7 +1238,7 @@ def fetch_file_contents_with_base(
     head_sha: str,
     file_changes: List['FileChangeType'],
     token: str,
-) -> Dict[str, FileContentPair]:
+) -> tuple[Dict[str, FileContentPair], bool]:
     """Fetch old and new versions of files in batches so large PRs don't hit complexity limits.
 
     Args:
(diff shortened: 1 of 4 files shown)


anderdc:
Closing along with #644 after thinking through the precedent this sets.

**Three layered reasons:**

**1. The failure mode is rare and self-healing.**

`get_pull_request_file_changes` already retries 3 times with exponential backoff, and `execute_graphql_query` has its own retry layer. A failure that survives both means GitHub is genuinely down or rate-limiting hard. When it does happen, impact is bounded: one PR scores 0 for one cycle, and next cycle (minutes later) it recovers. Cumulative lost value across a miner's lifetime is small.

**2. The fix's whole-miner fallback overcorrects.**

The PR's `score_miner_prs` wires one PR's content-fetch failure into `miner_eval.github_pr_fetch_failed = True` + `return` — aborting the entire round. Cache fallback then restores the **previous** evaluation wholesale, which has two new problems:

- **Loses fresh successful work**: if a miner has 10 merged PRs this cycle and PR #7 fails content fetch, PRs #1-6 already scored correctly get thrown away
- **Misses new PRs**: if the miner merged PRs since the last successful eval, the cached eval doesn't include them, so the miner gets stuck at a stale baseline

A per-PR failure triggering whole-miner rollback is wildly disproportionate.

**3. The fix sets a precedent that invites scope creep.**

"Skip on fetch failure after retries" as a principle applies equally to:

- Issue fetch failure (→ skip issue_multiplier)
- Review fetch failure (→ skip review_quality_multiplier)
- Label fetch failure
- Credibility data fetch failure
- Timeline event fetch failure

Each is defensible by the same logic. Six months later the scoring pipeline has a dozen "skip on failure" branches, each with its own test file, each consuming cognitive load. And those "skips" aren't free — they're actively **non-deterministic**. Same miner evaluated at the same time gets different scores depending on which pieces happened to fetch successfully. That's a worse state than "score is deterministically slightly wrong during outages."

**The current deterministic behavior is the better default.**

Content fetch fails after retries → `FileContentPair(None, None)` → `base_score = 0` for that PR only. Other PRs in the same cycle score correctly. Miner's total includes the one zero, bounded to that PR. DB records the fresh (slightly-low) eval. Next cycle recovers.

Properties of this behavior:

- **Deterministic**: same input produces same output
- **Simple**: no branching logic on fetch state
- **Bounded**: one PR's failure affects only that PR's contribution
- **Self-healing**: next cycle recovers

**The path forward if this becomes real:**

If validator logs show this failure mode happening with meaningful frequency in production, address it holistically: a single fetch-failure-handling policy applied consistently across all fetch points, not piece-by-case additions. Until we have that data, the current default is the right one.

Appreciate the diagnosis and the thorough testing — the bug is real, just the fix-shape invites more problems than it solves.

MkDev11:
@anderdc I addressed your comments and pushed the update

anderdc:
Appreciate the revision stripping the whole-miner abort and cache fallback, those were the right changes. But with those removed, what's left doesn't functionally differ from current behavior in any meaningful way:

**Functional impact of the revised PR:**

- **Skip vs `base_score=0`**: both contribute **0** to the miner's score sum. No emission/weights change.
- **The one real delta**: on skip, `miner_eval.unique_repos_contributed_to.add(repo)` is not called. This is actually **stricter** on the miner, not looser, which is the opposite of what you'd want when covering for a transient fetch failure.
- **Log message wording**: the only genuine improvement, and it's a small one.

**The cost for that small improvement:**

- 3 function signatures change to return `(data, fetch_failed)` tuples
- Every caller unpacks the tuple
- `get_pull_request_file_changes` return type changes from `List` to `Optional[List]`
- ~247 lines added, mostly tests scaffolding the new tuple contracts

**The proportionate fix for observability:**

If the goal is "operator can grep which specific PR hit a fetch failure," the minimum-viable change is to thread `pr_number` (or a context string) into the existing warning log inside `_fetch_file_contents_with_base_batch`:

```python
bt.logging.warning(f'Failed to fetch file contents for {repo_owner}/{repo_name} (PR #{pr_number})')
```

That's ~3-5 lines total. No tuple returns, no skip-vs-zero branching, no `None` vs `[]` distinction, no new test file. Keeps the current deterministic `FileContentPair(None, None)` fallback intact.

**Requesting changes:**

- Revert the tuple returns on `_fetch_file_contents_with_base_batch`, `fetch_file_contents_with_base`, and `fetch_file_contents_for_pr`
- Revert `get_pull_request_file_changes` back to returning `[]` on failure
- Revert the skip-early branches in `score_pull_request`
- Keep **only**: a PR-number (or context) argument threaded into the existing warning log
- Drop `tests/validator/test_scoring_fetch_failure.py`; update existing tests only if the log message format changes

Happy to merge the scoped-down version once it lands. The core diagnosis was real, the observability win is legitimate, just doesn't warrant the API surface change.

PR #651, opened 2026-04-21
Fix #650: classify miner-solved issues correctly in repo scan

### LINKED ISSUE #650: Issue discovery misclassifies miner-solved issues as non-miner solved
## Description

In `issue_discovery/repo_scan.py`, unmatched closed issues are classified as “solved by non-miner PR” whenever `solver_id is not None`.
This does not verify whether the solver is a miner. As a result, miner-solved issues can be incorrectly counted as non-miner solved, which can bias credibility scoring.

## Steps to Reproduce

1. Run issue discovery scan with at least one closed miner-authored issue not present in known_issues.
2. Ensure find_solver_from_cross_references(...) returns a valid solver_id that belongs to a miner.
3. Observe classification logic in _scan_repo: it sets issue.closed_at (Case 2 positive credibility) for any non-null solver, without checking miner membership.

## Expected Behavior

Issues should be marked as “solved by non-miner PR” only if solver_id is present and not in miner_github_ids.

## Actual Behavior

Any non-null solver_id is treated as non-miner solved:
- Miner solver IDs are not filtered out.
- Miner-solved issues can be misclassified as Case 2 (positive credibility signal).

## Environment

- OS: Linux 6.8.0-100-generic
- Python version: 3.12 

## Additional Context

Relevant code path in `gittensor/validator/issue_discovery/repo_scan.py`:
- Current logic:
      - `if solver_id is not None: ... # Case 2: solved by non-miner PR`
- Suggested fix:
      - Use miner membership check, e.g. classify Case 2 only when `solver_id` is not `None` and `str(solver_id) not in miner_github_ids`.

bohdansolovie:
**Summary**

- Fix issue discovery classification so Case 2 applies only when the solver is a non-miner.
- Prevent miner-solved unmatched issues from being incorrectly treated as non-miner solved.
- Add regression tests for both solver paths:

     - non-miner solver -> Case 2 (closed_at set)
     - miner solver -> Case 3 (closed_at remains None)

**Why**
_scan_repo previously treated any non-null solver_id as “solved by non-miner PR”, which could misclassify miner-solved issues and bias credibility outcomes.
**Testing**
`uv run pytest -q tests/validator/test_issue_discovery_repo_scan.py`
**Issues**
Closes #650

### DIFF
--- a/gittensor/validator/issue_discovery/repo_scan.py
+++ b/gittensor/validator/issue_discovery/repo_scan.py
@@ -210,7 +210,7 @@ async def _scan_repo(
             state_reason=(issue_raw.get('state_reason') or '').upper() or None,
         )
 
-        if solver_id is not None:
+        if solver_id is not None and str(solver_id) not in miner_github_ids:
             # Case 2: solved by non-miner PR → positive credibility
             issue.closed_at = _parse_iso(issue_raw.get('closed_at'))
         else:
--- a/tests/validator/test_issue_discovery_repo_scan.py
+++ b/tests/validator/test_issue_discovery_repo_scan.py
@@ -83,3 +83,61 @@ def test_scan_repo_sets_state_reason_none_when_missing(monkeypatch):
     assert '1001' in result
     assert result['1001'][0].state_reason is None
     assert result['1001'][0].is_transferred is False
+
+
+def test_scan_repo_treats_non_miner_solver_as_case_2(monkeypatch):
+    monkeypatch.setattr(
+        repo_scan,
+        '_fetch_closed_issues',
+        lambda repo_name, since, token: [_make_raw('completed', 46)],
+    )
+    monkeypatch.setattr(
+        repo_scan,
+        'find_solver_from_cross_references',
+        lambda repo, issue_number, token: (2002, 99),
+    )
+
+    result: dict = {}
+    asyncio.get_event_loop().run_until_complete(
+        repo_scan._scan_repo(
+            repo_name='test/repo',
+            lookback_date='2026-01-01T00:00:00Z',
+            validator_pat='x',
+            miner_github_ids={'1001'},
+            known_issues=set(),
+            result=result,
+            lookup_cap=10,
+        )
+    )
+
+    assert result['1001'][0].closed_at is not None
+    assert result['1001'][0].pr_number == 99
+
+
+def test_scan_repo_treats_miner_solver_as_case_3(monkeypatch):
+    monkeypatch.setattr(
+        repo_scan,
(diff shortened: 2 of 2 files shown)


anderdc:
Closing along with #650. Appreciate the close reading of the classification logic, but the proposed fix has a correctness issue that makes it worse than the current behavior, and the underlying bug is a narrow edge case that rarely triggers.

**The fix produces wrong classification when it matters.**

Current classification:

```python
if solver_id is not None:
    # Case 2: solved by non-miner PR → positive credibility
```

Proposed classification:

```python
if solver_id is not None and str(solver_id) not in miner_github_ids:
    # Case 2: solved by non-miner PR → positive credibility
else:
    # Case 3: closed without PR → negative credibility
```

The `else` branch now captures three distinct cases:

- `solver_id is None` → correct (no solver, negative credibility)
- `solver_id` non-miner → correct (handled by `if`)
- **`solver_id` is a miner → INCORRECT** — the issue WAS solved (by a miner), but it's now classified as "closed without PR" and the discoverer gets penalized with `closed_count += 1`

Semantically, a miner-solved issue is still a solved issue. Classifying it as Case 3 penalizes the discoverer for having their issue fixed by someone who happens to be a miner. That's worse than the current "slight over-credit via Case 2" outcome.

**The underlying edge case is rare.**

Miner-solved issues normally don't reach the scan path at all. They're added to `known_issues` via the miner's PR `closingIssuesReferences` and get filtered out as "matched" before `_scan_repo` sees them. For a miner-solved issue to leak into the scan:

- Partial fetch failure on the miner's PR
- GraphQL didn't populate `closingIssuesReferences` for that PR
- PR was merged outside the lookback window
- Some other pruning edge case

All rare. The current Case 2 over-credit in this rare path is bounded; the fix's Case 3 under-credit is a reliably wrong penalty.

**If a fix is warranted, the shape should be different.**

Miner-solved issues should be **skipped entirely** at the scan path, not routed to Case 3 — because the PR path is the canonical source for miner-solved classification:

```python
if solver_id is not None and str(solver_id) in miner_github_ids:
    continue  # miner solver — PR path handles classification
if solver_id is not None:
    # Case 2: solved by non-miner PR
    issue.closed_at = _parse_iso(issue_raw.get('closed_at'))
else:
    # Case 3: closed without PR
    issue.closed_at = None
```

But given how rarely this edge case triggers in practice, even that correctly-shaped fix probably isn't worth the complexity. Happy to reopen if validator logs show this failure mode occurring meaningfully in production.

Thanks for the diligence on the classification logic — the observation is real, just the proposed fix's semantics land in the wrong place.

PR #668, opened 2026-04-21
fix(storage): rollback partial evaluation writes

jamesrayammons:
## Summary

- stop repository helpers from committing or rolling back inside an outer transaction
- fail `store_evaluation()` when any non-empty sub-write returns a falsey result
- add regression coverage for both transaction ownership and partial-write rollback

## Related Issues

- None

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -37,6 +37,10 @@ class BaseRepository:
         self.db = db_connection
         self.logger = logging.getLogger(self.__class__.__name__)
 
+    def _manages_own_transaction(self) -> bool:
+        """Return True when repository methods should commit/rollback directly."""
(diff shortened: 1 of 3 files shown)


anderdc:
Closing — the atomicity issue is real but no linked issue or evidence of partial-write corruption in production scoring data. The PR also bundles a separate behavior change on top of the fix: hard-failing evaluations when `stored_count != expected_count`, which is a policy decision rather than an atomicity fix. Open an issue first if this is causing observable problems.

PR #669, opened 2026-04-21
fix(validator): prevent active github identity collisions

jamesrayammons:
## Summary

- reject PAT broadcasts that reuse a GitHub identity already held by another active miner
- reclaim stale GitHub identity entries when the old hotkey is no longer active on the subnet
- add regression coverage for both active collisions and stale-entry recovery

## Related Issues

- None

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -61,6 +61,22 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
                 'GitHub identity is locked for this hotkey. Deregister and re-register to change GitHub accounts.'
             )
 
+    # 3b. Enforce one active miner per GitHub account across the subnet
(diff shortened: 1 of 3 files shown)


anderdc:
Closing — the stale-reclaim branch contradicts GitHub identity pinning. Re-registering on a new hotkey should require a new GitHub account, which is what the existing error at `pat_handler.py:60-62` already implies ("Deregister and re-register to change GitHub accounts"). Silently rebinding the same github_id to a new hotkey defeats that.

The active-collision case is already handled by the duplicate-account penalty in `oss_contributions/inspections.py:16-46`, which zeroes scores of all colliding UIDs.

PR #672, opened 2026-04-21
Fix/inline test per block scoring

### LINKED ISSUE #631: [BUG] has_inline_tests decides test-file classification on post-state only — asymmetric
## Summary

In `calculate_token_score_from_file_changes`, the inline-test detection branch at L343-346 calls `has_inline_tests(new_content, ext)` to decide whether to apply the 20× `TEST_FILE_CONTRIBUTION_WEIGHT = 0.05` factor to the file's breakdown. Because `old_content` is not inspected, the classification depends only on the file's post-PR state. Diffs that add inline-test markers downweight the entire file; diffs that remove them do not.

## Affected code

`gittensor/validator/utils/tree_sitter_scoring.py:336-350`:

```python
file_breakdown = score_tree_diff(old_content, new_content, ext, weights)

lang_config = programming_languages.get(ext)
lang_weight = lang_config.weight if lang_config else 1.0

# For non-test files in inline-test languages, check if the current
# file contains inline tests and downweight the entire file if so
if not is_test_file and ext in INLINE_TEST_EXTENSIONS:
    if has_inline_tests(new_content, ext):
        is_test_file = True
        file_weight = TEST_FILE_CONTRIBUTION_WEIGHT

combined_weight = lang_weight * file_weight
file_breakdown = file_breakdown.with_weight(combined_weight)
```

`TEST_FILE_CONTRIBUTION_WEIGHT = 0.05` (`constants.py:35`). `INLINE_TEST_EXTENSIONS = frozenset({'rs', 'zig', 'd'})` (`constants.py:130`).

## Asymmetry table

| Scenario | old_content | new_content | has_inline_tests(new) | file_weight | Result |
|---|---|---|---|---|---|
| A — remove inline tests | tests present | tests removed | False | 1.0 | deleted tokens (including removed test code) scored at full language weight |
| B — add inline tests | no tests | tests present | True | 0.05 | all additions (production + tests in the same file) downweighted 20× |
| C — edit production in test-containing file | tests present | tests still present | True | 0.05 | production-only diff downweighted 20× because the file currently has tests |

Scenarios A and B touch the same test-code boundary in opposite directions; only B triggers the downweight.

## Reproduction

```python
from gittensor.validator.utils.tree_sitter_scoring import has_inline_tests

rust_with_tests = '''
pub fn foo() -> i32 { 42 }

#[cfg(test)]
mod tests {
    #[test]
    fn test_foo() { assert_eq!(foo(), 42); }
}
'''
rust_no_tests = 'pub fn foo() -> i32 { 42 }\n'

# new_content is what the call site feeds to has_inline_tests.

# Scenario A — remove inline tests: new_content has none
assert has_inline_tests(rust_no_tests, 'rs') is False

# Scenario B — add inline tests: new_content has them
assert has_inline_tests(rust_with_tests, 'rs') is True

# old_content is never inspected by the call site, so whether the
# diff added or removed the test markers does not factor into the
# file_weight selection.
```

## Context

PR #545 recently reduced the weight of deleted code at the `ScoreBreakdown` aggregation layer. This issue is a narrower case at a different layer: the binary test-file classification (which gates a 20× multiplier on the file's entire breakdown) is decided on post-PR state only.

## Scope note

Filing as a behavior note. Whether the resolution is a code change or a documentation change is for maintainers to decide.

RenzoMXD:
## Summary

Replace the file-level `has_inline_tests(new_content)` demotion in `tree_sitter_scoring.py` with per-block AST classification. Production subtrees score at the language's full weight (Rust 2.0); only nodes inside `#[cfg(test)] mod`, `#[test]` / `#[tokio::test]` functions, or `#![cfg(test)]`-gated files get `TEST_FILE_CONTRIBUTION_WEIGHT` (0.05). Inline-test-containing files emit two `FileScoreResult`s — one SOURCE, one TEST — so `source_density` is computed from production lines only.

## Problem

`calculate_token_score_from_file_changes` tags the **entire** file as TEST when `has_inline_tests(new_content, ext)` finds any `#[test]`-like marker in the post-PR content. The file is reclassified into `ScoringCategory.TEST`, emptying SOURCE. `source_token_score = 0 < MIN_TOKEN_SCORE_FOR_BASE_SCORE` sets `initial_base_score = 0`, leaving only the small `contribution_bonus`. Miners editing production code in files that happen to contain an unrelated `#[cfg(test)]` block are silently penalized 20×.

### Real example — [denoland/deno#32242](https://gittensor.io/miners/pr?repo=denoland%2Fdeno&number=32242)

"fix: deno install --force --compile" by statxc — when `deno install --force --compile` overwrites an existing installation, remove the old file before the compile step so its safety check doesn't abort.

- 22 insertions / 4 deletions in `cli/tools/installer/mod.rs`; the rest of the diff is under `tests/specs/install/…` (correctly classified by path).
- `cli/tools/installer/mod.rs` already contains an unrelated `#[test]` block — untouched by this PR.
- Under the bug: whole file demoted, SOURCE empty, `baseScore = 0.01`, earned ≈ 0.
- Under the fix: 26 production lines stay in SOURCE at Rust weight 2.0, `baseScore ≈ 6.13`.

## Approach

- `_inline_test_line_ranges(content, ext)` — single linear pass using `INLINE_TEST_PATTERNS` plus brace-balanced delimiter tracking to resolve `#[cfg(test)] mod X { ... }` and `#[test] fn Y() { ... }` ranges. `#![cfg(test)]` at crate root returns the whole file.
- `collect_node_signatures` — optional `test_ranges` parameter tags each AST signature with its inline-test membership via `node.start_point`.
- `_score_tree_diff_split` — partitions the symmetric-diff result into `(prod_bd, test_bd)` by signature tag.
- `calculate_token_score_from_file_changes` — for files in `INLINE_TEST_EXTENSIONS` with inline tests in either old or new content, emit two `FileScoreResult`s with proper categories and weights; `file.changes` is apportioned between the two buckets by node count.

Non-inline-test extensions and non-Rust/Zig/D files use the existing single-breakdown path unchanged. `score_tree_diff`'s external signature is preserved.

## Related Issues

Closes #631

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- 9 new tests in `test_inline_test_detection.py` covering `_inline_test_line_ranges` (non-inline-test ext, production-only, `#[cfg(test)] mod`, standalone `#[test] fn`, `#![cfg(test)]` crate gate, `#[tokio::test]`) and `_score_tree_diff_split` (production + test edits, no-inline-tests, test-only changes).
- Full validator suite: **242 passed**.

## Checklist

- [x] Self-reviewed
- [x] Tests added for new functionality
- [x] All tests passing

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -304,12 +304,16 @@ class PullRequest:
 
         current = {(n.get('name') or '').lower() for n in (pr_data.get('labels') or {}).get('nodes') or [] if n}
         label: Optional[str] = None
-        if not current.isdisjoint(LABEL_MULTIPLIERS):
+        scoring_labels = current & LABEL_MULTIPLIERS.keys()
+        if scoring_labels:
             for event in reversed((pr_data.get('timelineItems') or {}).get('nodes') or []):
                 name = ((event or {}).get('label') or {}).get('name', '').lower()
-                if name in current and name in LABEL_MULTIPLIERS:
+                if name in scoring_labels:
                     label = name
                     break
+            if label is None:
+                # Timeline truncated — fall back to highest-multiplier currently-applied label
+                label = max(scoring_labels, key=lambda n: (LABEL_MULTIPLIERS[n], n))
 
         return cls(
             number=pr_data['number'],
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -10,14 +10,13 @@ import os
 import re
 import struct
 import sys
-import urllib.error
-import urllib.request
 from contextlib import nullcontext
 from decimal import Decimal, InvalidOperation
 from pathlib import Path
 from typing import Any, Callable, ContextManager, Dict, List, Optional, Tuple, TypeVar
 
 import click
+import requests
 from rich.console import Console
 
 from gittensor.cli.issue_commands.tables import build_pr_table
@@ -366,19 +365,22 @@ def validate_repository(repo: str, verify_exists: bool = True) -> Tuple[str, str
     owner, repo_name = repo.split('/', 1)
 
     if verify_exists:
-        url = f'https://api.github.com/repos/{owner}/{repo_name}'
         try:
-            req = urllib.request.Request(url, headers={'User-Agent': 'gittensor-cli'})
-            urllib.request.urlopen(req, timeout=GITHUB_API_TIMEOUT)
-        except urllib.error.HTTPError as e:
-            if e.code == 404:
+            resp = requests.get(
+                f'https://api.github.com/repos/{owner}/{repo_name}',
+                headers={'User-Agent': 'gittensor-cli'},
+                timeout=GITHUB_API_TIMEOUT,
+            )
+            if resp.status_code == 404:
                 raise click.BadParameter(
                     f"Repository '{owner}/{repo_name}' not found on GitHub",
                     param_hint='--repo',
                 )
-            # Non-404 HTTP errors: warn but don't block
-            console.print(f'[yellow]Warning: GitHub API returned {e.code} — skipping existence check[/yellow]')
-        except (urllib.error.URLError, OSError):
+            if not resp.ok:
+                console.print(
+                    f'[yellow]Warning: GitHub API returned {resp.status_code} — skipping existence check[/yellow]'
+                )
+        except requests.RequestException:
             console.print('[yellow]Warning: Could not reach GitHub API — skipping existence check[/yellow]')
 
     return owner, repo_name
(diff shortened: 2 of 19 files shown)


anderdc:
Closing — lexical pattern matching (regex + brace-depth counting) for inline-test block detection defeats the point of AST-based scoring. `depth = lines[j].count('{') - lines[j].count('}')` breaks on strings, macros, and multi-line attributes, and we already have a tree-sitter parse to work from.

If this gets revisited, detection should walk the AST for `attribute_item` / `mod_item` nodes with `cfg(test)`, not character counts. Appreciate the effort — approach isn't what we want.

RenzoMXD:
@anderdc 
Thanks for your reply. 
But I think this feature is helpful to check test cases correctly.
And I have now several miner's PR examples which are from gittensor.io. 
For example, https://github.com/denoland/deno/pull/32242, https://github.com/denoland/deno/pull/32225, https://github.com/zed-industries/zed/pull/54307, https://github.com/zed-industries/zed/pull/54300 and so on.

I think we need to fix them. If you give me any advice, I'll update my code. Thanks.

PR #679, opened 2026-04-21
fix: keep miner post json success for zero accepts

reganbuilds:
## Summary
- keep `success: true` for completed `gitt miner post --json-output` runs even when zero validators accept the PAT

## Problem
The command currently uses `success` to mean "at least one validator accepted the PAT" instead of "the command completed successfully". That makes a completed zero-accept result look like a CLI failure to JSON consumers.

## Validation
- ran `python3 -m py_compile gittensor/cli/miner_commands/post.py`


anderdc:
These PRs are just causing churn

PR #680, opened 2026-04-21
fix: make is_valid_issue close-window directional

### LINKED ISSUE #676: [BUG] is_valid_issue close-window uses abs()
## Description

`is_valid_issue` at [scoring.py:491-497](gittensor/validator/oss_contributions/scoring.py#L491-L497) checks that the issue's `closed_at` and the PR's `merged_at` fall within `MAX_ISSUE_CLOSE_WINDOW_DAYS = 1`:

```python
if issue.closed_at and pr.merged_at:
    days_diff = abs((issue.closed_at - pr.merged_at).total_seconds()) / SECONDS_PER_DAY
    if days_diff > MAX_ISSUE_CLOSE_WINDOW_DAYS:
        return False
```

`abs()` makes the window symmetric. An issue closed up to 24 hours **before** the PR merges passes the gate — even though the PR could not have caused the close.

## Gaming vector

GitHub's `closingIssuesReferences` is populated by keyword parsing of the PR body (`Closes #N` / `Fixes #N` / `Resolves #N`). It does not validate that the PR actually fixes the target issue, and it still returns references when the target is already closed.

1. A legitimate PR by miner A (or a maintainer) closes issue #100. `issue.state=CLOSED, state_reason=COMPLETED, closed_at=t0`.
2. Miner B has an unrelated PR in review. Before merge, B adds `Closes #100` to the PR body.
3. Miner B's PR merges at `t0 + 23h`.
4. Every gate in `is_valid_issue` passes for (B's PR, issue #100): author present, different authors, `issue.created_at ≤ pr.created_at`, no post-merge edit, `state=CLOSED`, `state_reason=COMPLETED`, `abs(t0 − (t0+23h)) = 0.96d ≤ 1d`.
5. `calculate_issue_multiplier` returns `1.33×` or `1.66×` (maintainer-authored #100) on top of B's own `token_score`.

No downstream check verifies that the PR actually caused the issue close.

## Expected behavior

Require the issue to close **at or shortly after** the PR merges — i.e. the issue close must be plausibly a consequence of this PR's merge.

## Actual behavior

Any issue closed within 24 hours on either side of the PR merge passes, including issues closed by unrelated work before the PR exists.

## Scope

Multiplier inflation only (1.33×–1.66×). Does not affect `base_score` directly; does not apply when the PR's own `token_score < 5` since the PR wouldn't receive `base_score` in the first place. Issue discovery credibility path is unaffected (it doesn't use this check).

## Related

- #632 (merged) — added `state_reason != 'COMPLETED'` gate. Does not address the window direction.
- #601 (closed unmerged) — proposed `MIN_ISSUE_AGE_HOURS` on issue creation vs PR creation. Orthogonal to close vs merge.
- #462 (open) — self-created-and-immediately-solved issues. Different attack surface.
- #53 (merged) — blocks `Closes #N` inserted **after** merge via `last_edited_at > merged_at`. The attack here inserts the link **before** merge, so #53's check does not catch it.

This preserves the existing upper bound (issue must close within 1 day of merge) while rejecting the gaming case where `issue.closed_at` predates `pr.merged_at`.

plind-junior:
The previous abs() check accepted issues closed up to 24h BEFORE the PR merged. GitHub's closingIssuesReferences is keyword-parsed from the PR body and still returns references to already-closed issues, so an unrelated PR could add `Closes #N` pointing to an issue someone else just closed, pass every is_valid_issue gate, and inherit the 1.33x / 1.66x multiplier.

Require issue.closed_at >= pr.merged_at - 60s (clock-skew buffer) and <= pr.merged_at + 1d. The legitimate auto-close case (issue closes at or just after merge) still passes; the "already closed by unrelated work" case is rejected.

Fixes #676


anderdc:
Fix is correct but over-engineered. Should be a one-liner:

```python
if days_diff > MAX_ISSUE_CLOSE_WINDOW_DAYS or days_diff < 0:
```

Just drop `abs()` and add `< 0`. Skip the new `ISSUE_CLOSE_CLOCK_SKEW_SECONDS` constant — GitHub's merge auto-close timestamps are tightly coupled; the buffer solves a problem that doesn't exist.

Tests: 6 separate functions for one boolean check is too much. Either parametrize into a single test or edit the existing close-window assertion in `tests/validator/test_is_valid_issue.py`. Don't add a new file.

plind-junior:
Thanks for your review @anderdc 
Made an update by following your suggestion

anderdc (approved):

PR #687, opened 2026-04-22
perf: pool http connections per miner scoring iteration

seroperson:
Validator scoring makes many GitHub REST and GraphQL requests per miner and each of those previously opened a new TCP socket and negotiated a new TLS session. This PR introduces a `requests.Session`-based pool scoped to one miner's fetch iteration, so every http call within that iteration reuses the same kept-alive TCP connection.

Re-using existing connections makes things bit faster and overall it's a good practice when you connect to the same host again and again.

## Summary

 - New `session_scope()` context manager holds a module-level `_session_cache: Optional[Dict[str, requests.Session]]`, populated on enter and cleaned up (sessions `.close()`d, cache set back to `None`) on exit.
- New `get_session(token)` returns a per-PAT cached `Session` inside a scope and a fresh uncached session outside it. Tokenless calls route through the same helper via `get_session(token or '')`.
- New `make_anonymous_headers()` mirrors the existing `make_headers` / `make_graphql_headers` pattern for unauthenticated GitHub calls.
- All call sites now use `session = get_session(token)`
- `session_scope()` wraps `load_miners_prs` + `score_miner_prs` inside `evaluate_miners_pull_requests` so one cache covers an entire miner's load+score cycle; connections close before the function returns.

## Testing

No new tests are introduced. Added the `tests/conftest.py`, which installs an autouse `_ForwardingSession` shim that delegates `.get/.post` back to the module-level `requests.get/post`, so existing dozens of tests which use `@patch('requests.get'/'requests.post')` keep working unchanged.

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -175,7 +175,11 @@ def make_headers(token: str) -> Dict[str, str]:
 
 def make_graphql_headers(token: str) -> Dict[str, str]:
     """Build GitHub GraphQL headers for a PAT."""
-    return {'Authorization': f'Bearer {token}', 'Content-Type': 'application/json'}
+    return {
+        'Authorization': f'Bearer {token}',
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+    }
 
 
 def make_anonymous_headers() -> Dict[str, str]:
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -3,10 +3,11 @@ import base64
 import fnmatch
 import os
 import time
+from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from math import ceil
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
 
 from gittensor.utils.utils import parse_repo_name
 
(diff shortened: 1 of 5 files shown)


anderdc:
Can you prove the actual performance gain from this change? Otherwise it seems like changing something that already is proven to work

seroperson:
> Can you prove the actual performance gain from this change? Otherwise it seems like changing something that already is proven to work

Not a game-changing x10 of course, but still something. Considering how many requests it does during a round, it should result into noticeable gain.

```
$ > GITHUB_PAT=... python /tmp/bench_session_scope.py

N=15 sequential authenticated /user calls, 3 repetitions
==============================================================================
 run 1:
  fresh Session per call        6306.4 ms  (420.4 ms/call)  TCP conns: 15
  one shared Session            4796.4 ms  (319.8 ms/call)  TCP conns: 1
 run 2:
  fresh Session per call        6106.1 ms  (407.1 ms/call)  TCP conns: 15
  one shared Session            4481.7 ms  (298.8 ms/call)  TCP conns: 1
 run 3:
  fresh Session per call        7436.8 ms  (495.8 ms/call)  TCP conns: 15
  one shared Session            4672.7 ms  (311.5 ms/call)  TCP conns: 1
```

[bench_session_scope.py](https://github.com/user-attachments/files/26982724/bench_session_scope.py)

seroperson:
Updated to match current `test`. I see `MirrorClient` also uses sessions, so now it has bit more value to get in line =)

anderdc (changes requested):
Bench is convincing — happy to take this. Three asks before merge:

1. **Mirror parity.** `MirrorClient` already holds a `requests.Session` (`gittensor/utils/mirror/client.py:44`), but in `evaluate_miners_pull_requests` (`reward.py:75-83`) `load_mirror_miner_prs` and `score_mirror_miner_prs` each default `client = client or MirrorClient()`, so per miner there are two separate Session pools to `mirror.gittensor.io` and no connection reuse across the load→score boundary — the same waste this PR fixes for the PAT path. Instantiate one `MirrorClient()` in `evaluate_miners_pull_requests` and thread it through both calls via the existing `client=` kwarg.

2. **Accept-header bleed into GraphQL.** `_build_session` applies `make_headers(token)` to `session.headers`, which sets `Accept: application/vnd.github.v3+json`. In `execute_graphql_query` and `get_github_graphql_query` the per-request `headers=make_graphql_headers(token)` only overrides `Authorization`/`Content-Type`, so GraphQL POSTs now carry the v3 REST `Accept` they didn't before. GitHub tolerates it, but it's an unintended drift. Either drop `Accept` from `_build_session` (let call sites own it) or add `Accept: application/json` to `make_graphql_headers`.

3. **Cover \`session_scope\` directly.** The new \`tests/conftest.py\` shim swaps \`get_session\` autouse, so the suite doesn't exercise \`_session_cache\`, the re-entrance guard, or close-on-exit. Add a small test: same Session returned for same token within a scope, distinct per token, re-entrance raises, sessions \`.close()\`d on exit. While there: the shim calls \`make_headers(token)\` unconditionally, but prod uses \`make_anonymous_headers()\` for \`token=''\` — minor test/prod drift on the unauth REST search fallback.

anderdc (approved):

PR #688, opened 2026-04-22
fix: guard remaining null-author PR paths

reganbuilds:
## Summary
- guard the remaining direct `author["login"]` reads when GraphQL returns a null PR author
- default `PullRequest.from_graphql_response()` to `ghost` for missing authors instead of crashing
- add regressions for both the class parser and self-merge guard path

## Validation
- `uv run --python 3.12 --with pytest pytest tests/test_classes.py tests/utils/test_null_author_guards.py -q`
- `python3 -m py_compile gittensor/classes.py gittensor/utils/github_api_tools.py tests/test_classes.py tests/utils/test_null_author_guards.py`


anderdc:
This PR is just churn

PR #709, opened 2026-04-22
refactor: remove dead setup_events_logger helper and unused events.log handler

### LINKED ISSUE #708: `setup_events_logger` is dead template code - never used
## Summary

`setup_events_logger` in `gittensor/utils/logging.py` is inert Bittensor subnet-template code. It runs on every validator startup (via `BaseNeuron.__init__` → `check_config`) and does three things:

- Monkey-patches `logging.Logger.event` onto the stdlib `Logger` class.
- Opens a `RotatingFileHandler` on `events.log`.
- Registers the logger with bittensor as a primary logger.

Nothing in the repo ever calls `.event(...)`. A quick search confirms it:

```
$ rg -n '\.event\(' --type py
(no matches)
```

So the custom level, the monkey-patch, and the rotating file handler all run at startup for an event stream that no code ever emits.

## Motivation

The real cost is not disk space - the file is basically always empty. The cost is:

- **Stdlib side effect.** `logging.Logger.event = event` adds a method onto every logger in the process, just because this helper ran during `check_config`. That leaks well outside gittensor's own code.
- **Open file handle + rotating-handler lifecycle** for a file that nobody writes to, held for the full life of the validator process.
- **Runs on every validator boot** - not CLI-only.

Same shape as `is_cuda_available()` removed in #675: a Bittensor-template helper that does real work at startup for a feature gittensor does not use.

Crystora:
## Summary

Removes `setup_events_logger()` from `gittensor/utils/logging.py` and its single call site in `check_config()`. The helper ran on every validator startup, monkey-patching `logging.Logger.event` onto the stdlib `Logger` class and opening a `RotatingFileHandler` on `events.log`, but nothing in the repository ever calls `.event()`. 

Normal `bt.logging.*` calls and the `bittensor.log` rotating file are unaffected - those flow through bittensor's `LoggingMachine`, not through the removed helper.

## Related Issues

Closes #708 

## Type of Change

- [ ] Bug fix
- [x] Refactor
- [ ] New feature
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [ ] Tests added/updated
- [x] Manually tested

Full test suite passes (410 tests). `ruff check`, `ruff format`, and `pyright` all clean.

## Checklist

- [x] Self-review completed


Crystora:
@anderdc would you please review this PR?

anderdc (changes requested):
Cleanup is incomplete — the CLI surface for the removed feature is still registered. Please also drop the two orphan `add_argument` blocks in `gittensor/utils/config.py`:

- `--neuron.events_retention_size` (lines 80-85)
- `--neuron.dont_save_events` (lines 87-92)

Nothing consumes them after this PR, so they'd just show up in `--help` and silently accept values that do nothing.

Crystora:
@anderdc Updated!

anderdc (approved):

PR #42, opened 2025-12-01
Add comprehensive test suite for scoring module

NikolaPantel:
- Add 38 unit tests covering all scoring functions
- Test issue multiplier calculation with various scenarios
- Test issue validation logic (gaming prevention)
- Test time decay multiplier (sigmoid curve)
- Test PR spam penalty calculation
- Test repository contributor counting
- Include edge cases and error handling

Contribution by Gittensor, learn more at https://gittensor.io/

PR #713, opened 2026-04-22
fix(cli): fail closed on GitHub verification skip in issues register

### LINKED ISSUE #712: [Bug] `gitt issues register` submits the on-chain transaction even when GitHub verification was skipped
## Description

`gitt issues register` is the owner-only command that writes a new issue bounty. Before writing, it calls `validate_github_issue()` to check with GitHub that the target issue actually exists, is an issue (not a PR), and is reachable.

The helper has three possible outcomes:

- Issue confirmed OK → returns a dict with the issue data.
- Issue definitely invalid (404, or it turns out to be a PR) → raises and the CLI aborts.
- Verification could not be completed (network error, 403, 500, 502, 503, etc.) → prints a yellow warning like `Warning: GitHub API returned 503 — skipping issue check` and returns `None`.

The third case is the problem. In `gittensor/cli/issue_commands/mutations.py:139`, `issue_register()` calls the helper but does not use its return value:

```python
validate_github_issue(owner, repo_name, issue_number)
```

No `result = ...`, no `if result is None: abort`. The warning prints, execution continues, and at `mutations.py:210-220` the CLI submits `register_issue` on-chain. In interactive mode the user at least sees the yellow warning before a default-Yes confirm prompt, but with `-y` / `--yes` or in non-interactive environments (CI, scripts) the prompt is skipped and the write happens with nothing in between.

So any GitHub outage or 5xx / 403 response during a registration attempt is currently enough for the CLI to put a bounty on-chain for an issue that was never actually verified.

## Steps to Reproduce

1. Run `gitt issues register --repo <owner>/<repo> --issue <n> --bounty <amount> -y` as the contract owner.
2. Arrange for `validate_github_issue()` to hit the verification-skipped path. In practice this means GitHub temporarily returns 5xx / 403, or the HTTP call raises `requests.RequestException`. Reproduced here by mocking `helpers.requests.get` to return a 503 response and by forcing a `RequestException`.
3. Observe: the CLI prints the yellow "skipping issue check" warning, and then still calls `contract_instance.exec('register_issue', ...)` at `mutations.py:210`. Exit code is 0.

## Expected Behavior

The owner-only mutation path should fail closed when GitHub verification cannot be completed. If the CLI cannot confirm that:

- the repository exists,
- the target is an issue and not a PR,
- the issue state was actually fetched,

then it should not submit the on-chain transaction. Something like a plain message along the lines of *"Could not verify issue #N in owner/repo on GitHub (status 503). Try again when GitHub is reachable."* and a non-zero exit would be the expected behavior for a command that is about to spend real ALPHA.

Read-only commands that also use these helpers can keep the current warn-and-skip behavior; only the mutation path needs to be strict.

## Actual Behavior

The CLI prints the warning, returns from the helper, proceeds through the confirmation prompt (or skips it with `-y`), and submits the `register_issue` transaction. An on-chain bounty entry is created whose GitHub target was never verified.

The validator side does not auto-clean entries like this:

- `gittensor/utils/github_api_tools.py:1099-1101` — `check_github_issue_closed()` returns `None` on any non-200 response, without distinguishing "issue does not exist" from "GitHub is flaky".
- `gittensor/validator/issue_competitions/forward.py:92-98` — when `check_github_issue_closed()` returns `None`, the validator just logs a warning and `continue`s.

So an entry registered during a verification skip can sit on-chain indefinitely, logged each cycle but never resolved or cancelled, until the owner manually runs the `cancel_issue` admin command.

## Environment

- OS: Linux 6.8.0-107-generic
- Python version: 3.12 (project requires `>=3.12`)

carlos4s:
## Summary

`gitt issues register` is owner-only and spends real ALPHA. Before writing to the contract, it calls `validate_repository()` and `validate_github_issue()` to check the target on GitHub — but both helpers silently warn-and-continue on network errors, 5xx, or 403 rate-limits. `issue_register` drops the return value, so a GitHub blip is enough for the CLI to fall through to `contract_instance.exec('register_issue', ...)` with nothing verified. With `-y` there isn't even a confirm prompt in between.

The validator side doesn't clean these up either — unresolvable entries just get logged each cycle until the owner cancels them manually.

Fix is narrow and opt-in: add a `require_verified_exists=False` flag to both helpers. Default keeps every existing caller (view, list, vote, admin, harvest) exactly as it was. Only `issue_register` flips it to `True`, and the register flow now aborts with a clear error before the on-chain call if GitHub verification couldn't complete. A 404 still raises its existing "not found" message; the flag only tightens the skip branches.

## Changes

- **`gittensor/cli/issue_commands/helpers.py`** — new private `_raise_github_verification_required()` helper, and a `require_verified_exists` kwarg on `validate_repository` / `validate_github_issue` that turns the warn-and-skip branches into `click.BadParameter`.
- **`gittensor/cli/issue_commands/mutations.py`** — `issue_register` passes `require_verified_exists=True` to both validators. Nothing else changes; the existing `except click.BadParameter` already surfaces it as a non-zero `ClickException`.
- **`tests/cli/test_cli_helpers.py`** — unit tests for 503 / 403 / `ConnectionError` / `Timeout`, default-off back-compat, 404 still winning. Plus two end-to-end tests that run `gitt issues register ... -y`, mock GitHub into a 503 / `ConnectionError`, and use a sentinel to assert `contract_instance.exec` is never reached.

## Test plan

- [x] `uv run --with pytest pytest tests/cli/test_cli_helpers.py -q` — 84 passed
- [x] `uv run --with ruff ruff check gittensor/cli/issue_commands/ tests/cli/`
- [x] `uv run --with ruff ruff format --check gittensor/cli/issue_commands/ tests/cli/`
- [x] Read-only paths (`gitt issues list`, `gitt vote ...`, `gitt admin ...`, `gitt harvest`) still warn-and-skip on network errors — they don't pass the flag.

Closes #712

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -413,7 +413,7 @@ def validate_repository(
                 )
         except requests.RequestException as exc:
             if require_verified_exists:
-                detail = 'timeout' if isinstance(exc, requests.Timeout) else 'network error'
+                detail = type(exc).__name__
                 _raise_github_verification_required(
                     f"repository '{owner}/{repo_name}'",
                     detail,
@@ -446,7 +446,7 @@ def validate_github_issue(
         )
     except requests.RequestException as exc:
         if require_verified_exists:
-            detail = 'timeout' if isinstance(exc, requests.Timeout) else 'network error'
+            detail = type(exc).__name__
             _raise_github_verification_required(
                 f'issue #{issue_number} in {owner}/{repo}',
                 detail,
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -347,12 +347,37 @@ def validate_bounty_amount(bounty: str) -> int:
     return raw
 
 
-def validate_repository(repo: str, verify_exists: bool = True) -> Tuple[str, str]:
+def _raise_github_verification_required(
+    check: str,
+    detail: str,
+    *,
+    param_hint: str,
+) -> None:
+    """Raise BadParameter when a GitHub existence probe could not complete.
+
+    Used by the strict variants of validate_repository and validate_github_issue
+    so mutation commands never fall through to on-chain writes on transient failures.
+    """
+    raise click.BadParameter(
+        f'Could not verify {check} on GitHub ({detail}). Try again when GitHub is reachable.',
+        param_hint=param_hint,
+    )
+
+
+def validate_repository(
+    repo: str,
+    verify_exists: bool = True,
+    *,
+    require_verified_exists: bool = False,
+) -> Tuple[str, str]:
     """Validate owner/repo format and optionally verify it exists on GitHub.
 
-    Returns (owner, repo_name) on success.
-    Raises click.BadParameter on failure.
+    Returns (owner, repo_name) on success. Raises click.BadParameter on failure.
+    Pass require_verified_exists=True to abort on transient GitHub errors instead
+    of warning and continuing; requires verify_exists=True.
     """
+    if require_verified_exists and not verify_exists:
+        raise ValueError('require_verified_exists requires verify_exists=True')
     repo = repo.strip()
 
     if not REPO_PATTERN.match(repo):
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
Functionality looks correct and the fix is proportionate. A few things to clean up before merge.

**1. Docstring style doesn't match the repo convention**

Existing docstrings in `helpers.py` (and the rest of the CLI modules) are terse, plain-text, no backticks, typically 1-2 sentences. Examples from the same file:

- `"""Apply Click decorators in the declared display order."""`
- `"""Format raw token amount (9-decimal) as human-readable ALPHA string."""`
- `"""Wrap status text with the appropriate Rich color tag."""`

The new docstrings in this PR use RST-style double-backticks (``` ``validate_repository`` ```, ``` ``click.BadParameter`` ```) and multi-paragraph design-justification prose. The CLI modules have **zero** prior uses of double-backticks, so this PR would be introducing a new style into files that have never had it.

Please rewrite the new docstrings to match. For `_raise_github_verification_required`, something like:

```python
def _raise_github_verification_required(...) -> None:
    \"\"\"Raise BadParameter when a GitHub existence probe could not complete.

    Used by the strict variants of validate_repository and validate_github_issue
    so mutation commands never fall through to on-chain writes on transient failures.
    \"\"\"
```

Same for the updated docstrings on `validate_repository` and `validate_github_issue` — keep them terse, drop the backticks, drop the multi-paragraph rationale.

**2. Error messages leak Python exception class names to end users**

The current format produces:

- `Could not verify issue #42 in owner/repo on GitHub (status 503). ...` (readable)
- `Could not verify issue #42 in owner/repo on GitHub (ConnectionError). ...` (leaks Python class name)
- `Could not verify issue #42 in owner/repo on GitHub (Timeout). ...` (leaks Python class name)

`ConnectionError` / `Timeout` are Python class names, not meaningful to CLI users. Map exception paths to human-readable terms:

```python
# Currently:
_raise_github_verification_required(..., type(exc).__name__, param_hint='--issue')

# Better:
_raise_github_verification_required(..., 'network error', param_hint='--issue')
```

If you want to distinguish timeout from connection error, map them explicitly:

```python
detail = 'timeout' if isinstance(exc, requests.Timeout) else 'network error'
```

**3. Optional — latent parameter combo**

`validate_repository` accepts both `verify_exists=True` (default) and `require_verified_exists=False` (default). If a future caller passes `verify_exists=False, require_verified_exists=True`, the verification block is skipped entirely and `require_verified_exists` silently has no effect. Not an issue today since `issue_register` uses the right combination, but worth a guard:

```python
if require_verified_exists and not verify_exists:
    raise ValueError('require_verified_exists requires verify_exists=True')
```

Optional, take it or leave it depending on your defensive-coding taste.

**Otherwise:**

The bug is real and well-documented (bittoby's issue #712 was excellent), the fix is narrow (opt-in flag, default preserves all existing callers), the tests are unusually good (back-compat lock-in, 404 still wins, sentinel tests that `contract_instance.exec` is never reached). Once the docstring style and the exception-name leak are fixed, this is ready to merge.

carlos4s:
@anderdc Thanks for your feedback. I updated all based on your feedback. Please review again.

anderdc:
can you provide screenshots of the commands working correctly/failing properly

anderdc (approved):

Issue #714, opened 2026-04-22
feat(cli): confirm destructive admin and vote commands


wmagev:
## Summary
- Adds a y/n confirmation prompt before executing destructive owner/validator commands: `admin cancel`, `admin payout`, `admin set-owner`, `admin set-treasury`, `admin add-validator`, `admin remove-validator`, `vote solution`, `vote cancel-issue`.
- Bypassable with `--yes` / `-y` and auto-skipped when stdin is not a TTY, so scripts and CI pipelines are unaffected.
- `admin set-owner` panel now calls out that the transfer is irreversible, since a mistyped address permanently bricks the contract.
- Confirmation lives in a shared `confirm_or_abort` helper in `issue_commands/helpers.py` reusing the existing `with_cli_behavior_options(include_yes=True)` decorator.

## Test plan
- [ ] `gitt admin cancel <id>` in a TTY prompts and aborts cleanly on `n`
- [ ] `gitt admin cancel <id> --yes` skips the prompt
- [ ] `echo | gitt admin cancel <id>` (non-TTY) does not hang on the prompt
- [ ] `gitt admin set-owner <addr>` shows the red irreversibility warning before prompting
- [ ] `gitt vote solution …` / `gitt vote cancel-issue …` prompt and honor `--yes`

PR #715, opened 2026-04-22
feat(cli): confirm destructive admin and vote commands

plind-junior:
## Summary
- Adds a y/n confirmation prompt before executing destructive owner/validator commands: `admin cancel`, `admin payout`, `admin set-owner`, `admin set-treasury`, `admin add-validator`, `admin remove-validator`, `vote solution`, `vote cancel-issue`.
- Flag mirrors btcli: `--no-prompt` skips the gate (with `--yes` and `-y` kept as aliases so existing gittensor scripts still work). Auto-skipped when stdin is not a TTY, so scripts and CI pipelines are unaffected.
- `admin set-owner` panel now calls out that the transfer is irreversible, since a mistyped address permanently bricks the contract.
- Confirmation lives in a shared `confirm_or_abort` helper in `issue_commands/helpers.py` reusing the existing `with_cli_behavior_options(include_yes=True)` decorator.
- Existing `issues register` already has this gate (introduced in #211); this PR extends the same pattern to the remaining on-chain writes. `register`'s inline check could later be migrated onto `confirm_or_abort` — left as a follow-up to keep this PR tight.

## Why — matches the `btcli` precedent
`btcli`, the reference Bittensor CLI, prompts for y/n confirmation before every state-changing / on-chain action and exposes `--no-prompt` / `-y` to skip it for scripting. Operators already expect this flag across the Bittensor ecosystem. Examples of what btcli gates:

- `btcli wallet transfer` — confirms recipient + amount before signing
- `btcli stake add` / `btcli stake remove` — confirms amount and hotkey
- `btcli subnet register` / `btcli root register` — confirms the burn cost
- `btcli root set-weights`, `btcli root boost`, `btcli root slash` — confirms weight changes
- `btcli subnet create` — confirms the lock cost
- `btcli wallet regen-*` / wallet deletion — confirms overwrite of existing keys

The gittensor admin and validator-vote commands are the same class of action — owner-only contract writes and validator votes that trigger payouts — so applying the same gate (print the details in a Panel, then `click.confirm`, with `--no-prompt` / `--yes` / `-y` to bypass and auto-skip on non-TTY) keeps UX consistent with what Bittensor operators already expect.

## Test plan
- [ ] `gitt admin cancel <id>` in a TTY prompts and aborts cleanly on `n`
- [ ] `gitt admin cancel <id> --no-prompt` skips the prompt (btcli-style)
- [ ] `gitt admin cancel <id> --yes` / `-y` also skip the prompt (back-compat)
- [ ] `echo | gitt admin cancel <id>` (non-TTY) does not hang on the prompt
- [ ] `gitt admin set-owner <addr>` shows the red irreversibility warning before prompting
- [ ] `gitt vote solution …` / `gitt vote cancel-issue …` prompt and honor `--no-prompt`
- [ ] `gitt admin cancel-issue --help` lists `--yes, --no-prompt / -y`

#714

### DIFF
--- a/gittensor/cli/issue_commands/admin.py
+++ b/gittensor/cli/issue_commands/admin.py
@@ -21,6 +21,7 @@ from .helpers import (
     _handle_command_error,
     _make_contract_client,
     _resolve_contract_and_network,
+    confirm_or_abort,
     console,
     format_alpha,
     print_error,
@@ -28,6 +29,7 @@ from .helpers import (
     print_success,
     require_valid_issue_id,
     require_valid_ss58,
+    with_cli_behavior_options,
     with_network_contract_options,
     with_wallet_options,
 )
@@ -46,7 +48,10 @@ def admin():
 @click.argument('issue_id', type=int)
 @with_wallet_options()
 @with_network_contract_options('Contract address (uses config if empty)')
-def admin_cancel(issue_id: int, network: str, rpc_url: str, contract: str, wallet_name: str, wallet_hotkey: str):
+@with_cli_behavior_options(include_yes=True)
+def admin_cancel(
+    issue_id: int, network: str, rpc_url: str, contract: str, wallet_name: str, wallet_hotkey: str, yes: bool
+):
     """Cancel an issue (owner only).
 
     [dim]Immediately cancels an issue without validator consensus. Bounty funds are returned to the alpha pool.[/dim]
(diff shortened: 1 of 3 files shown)


plind-junior:
@anderdc Ready for reviwe

anderdc:
please provide screenshots of the CLI and the new outputs

plind-junior:
image
Check for y/n prompt

plind-junior:
image

plind-junior:
@anderdc Uploaded the screenshots for the update

plind-junior:
image

anderdc (approved):

PR #717, opened 2026-04-22
fix(cli): handle partial miner responses and require explicit register success

andriypolanski:


andriypolanski:
Hi, @anderdc 

Could you please reopen my pr?

PR #434, opened 2026-04-15
feat(issue-discovery): precise post-merge edit detection upgrade

corevibe555:
Closes https://github.com/entrius/gittensor/issues/402

## Summary

Replaces the `issue.updated_at` anti-gaming logic in issue discovery scoring with precise timeline-based detection for body and title edits only. The old check false-positived on any activity (bot comments, label changes, reactions, stale-bots), silently demoting legitimate solved issues to `closed_count` and unfairly cutting miner credibility in active repos.

Body edit and title edit both should be checked.
This is because title and body are connected.
If title change allowed, it's possible to confuse the reviewer by changing title.
From the purpose of this feature from doc: https://docs.gittensor.io/issue-discovery.html, title is important to define scope, it's possible to inflate by just changing title.

## Related Issues

Closes #402

## Type of Change

- [x] New feature / enhancement (anti-gaming precision upgrade)
- [ ] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

**Added:** `tests/validator/test_issue_discovery_post_merge_edit.py` — 3 cases:

1. Benign `updated_at` bump after merge (bot activity) → issue stays `solved`.
2. Real body edit after merge → issue demoted to `closed`, no score.
3. Body edit before merge → ignored, stays `solved`.

**Results:** 3/3 new tests pass. Broader existing suite (66 validator tests) still green. Pre-existing `tree_sitter` import errors in unrelated test modules are not affected by this change.

### Manual testing
1. Create an issue on open source repo. in my case, created one issue here: https://github.com/corevibe555/corevibe555/issues
2. Create a new branch and create PR, mention "Closes #3" in PR description, merge PR
3. If you go to issues list, the issue is closed.
4. The best way to see updated_at is changed is adding a comment, so please add a new comment to that closed issue #3, see the screenshot.
5. Change title case, attached screenshot
6. Change body case, attached screenshot

**Before adding comment**
graphql-before-add-comment-to-closed-issue

**After adding commment**
graphql-after-add-comment-to-closed-issue

**After changing title**
graphql-after-title-change

**After chaing body**
graphql-after-body-change

## Checklist

- [x] Code follows project style
- [x] Self-review completed (follow-up noted: `body_or_title_edited_at` is not persisted to DB — fine today since scoring re-fetches from GraphQL each run, but worth revisiting if any code path reconstructs `Issue` from storage)
- [x] Comments explain non-obvious *why*, not *what*
- [x] New tests added and passing
[...]

PR #732, opened 2026-04-23
fix: use closed_at as edit anchor to prevent auto-close false positives in post-merge edit detection

### LINKED ISSUE #443: Post-merge edit detection incorrectly penalises issues auto-closed by merged PRs
## Description

The anti-gaming post-merge edit detection in `_collect_issues_from_prs` (`gittensor/validator/issue_discovery/scoring.py`) incorrectly flips solved issues to closed when the issue was auto-closed by the merged PR.

When a PR merge auto-closes a linked issue, GitHub sets the issue's `updated_at` to the close timestamp, which is always slightly after `merged_at`. The current check `issue.updated_at > pr.merged_at` treats this normal auto-close as a post-merge edit, causing:

- `solved_count` to be decremented and `closed_count` to be incremented
- The issue to be classified as "closed" instead of "solved"
- Credibility to drop (e.g. from 100% to 0.0%)
- The `valid_solved_count` to not be incremented (bypassed by the early `continue`)

## Steps to Reproduce

1. A miner opens an issue on a tracked repository
2. A PR referencing the issue (e.g. `Fixes #N`) is merged
3. GitHub auto-closes the issue, setting `updated_at` to the close timestamp (a few seconds after `merged_at`)
4. The validator scores issue discovery for the miner
5. The issue is classified as "closed" instead of "solved"

Real example: [`entrius/gittensor-ui#172`](https://github.com/entrius/gittensor-ui/issues/172) — PR merged at `19:05:04Z`, issue auto-closed at `19:05:06Z`. The 2-second gap triggers the penalty.

## Expected Behavior

Issues auto-closed by a merged PR should count as "solved" for credibility and eligibility purposes. The post-merge edit detection should only penalise issues whose content was genuinely edited after both the merge and close events.

## Actual Behavior

The issue is reclassified from solved to closed, reducing the miner's `solved_count` and inflating `closed_count`. This lowers credibility and can prevent the issue from contributing to `valid_solved_count`, impacting eligibility.

Tet-9:
Fixes #443

## Problem
The post-merge edit detection in `_collect_issues_from_prs` used
`pr.merged_at` as the anchor for detecting post-merge body/title edits:

```python
if issue.body_or_title_edited_at > pr.merged_at:
When a PR merge auto-closes a linked issue, GitHub sets
body_or_title_edited_at to the close timestamp, which is always
slightly after merged_at (typically 1-5 seconds). This caused the
check to fire on the normal auto-close event, incorrectly penalizing
the discoverer by decrementing solved_count and incrementing
closed_count.
Real example: entrius/gittensor-ui#172 — PR merged at 19:05:04Z,
issue auto-closed at 19:05:06Z. The 2-second gap triggered the penalty.
Fix
Use closed_at as the edit anchor when it is available and occurs after
merged_at. This means only edits that happen after the issue was
closed are treated as post-merge gaming — the normal auto-close window
between merged_at and closed_at is excluded.
edit_anchor = issue.closed_at if issue.closed_at and issue.closed_at >= pr.merged_at else pr.merged_at
if issue.body_or_title_edited_at and issue.body_or_title_edited_at > edit_anchor:
If closed_at is unavailable, the behavior falls back to the original
merged_at anchor, preserving anti-gaming protection for legacy data.
Changes
gittensor/validator/issue_discovery/scoring.py — use closed_at
as edit anchor in post-merge edit detection

### DIFF
--- a/gittensor/validator/issue_discovery/scoring.py
+++ b/gittensor/validator/issue_discovery/scoring.py
@@ -248,7 +248,10 @@ def _collect_issues_from_prs(
 
                 # Anti-gaming: post-merge body/title edit detection
                 # Not issue.updated_at: it fires on bot comments, labels, reactions.
-                if issue.body_or_title_edited_at and pr.merged_at and issue.body_or_title_edited_at > pr.merged_at:
+                # Use closed_at as the anchor when available — edits between merged_at and
+                # closed_at are part of the normal auto-close flow, not post-merge gaming.
+                edit_anchor = issue.closed_at if issue.closed_at and issue.closed_at >= pr.merged_at else pr.merged_at
+                if issue.body_or_title_edited_at and issue.body_or_title_edited_at > edit_anchor:
                     bt.logging.info(
                         f'Issue #{issue.number} body/title edited after PR #{pr.number} merge — 0 score, counts as closed'
                     )


anderdc:
Closing. The bug this PR targets no longer exists, and the proposed change would weaken anti-gaming.

**The bug is already fixed:**

- Issue #443 was filed 2026-04-15 against the old code that used `issue.updated_at > pr.merged_at`. That field fires on auto-close events, bot comments, labels, and reactions, which is why false positives like `gittensor-ui#172` occurred.
- **PR #434 (merged 2026-04-16)** replaced the check with `issue.body_or_title_edited_at`, a precise field computed from GraphQL `userContentEdits` and title-rename events. Auto-close does NOT modify this field.
- Current code at `scoring.py:251` already uses the precise field. The cited real-example scenario cannot trigger the penalty against current code.

**The proposed change would create a gaming window:**

Moving the edit anchor from `merged_at` to `closed_at` (when `closed_at >= merged_at`) creates a 2-5 second window where body/title edits go undetected:

- PR merges at `19:05:04`, GitHub auto-closes issue at `19:05:06`
- Within that 2-second window, a miner could edit the issue body to better match the merged PR's actual content
- Under current code: `body_or_title_edited_at > merged_at` fires, penalty applied
- Under this PR's change: `body_or_title_edited_at > closed_at` does not fire, gaming goes undetected

The window is small, but the change deliberately loosens an anti-gaming check to "fix" a false positive that was already eliminated by using the precise edit field.

**Recommendation:**

Before opening PRs against older issues, verify the current code state. Many issues filed weeks ago have been fixed by unrelated PRs since.

PR #745, opened 2026-04-24
fix: apply review_quality_multiplier to open PR collateral

### LINKED ISSUE #744: [Bug] Open PR collateral ignores review_quality_multiplier
## Summary

Open-PR collateral is computed from three multipliers (`repo_weight`, `issue`, `label`) but omits `review_quality_multiplier`, diverging from merged-PR earned scoring. An open PR with many maintainer `CHANGES_REQUESTED` reviews is currently charged the same collateral as a clean open PR, overstating its projected merged earned score.

This is the same class of asymmetry fixed by #613 (label_multiplier) and #691 (state_reason gate) — open-PR collateral should reflect what the PR would have earned on merge.

## Root cause

Three coupled sites:

**1. `gittensor/classes.py:298-303`** — `changes_requested_count` is only counted for merged PRs, despite the GraphQL query (`gittensor/utils/github_api_tools.py:122`) fetching `changesRequestedReviews` unconditionally:

```python
changes_requested_count = 0
if is_merged:
    cr_reviews = pr_data.get('changesRequestedReviews', {}).get('nodes', [])
    changes_requested_count = sum(
        1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS
    )
```

This gate was introduced by #305 when the feature only applied to merged PRs. Since then, the data has been fetched but discarded for open PRs.

**2. `gittensor/validator/oss_contributions/scoring.py` — `calculate_pr_multipliers`** — hardcodes `pr.review_quality_multiplier = 1.0` for open PRs, so even if the count were populated it would be ignored.

**3. `gittensor/validator/oss_contributions/scoring.py:502-518` — `calculate_open_pr_collateral_score`** — the multipliers dict is missing `review_quality`, and the docstring explicitly lists only "repo_weight, issue, label" as applicable.

## Impact

An open PR with N maintainer `CHANGES_REQUESTED` reviews:

- **Current**: collateral charged as if `review_quality_multiplier = 1.0`
- **Expected**: collateral reduced by `max(0.0, 1.0 - REVIEW_PENALTY_RATE × N)`, matching what the merged earned score would be

For N=3 (→ 0.55 multiplier), current collateral is ~1.8× the projected merged earned score it was meant to approximate.

## Expected behavior

Open-PR collateral should apply the same `review_quality_multiplier` that the merged-PR earned score applies, using the same `changes_requested_count` source (`changesRequestedReviews` from GraphQL, restricted to `MAINTAINER_ASSOCIATIONS`).

Multipliers that remain merge-specific (`time_decay`, `credibility`) stay excluded — those are computed from merge-time state.

## Related precedents

- #612 → #613 (merged 2026-04-20) — same shape: added `label_multiplier` to open PR collateral for merged/open consistency
- #690 → #691 (merged 2026-04-22) — same shape: extended `is_valid_issue` state_reason gate to open PR collateral

## Environment

- Repository: `entrius/gittensor`
- Components: `gittensor/classes.py`, `gittensor/validator/oss_contributions/scoring.py`
- Query path: `gittensor/utils/github_api_tools.py` GraphQL `QUERY` already fetches `changesRequestedReviews` for every PR (no new API cost)

Crystora:
## Summary

`calculate_open_pr_collateral_score` applied `repo_weight`, `issue`, and `label` multipliers but silently ignored `review_quality_multiplier`. An open PR with maintainer `CHANGES_REQUESTED` reviews was charged the same collateral as a clean PR, overstating the merged earned score it is meant to approximate.

Three coupled sites were hiding the bug:

1. `gittensor/classes.py` - `changes_requested_count` was gated behind `if is_merged:`, so open PRs got `0` even though the GraphQL query already fetches `changesRequestedReviews` for every PR.
2. `gittensor/validator/oss_contributions/scoring.py` - `calculate_pr_multipliers` hardcoded `pr.review_quality_multiplier = 1.0` in the open-PR branch.
3. Same file - `calculate_open_pr_collateral_score` omitted `review_quality` from its multipliers dict.

## Changes

- Extract `_count_maintainer_changes_requested` in `classes.py` and call it unconditionally so both merged and open PRs get a real count.
- Lift `review_quality_multiplier` out of the `if is_merged:` branch in `calculate_pr_multipliers`.
- Add `'review_quality'` to the multipliers dict in `calculate_open_pr_collateral_score` and update its docstring.

No GraphQL changes - `changesRequestedReviews` is already fetched inline (#519), so there is no new API cost.

## Related precedents

- #613 - added `label_multiplier` to open-PR collateral (same shape)
- #691 - extended `is_valid_issue` state_reason gate to open-PR collateral (same shape)

## Related Issues

Closes #744 

## Type of Change

- [x] Bug fix

## Testing

- [x] Tests added/updated
- [x] Manually tested

21 tests in `tests/validator/test_review_quality_multiplier.py` pass. New coverage:

- `TestChangesRequestedCountFromGraphQL.test_open_pr_also_counts_maintainer_reviews` - flips the old assertion; open PRs now parse `changesRequestedReviews`.
- `TestReviewQualityMultiplierOnOpenPRCollateral` - direct collateral math under clean, penalized, and zero-multiplier conditions.
- `TestReviewQualityMultiplierThroughScoringPipeline` - end-to-end regression guards:
  - `calculate_pr_multipliers` applies `review_quality` to open PRs
  - merged and open PRs derive the same multiplier from the same CR count
  - open-PR collateral equals the projected merged earned score × `OPEN_PR_COLLATERAL_PERCENT`

Full validator suite: **254 passed**. `ruff check`, `ruff format`, and `pyright` all clean.

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -194,6 +194,8 @@ MAINTAINER_ASSOCIATIONS = ['OWNER', 'MEMBER', 'COLLABORATOR']
 
 # PR Review Quality Multiplier
 REVIEW_PENALTY_RATE = 0.15  # 15% deduction per CHANGES_REQUESTED review from a maintainer
+OPEN_PR_REVIEW_COLLATERAL_RATE = REVIEW_PENALTY_RATE  # Same per-review step, but increases open PR collateral
+MAX_OPEN_PR_REVIEW_COLLATERAL_MULTIPLIER = 2.0  # Cap open PR collateral growth from review iterations
 
 # Issue multiplier (flat values, no age scaling)
 STANDARD_ISSUE_MULTIPLIER = 1.33  # Non-maintainer issue author
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -194,7 +194,6 @@ MAINTAINER_ASSOCIATIONS = ['OWNER', 'MEMBER', 'COLLABORATOR']
 
 # PR Review Quality Multiplier
 REVIEW_PENALTY_RATE = 0.15  # 15% deduction per CHANGES_REQUESTED review from a maintainer
-OPEN_PR_REVIEW_COLLATERAL_RATE = REVIEW_PENALTY_RATE  # Same per-review step, but increases open PR collateral
 MAX_OPEN_PR_REVIEW_COLLATERAL_MULTIPLIER = 2.0  # Cap open PR collateral growth from review iterations
 
 # Issue multiplier (flat values, no age scaling)
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -305,12 +305,8 @@ class PullRequest:
         last_edited_at = parse_github_timestamp_to_cst(raw_edited_at) if isinstance(raw_edited_at, str) else None
         merged_at = parse_github_timestamp_to_cst(pr_data['mergedAt']) if is_merged else None
 
-        changes_requested_count = 0
-        if is_merged:
-            cr_reviews = pr_data.get('changesRequestedReviews', {}).get('nodes', [])
-            changes_requested_count = sum(
-                1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS
-            )
+        cr_reviews = (pr_data.get('changesRequestedReviews') or {}).get('nodes') or []
+        changes_requested_count = sum(1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS)
 
         current = {(n.get('name') or '').lower() for n in (pr_data.get('labels') or {}).get('nodes') or [] if n}
         label: Optional[str] = None
--- a/gittensor/validator/oss_contributions/mirror/scored_pr.py
+++ b/gittensor/validator/oss_contributions/mirror/scored_pr.py
@@ -72,6 +72,12 @@ class ScoredMirrorPR:
         return self.pr.repo_full_name
 
     @property
+    def changes_requested_count(self) -> int:
+        """Alias for the maintainer-only CHANGES_REQUESTED count used by
+        source-agnostic scoring helpers."""
+        return self.pr.review_summary.maintainer_changes_requested_count
+
+    @property
     def merged_at(self) -> Optional[datetime]:
         """Alias for ``self.pr.merged_at`` — matches legacy PullRequest attribute
         name so the unified pioneer-dividend walk treats both types identically."""
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -24,7 +24,6 @@ from gittensor.constants import (
     MAX_OPEN_PR_REVIEW_COLLATERAL_MULTIPLIER,
     MAX_OPEN_PR_THRESHOLD,
     OPEN_PR_COLLATERAL_PERCENT,
-    OPEN_PR_REVIEW_COLLATERAL_RATE,
     OPEN_PR_THRESHOLD_TOKEN_SCORE,
     PIONEER_DIVIDEND_MAX_RATIO,
     PIONEER_DIVIDEND_RATE_1ST,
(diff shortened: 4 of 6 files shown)


Crystora:
@anderdc Pls review this PR.

anderdc:
We generally just add 'multipliers' to the collateral. I'm not sure if review penalty should be added. Let me chew on it

Crystora:
> We generally just add 'multipliers' to the collateral. I'm not sure if review penalty should be added. Let me chew on it

`review_quality` is also part of the merged-PR formula. It just wasn't being applied to collateral. So the open-PR projection didn't match the merged-PR reality - that's the inconsistency I'm fixing.

Adding `review_quality` to collateral actually **lowers** the deduction on PRs with changes-requested reviews. It's not a new penalty. Clean PRs (no CR reviews) see zero change.

The reason `time_decay` and `credibility` aren't in collateral is they need merge data we don't have yet. `review_quality` doesn't have that problem - the data is already there for open PRs.

anderdc:
The point of collateral is to 'put your money where your mouth is' on a PR, with this addition you're more ok with review iterations because it eases up on your collateral (at the cost of your resulting PR). So still debating, the argument isn't particularly compelling but also not dismissed. Any more thoughts?

Crystora:
Fair point. But the thing is, collateral already multiplies by `repo_weight`, `issue`, and `label` today. So a refactor PR (0.5×) already pays less collateral than a feature PR (1.5×) for the same work. By the "fixed deposit" logic, those probably shouldn't be there either.
I just wanted `review_quality` to match what the other three already do. If you'd rather pull all four out and make collateral a flat `base × 20%`, I'm happy to flip the PR that way instead. Just let me know which option you want.

Crystora:
On the "more ok with review iterations" concern - I don't think it actually plays out that way.

The case we care about deterring is someone opening a bunch of speculative PRs and walking away. Those PRs haven't been touched by a maintainer yet, so `review_quality` is 1.0 and their collateral is unchanged. The spam protection still works the same way.

What does change is collateral on PRs that a maintainer actively reviewed and asked for changes on. Today those get hit twice: full collateral while they're open, AND lower earned score if they merge. With this change they get hit once, and the open-side collateral matches what the merged PR will actually be worth.

On the multipliers question more broadly: collateral today isn't a flat deposit, it's `base × repo × issue × label × 20%`. So it's already trying to estimate merged value. `review_quality` is just another part of that estimate. Either all four belong or none do - having three of them is what feels arbitrary.

#613 (label_multiplier) and #691 (state_reason gate) added missing pieces to collateral for the same consistency reason, so there's precedent.

If you'd actually prefer to strip all four and make collateral flat `base × 20%`, I can open that as a separate PR. But this one is the smaller step in the direction collateral already goes.

LandynDev:
We should not make collateral less for an open PR that has requested changes. If we do this, we are incentivizing open PRs to go through iterations as it becomes **less** of a tax on miners. I'd prefer if the inverse happened and it became more of a tax, if anything. The goal is to find ways to have the network produce consistently merge-able PRs from the second they're made (0 iterations/requested changes), this doesn't align with that goal.

Yes, it's more 'accurate' and 1:1 with what the merged score would be.
But miner failing to do their job (perfect pr with 0 iterations) should not alleviate the collateral burden imo

Crystora:
> The goal is to find ways to have the network produce consistently merge-able PRs from the second they're made (0 iterations/requested changes), this doesn't align with that goal.

@LandynDev Thanks for clarifying. I agree collateral should not become lower when a PR gets requested changes. If not, it can be another spam gap.
Instead of reducing the collateral, I’ll consider increasing when requesting changes: 
Simple direction I propose:
- clean PR: no change
- PR with requested changes: higher collateral, not lower.

```
review_penalty_count = number of maintainer CHANGES_REQUESTED reviews
review_collateral_multiplier = 1 + (review_penalty_count × 0.15)
open_pr_collateral =
base_score × repo_weight × issue × label × review_collateral_multiplier × 20%

```

```
Example:
0 requested changes:
100 × 1.00 × 20% = 20

1 requested change:
100 × 1.15 × 20% = 23

2 requested changes:
100 × 1.30 × 20% = 26

3 requested changes:
100 × 1.45 × 20% = 29
```

I would also cap it so it cannot grow forever:
`review_collateral_multiplier = min(2.0, 1 + requested_changes_count × 0.15)`
So max collateral from review iterations is 2x the normal collateral.

That keeps the incentive aligned with merge-ready PRs while still using the review quality signal.

-----

What do you think about it? @anderdc @LandynDev 
I'd appreciate any feedback

anderdc:
Let's actually do that and see how it goes, that makes more sense to me. The goal is mergeable quality PRs from the get-go.

Crystora:
@anderdc Ready to review

gittensor/constants.py
```diff
@@ -194,6 +194,8 @@
 
 # PR Review Quality Multiplier
 REVIEW_PENALTY_RATE = 0.15  # 15% deduction per CHANGES_REQUESTED review from a maintainer
+OPEN_PR_REVIEW_COLLATERAL_RATE = REVIEW_PENALTY_RATE  # Same per-review step, but increases open PR collateral
```
anderdc: no need for another separate constant for this, let's just make them both use `REVIEW_PENALTY_RATE`

anderdc (changes requested):

Crystora:
@anderdc I've updated. Pls check again.

anderdc (approved):

PR #756, opened 2026-04-24
fix: isolate concurrent_forward failures so siblings aren't cancelled

### LINKED ISSUE #755: [Bug] concurrent_forward() cancels all sibling forwards when one raises
## Description

In `neurons/base/validator.py:105-107`, `concurrent_forward()` calls `asyncio.gather(*coroutines)` without `return_exceptions=True`. As soon as any single `self.forward()` coroutine raises, `asyncio.gather` propagates the first exception and cancels every other in-flight forward. With `--neuron.num_concurrent_forwards > 1`, a transient failure in one forward (network blip, chain-sync hiccup, unhandled exception in reward computation) discards the work already done by its siblings — queries sent to miners, responses collected, scores computed — and the surviving sibling errors are never surfaced. The outer handler in `run()` logs only the first exception and continues, so operators have no visibility into per-coroutine failures.

The default value of `num_concurrent_forwards` is `1`, so default deployments are unaffected; the bug bites operators who raise concurrency to improve throughput.

## Steps to Reproduce

1. Configure a validator with `--neuron.num_concurrent_forwards 4` (or any value > 1).
2. Arrange for one of the concurrent `forward()` calls to raise an exception (e.g. inject a transient network error in the `dendrite` / reward pipeline, or patch `self.forward` in a test to raise on the 2nd invocation).
3. Observe the behavior of the other three concurrent forwards.

## Expected Behavior

- Each concurrent forward runs independently to completion.
- A failure in one forward is logged with its index and cause, and does **not** cancel sibling forwards.
- Work done by sibling forwards (scoring, weight contributions) is preserved.

## Actual Behavior

- The first exception propagates out of `asyncio.gather`, cancelling all in-flight sibling coroutines.
- Sibling work is silently discarded; only the first exception is visible to the outer handler in `run()`, and the identity of the failing coroutine is not logged.
- The validator continues to the next step having effectively lost multiple cycles of miner scoring for that block.

## Environment

- Repository: `entrius/gittensor` (Python validator codebase)
- File: `neurons/base/validator.py:105-107`
- Affects: any validator run with `--neuron.num_concurrent_forwards > 1`
- Python / asyncio behavior is standard; no version-specific dependency.

## Additional Context

This preserves the output of successful sibling forwards and makes each failure individually visible. `BaseException` is appropriate here (rather than `Exception`) so that `CancelledError` from a true shutdown request is still surfaced in the logs rather than swallowed as a "success". Consider also logging the traceback at debug level (matching the pattern already used in `BaseValidatorNeuron.run()`).

edwin-rivera-dev:
## Summary

`BaseValidatorNeuron.concurrent_forward()` called `asyncio.gather(*coroutines)` without `return_exceptions=True`, so a single `forward()` raising would propagate the first exception and **cancel every other in-flight forward**. With `--neuron.num_concurrent_forwards > 1`, one transient failure discarded the sibling coroutines' already-completed work (queries sent to miners, responses scored) and surfaced only the first exception to the outer handler in `run()`, with no indication which coroutine failed.

This PR switches to `asyncio.gather(..., return_exceptions=True)` and logs each failed coroutine individually by index, with a full traceback at debug level. Successful sibling forwards now run to completion and their scoring work is preserved.

Changes:
- `neurons/base/validator.py:105-113` — pass `return_exceptions=True`, iterate results, log each `BaseException` with its coroutine index (`error`) and traceback (`debug`), matching the existing logging pattern in `run()`.
- `BaseException` is used instead of `Exception` so that `asyncio.CancelledError` during a real shutdown is still visible rather than silently swallowed.

## Related Issues

Closes #755 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [ ] Tests added/updated
- [x] Manually tested

**Manual verification**

Ran a focused async reproduction against `asyncio.gather` semantics to confirm the behavior change, since the bug is isolated to the gather call and does not require network/chain state:

- Before the patch: gathered 4 coroutines where the 2nd raises — `asyncio.gather` propagates the exception and the remaining 3 report `CancelledError`; any work they had done is lost.
- After the patch: same 4 coroutines, 2nd raises — `return_exceptions=True` collects all 4 results, the 3 siblings complete normally, and the 2nd's exception is logged as `concurrent_forward[1] raised: <err>` with the traceback at debug level.

No changes to the validator's public surface, config flags, or scoring behavior on the happy path; the default `--neuron.num_concurrent_forwards=1` run path is unchanged in observable behavior (a single failure is still logged and `run()` continues).

**Tests not added**

Per `CONTRIBUTING.md` ("We do NOT accept PRs that are only testing changes/additions"), and because the change is a two-line contract adjustment on a stdlib call whose semantics are exhaustively covered in CPython, no new unit test is added. Happy to add an async test that monkeypatches `self.forward` to raise on one index and asserts siblings complete, if maintainers prefer explicit coverage of this call site.

## Checklist

- [x] Self-review completed

### DIFF
--- a/neurons/base/validator.py
+++ b/neurons/base/validator.py
@@ -104,7 +104,13 @@ class BaseValidatorNeuron(BaseNeuron):
 
     async def concurrent_forward(self):
         coroutines = [self.forward() for _ in range(self.config.neuron.num_concurrent_forwards)]
-        await asyncio.gather(*coroutines)
+        # return_exceptions=True so one forward's failure does not cancel its siblings
+        # and discard their already-completed work.
+        results = await asyncio.gather(*coroutines, return_exceptions=True)
+        for i, result in enumerate(results):
+            if isinstance(result, BaseException):
+                bt.logging.error(f'concurrent_forward[{i}] raised: {result}')
+                bt.logging.debug(str(print_exception(type(result), result, result.__traceback__)))
 
     def run(self):
         """


edwin-rivera-dev:
Hi, @anderdc 

Could you tell me the reason of the closed state? I think this is the clear PR that have to be fixed.
I will look forward to you.
Appreciate

PR #766, opened 2026-04-24
fix: classify Android androidTest/ sources and *IT.java/*IT.kt as test files

### LINKED ISSUE #765: [Bug] `is_test_file` misses the Android `androidTest/` source set and the `IT` integration-test filename suffix
## Description

`FileChange.is_test_file()` in `gittensor/classes.py:70-93` decides whether a file is a test by matching a few regex patterns:

```python
test_dir_patterns = [r'(^|/)tests?/', r'(^|/)__tests?__/']
test_patterns = [
    r'^test_', r'^spec_',
    r'_test\.[^.]+$', r'_tests\.[^.]+$',
    r'\.test\.[^.]+$', r'\.tests\.[^.]+$',
    r'\.spec\.[^.]+$',
    r'^test\.[^.]+$', r'^tests\.[^.]+$',
]
```

Neither set covers the Android Gradle convention. Android puts instrumented (on-device) tests in `app/src/androidTest/...`, separate from JVM unit tests in `app/src/test/...`. The directory regex matches `test/` but not `androidTest/`. The filename regexes don't cover the `*IT.java` / `*IT.kt` suffix used for those integration tests.

`nextcloud/android` is whitelisted in `gittensor/validator/weights/master_repositories.json` with `weight: 0.0778` and follows this layout. Every test file in its `androidTest/` tree is currently classified as production code, then scored at full Java/Kotlin language weight (1.75) instead of `lang_weight × TEST_FILE_CONTRIBUTION_WEIGHT (0.05)` in `gittensor/validator/utils/tree_sitter_scoring.py:276`. That is a 20× over-score per token for those files.

## Steps to Reproduce

1. Pick a real file from `nextcloud/android`, for example `app/src/androidTest/java/com/owncloud/android/UploadIT.java` (listed at `https://api.github.com/repos/nextcloud/android/contents/app/src/androidTest/java/com/owncloud/android`).
2. Construct a `FileChange` for it and call `is_test_file()`:

   ```python
   from gittensor.classes import FileChange

   fc = FileChange(
       pr_number=0,
       repository_full_name='nextcloud/android',
       filename='app/src/androidTest/java/com/owncloud/android/UploadIT.java',
       changes=100, additions=100, deletions=0, status='added',
   )
   print(fc.is_test_file())   # prints False
   ```

3. The same return value flows into `calculate_token_score_from_file_changes`, which sets `file_weight = 1.0` instead of `TEST_FILE_CONTRIBUTION_WEIGHT = 0.05` for the file's breakdown.

## Expected Behavior

`is_test_file()` returns `True` for any path under a Gradle test source-set directory (`androidTest/`, `androidTestGeneric/`, `androidTestGplay/`, etc.) regardless of the filename inside it.

Optionally, it also returns `True` for filenames using the standard Maven Failsafe / Android integration-test suffix (`*IT.java` / `*IT.kt`), with the case-preservation note in *Additional Context* below.

## Actual Behavior

`is_test_file()` returns `False` for all of these real files in `nextcloud/android`. Each was confirmed to exist via the GitHub Contents API at the time of writing:

`app/src/androidTest/java/com/owncloud/android/`
- `AbstractIT.java`
- `AbstractOnServerIT.java`
- `DownloadIT.java`
- `EncryptionIT.kt`
- `FileIT.java`
- `ScreenshotsIT.kt`
- `UploadIT.java`

`app/src/androidTest/java/com/nextcloud/client/`
- `ActivitiesFragmentIT.kt`
- `AuthenticatorActivityIT.java`
- `CommunityFragmentIT.kt`
- `FileDisplayActivityIT.kt`
- `FileDisplayActivityScreenshotIT.kt`
- `FirstRunActivityIT.kt`
- `SettingsActivityIT.kt`
- `SyncedFoldersActivityIT.kt`
- `UploadListActivityActivityIT.kt`

Each gets `file_weight = 1.0`. With `nextcloud/android` repo weight `0.0778` and Java/Kotlin language weight `1.75`, every token in these files contributes `1.75 × 0.0778 ≈ 0.136` to the file score instead of the intended `0.05 × 1.75 × 0.0778 ≈ 0.0068` — about a 20× inflation. The same applies to any future PR that touches them.

## Environment

- OS: any (pure-Python regex logic, environment-independent)
- Python version: 3.12
- Commit: `0ecafea` on `test`

## Additional Context

The closest precedent is #311 → #314 (merged 2026-04-01), which extended test detection to Rust/Zig/D inline tests for the same reason: a real class of test files was being scored at full production weight. This is the source-set / filename-suffix counterpart for Android.

### Suggested fix (primary, zero-risk)

Add `androidTest/` (and adjacent Gradle `*Test/` source-sets) to `test_dir_patterns`. The basename is lowercased before matching at line 71, so the addition is one regex:

```python
test_dir_patterns = [
    r'(^|/)tests?/',
    r'(^|/)__tests?__/',
    r'(^|/)androidtest[a-z]*/',   # androidTest, androidTestGeneric, androidTestGplay, etc.
    r'(^|/)integrationtest/',     # Gradle integrationTest source set
]
```

This alone catches every one of the 16 real files listed above — `nextcloud/android` puts all integration tests under `app/src/androidTest/...`. No other change is strictly required for the evidenced impact.

### Implementation note for the IT-suffix half

A naive addition like `r'it\.(java|kt|kts)$'` to `test_patterns` would produce many false positives because `is_test_file` lowercases the basename before matching. I tested it against common Java/Kotlin type names; **it incorrectly classifies all of these as test files**:

```
Edit.java     Audit.java    Commit.kt     Permit.java   Submit.kt
Implicit.java Limit.kt      Unit.kt       Wait.java     Init.kt
Visit.kt      Profit.java   Posit.kt      Hit.java      Split.kt
```

A correct IT-suffix detector needs to operate on the **original-case** basename and require a CamelCase boundary, e.g. `[A-Z][A-Za-z0-9]*IT\.(java|kt|kts)$`. This is a slightly bigger structural change to `is_test_file()` (split case-sensitive vs case-insensitive matching) and may not be worth the complexity given that the directory-pattern fix above already covers every cited file.

**Two reasonable scopes for the eventual PR:**

1. **Minimal**: only add the `androidTest/` directory pattern. Closes the entire evidenced gap with one regex, zero false-positive risk.
2. **Full**: add the directory pattern AND a case-preserving CamelCase IT-suffix detector. Catches the same files plus any future Java/Kotlin/C# test-naming convention that doesn't live under a `*Test/` directory.

Crystora:
Fixes #765

## Summary

`FileChange.is_test_file()` missed three real-world test conventions, causing those files to be scored at full production weight (1.0) instead of `TEST_FILE_CONTRIBUTION_WEIGHT` (0.05) — a 20× over-score per token:

- **Gradle Android test source sets**: `androidTest/`, `androidTestGeneric/`, `androidTestGplay/`, etc.
- **Gradle integration-test source sets**: `integrationTest/` — also covers Maven Failsafe `*IT.java`/`.kt`/`.kts` files, which by convention live under `src/test/` or `src/integrationTest/` and are caught by the directory rule rather than a filename-suffix rule.
- **Ruby RSpec**: the `spec/` directory and the `*_spec.{rb,js,ts,…}` filename suffix.

Impact is concrete: `nextcloud/android` keeps its integration tests under `app/src/androidTest/...` (16+ files like `UploadIT.java`, `AbstractIT.java`, `ActivitiesFragmentIT.kt`); Ruby projects keep specs under `spec/models/account_spec.rb`. Each was previously mis-scored at production weight.

## Changes

- `gittensor/classes.py`:
  - Added `(^|/)androidtest[a-z]*/`, `(^|/)integrationtest/`, and `(^|/)spec/` to the directory pattern set.
  - Added `_spec\.[^.]+$` to the basename pattern set.
  - Promoted both pattern sets to module-level `re.compile`-backed tuples (`_TEST_DIR_PATTERNS`, `_TEST_BASENAME_PATTERNS`) and extracted two small helpers (`_matches_test_dir_pattern`, `_matches_test_basename_pattern`) — avoids re-allocating and re-compiling the lists on every call to `is_test_file()`.
- `tests/test_classes.py`: 23 new parametrized assertions across 3 test functions, plus a regression check covering the pre-existing conventions — Gradle Android & integrationTest paths, Ruby RSpec `spec/` paths and `_spec.*` suffixes, and lookalike negatives (`Edit.java`, `build.gradle.kts`, `docs/androidtest.md`, `specification.rb`, `spectrum.rb`, `spec.rb`).

## Test Plan

- [x] `pytest tests/test_classes.py` — 25 passed
- [x] `pytest tests/` — 676 passed, 0 regressions
- [x] `pre-commit run --all-files --hook-stage pre-push` — ruff, pyright, pytest all pass
- [x] Verified the issue reproducer (`app/src/androidTest/java/...IT.java`) now returns `True`
- [x] Verified false-positive candidates (`Edit.java`, `Commit.java`, `Unit.kt`, `build.gradle.kts`, `docs/androidtest.md`, `specification.rb`, `spectrum.rb`, `spec.rb`) still return `False`

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -45,36 +45,6 @@ class Miner:
         return f'Miner(uid={self.uid}, hotkey={self.hotkey[:8]}..., github_id={self.github_id})'
 
 
-_TEST_DIR_PATTERNS = (
-    re.compile(r'(^|/)tests?/'),
-    re.compile(r'(^|/)__tests?__/'),
-    re.compile(r'(^|/)androidtest[a-z]*/'),
-    re.compile(r'(^|/)integrationtest/'),
-    re.compile(r'(^|/)spec/'),
-)
-
-_TEST_BASENAME_PATTERNS = (
-    re.compile(r'^test_'),
-    re.compile(r'^spec_'),
-    re.compile(r'_test\.[^.]+$'),
-    re.compile(r'_tests\.[^.]+$'),
-    re.compile(r'_spec\.[^.]+$'),
-    re.compile(r'\.test\.[^.]+$'),
-    re.compile(r'\.tests\.[^.]+$'),
-    re.compile(r'\.spec\.[^.]+$'),
-    re.compile(r'^test\.[^.]+$'),
-    re.compile(r'^tests\.[^.]+$'),
-)
-
-
-def _matches_test_dir_pattern(path_lower: str) -> bool:
-    return any(pattern.search(path_lower) for pattern in _TEST_DIR_PATTERNS)
-
-
-def _matches_test_basename_pattern(basename_lower: str) -> bool:
-    return any(pattern.search(basename_lower) for pattern in _TEST_BASENAME_PATTERNS)
-
-
 @dataclass
 class FileChange:
     """Represents a single file change in a PR"""
@@ -106,9 +76,31 @@ class FileChange:
     def is_test_file(self) -> bool:
         filename_lower = self.filename.lower()
         basename = filename_lower.split('/')[-1]
-        if _matches_test_dir_pattern(filename_lower):
+
+        test_dir_patterns = [
+            r'(^|/)tests?/',
+            r'(^|/)__tests?__/',
+            r'(^|/)androidtest[a-z]*/',
+            r'(^|/)integrationtest/',
+            r'(^|/)spec/',
+        ]
+        if any(re.search(pattern, filename_lower) for pattern in test_dir_patterns):
             return True
-        return _matches_test_basename_pattern(basename)
+
+        test_patterns = [
+            r'^test_',
+            r'^spec_',
+            r'_test\.[^.]+$',
+            r'_tests\.[^.]+$',
+            r'_spec\.[^.]+$',
+            r'\.test\.[^.]+$',
+            r'\.tests\.[^.]+$',
+            r'\.spec\.[^.]+$',
+            r'^test\.[^.]+$',
+            r'^tests\.[^.]+$',
+        ]
+
+        return any(re.search(pattern, basename) for pattern in test_patterns)
 
     @classmethod
     def from_github_response(cls, pr_number: int, repository_full_name: str, file_diff: DefaultDict) -> 'FileChange':
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -45,6 +45,36 @@ class Miner:
         return f'Miner(uid={self.uid}, hotkey={self.hotkey[:8]}..., github_id={self.github_id})'
 
 
+_TEST_DIR_PATTERNS = (
+    re.compile(r'(^|/)tests?/'),
+    re.compile(r'(^|/)__tests?__/'),
+    re.compile(r'(^|/)androidtest[a-z]*/'),
+    re.compile(r'(^|/)integrationtest/'),
+    re.compile(r'(^|/)spec/'),
+)
+
+_TEST_BASENAME_PATTERNS = (
+    re.compile(r'^test_'),
+    re.compile(r'^spec_'),
+    re.compile(r'_test\.[^.]+$'),
+    re.compile(r'_tests\.[^.]+$'),
+    re.compile(r'_spec\.[^.]+$'),
+    re.compile(r'\.test\.[^.]+$'),
+    re.compile(r'\.tests\.[^.]+$'),
+    re.compile(r'\.spec\.[^.]+$'),
+    re.compile(r'^test\.[^.]+$'),
+    re.compile(r'^tests\.[^.]+$'),
+)
+
+
+def _matches_test_dir_pattern(path_lower: str) -> bool:
+    return any(pattern.search(path_lower) for pattern in _TEST_DIR_PATTERNS)
+
+
+def _matches_test_basename_pattern(basename_lower: str) -> bool:
+    return any(pattern.search(basename_lower) for pattern in _TEST_BASENAME_PATTERNS)
+
+
 @dataclass
 class FileChange:
     """Represents a single file change in a PR"""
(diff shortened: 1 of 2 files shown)


anderdc (changes requested):
Drop the IT-suffix regex — the new androidtest[a-z]*/ and integrationtest/ directory patterns cover every real-world case (Maven Failsafe puts *IT.java under src/test/, Gradle Android puts them under androidTest/). The IT-suffix branch only catches hypothetical files outside any test directory and risks false-positives on production classes named *IT.{java,kt,kts}.

Crystora:
> Drop the IT-suffix regex — the new androidtest[a-z]*/ and integrationtest/ directory patterns cover every real-world case (Maven Failsafe puts *IT.java under src/test/, Gradle Android puts them under androidTest/). The IT-suffix branch only catches hypothetical files outside any test directory and risks false-positives on production classes named *IT.{java,kt,kts}.

Fixed all!

anderdc (changes requested):
Drop the module-level `_TEST_DIR_PATTERNS` / `_TEST_BASENAME_PATTERNS` tuples and the `_matches_test_dir_pattern` / `_matches_test_basename_pattern` helpers — `is_test_file` is the only caller, so inlining the lists the way the function had them before matches the existing style. Keep the new patterns (`(^|/)androidtest[a-z]*/`, `(^|/)integrationtest/`, `(^|/)spec/`, `_spec\.[^.]+$`) inline.

Crystora:
> Drop the module-level `_TEST_DIR_PATTERNS` / `_TEST_BASENAME_PATTERNS` tuples and the `_matches_test_dir_pattern` / `_matches_test_basename_pattern` helpers — `is_test_file` is the only caller, so inlining the lists the way the function had them before matches the existing style. Keep the new patterns (`(^|/)androidtest[a-z]*/`, `(^|/)integrationtest/`, `(^|/)spec/`, `_spec\.[^.]+$`) inline.

Updated!

anderdc (approved):
-

PR #774, opened 2026-04-24
fix: use GraphQL-only issue PR lookup

### LINKED ISSUE #773: [Bug] `_search_issue_referencing_prs_rest surfaces` false-positive PRs — bare-number query with no #N verification
## Description

`_search_issue_referencing_prs_rest` in `gittensor/utils/github_api_tools.py:500-557` is the REST fallback used by `find_prs_for_issue` whenever the GraphQL path returns empty or a PAT is unavailable. It builds its query with the issue number as a **bare token**:

```python
params={'q': f'repo:{repo} type:pr{state_clause} {issue_number} in:title,body', 'per_page': '50'},
```

GitHub's search then matches any PR whose title or body contains that number for unrelated reasons — version bumps (`Bump foo 3.41 -> 3.42`), percentages (`42% faster`), counts (`fix 42 failing tests`), ports, dates, etc. The response items are then mapped straight into `PRInfo` with no post-filter that would verify the PR actually references `#<issue_number>`:

```python
for item in resp.json().get('items', []):
    number = item.get('number')
    if number is None:
        continue
    user = item.get('user') or {}
    pr_info: PRInfo = {
        'number': number,
        'title': item.get('title') or '',
        ...
    }
    out.append(pr_info)
```

Contrast with `_search_issue_referencing_prs_graphql` (same file, ~L424), which queries `timelineItems(itemTypes: [CROSS_REFERENCED_EVENT], first: 50)` — GitHub's own cross-reference resolver, so the result set is already semantically verified. The REST path should converge on the same semantic guarantee.

## Steps to Reproduce

1. Pick any repository where the issue number `N` also appears as a bare digit in unrelated PR titles/bodies (version strings, percentages, line counts). Any active repo with a low-numbered issue reproduces this easily.
2. Call `_search_issue_referencing_prs_rest` directly, or drive it through the CLI without a PAT:

   ```bash
   unset GITTENSOR_MINER_PAT
   uv run gitt issues submissions --id <N>
   ```

   (The unauth REST path is always reached when no PAT is set; the authenticated REST path is reached whenever GraphQL returns an empty list for an issue with no real cross-references, or when GraphQL errors out.)

3. Inspect the submissions table — unrelated PRs whose body merely contains the digit string `N` are listed as "submissions" for issue `#N`.

## Expected Behavior

Only PRs that actually reference the issue (via `#N` or a GitHub issue/PR URL pointing at `N`) are returned, matching what the GraphQL `CROSS_REFERENCED_EVENT` path returns.

## Actual Behavior

Any PR whose title or body contains the bare number `N` as a search token is returned as a submission, including:

- `Bump numpy from 3.41 to 3.42` (matches `42`)
- `Improve perf by 42%`
- `Fix 42 failing tests`
- `Closes #421` (digit `42` still appears in `421` for some tokenizations)

Users see unrelated PRs in the `gitt issues submissions` table and in the `--json` payload consumed by downstream scripts.

## Impact

- User-facing: `gitt issues submissions` is misleading, especially for users without a `GITTENSOR_MINER_PAT` (those users always hit the REST fallback).
- Downstream: the JSON payload at `gittensor/cli/issue_commands/submissions.py:89-112` propagates the bad data to any automation consuming it.
- Validator scoring is **not** affected — `find_solver_from_cross_references` (`gittensor/utils/github_api_tools.py:1036`) uses the GraphQL path directly and never falls through to REST. This is a CLI data-quality bug, not a scoring bug.

mkdev5:
## Summary

REST issue/PR search matched the issue number as a bare text token, which can surface unrelated PRs whose title/body happen to contain that number. Rather than layering regex filtering on top of that imprecise source, this PR removes `_search_issue_referencing_prs_rest` and makes `find_prs_for_issue` rely only on GraphQL cross-reference timeline data.

Behavior now:

- No PAT: returns an empty list instead of risking wrong REST search data, and the human CLI output tells users to set `GITTENSOR_MINER_PAT`.
- PAT + GraphQL empty: returns the correct empty cross-reference set.
- PAT + GraphQL error: returns empty after the existing GraphQL retry/backoff path.

Validator scoring remains unaffected because `find_solver_from_cross_references` already uses GraphQL-only cross-reference data and never used the REST fallback.

## Related Issues

Closes #773

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Ran:

- `uv run pytest tests/utils/test_github_api_tools.py tests/cli/test_issue_submission.py` -> **82 passed**
- `uv run ruff check gittensor/utils/github_api_tools.py gittensor/cli/issue_commands/helpers.py tests/utils/test_github_api_tools.py tests/cli/test_issue_submission.py` -> passed

## Checklist

- [x] Self-review completed


anderdc (changes requested):
The premise is right (REST bare-token search returns false positives) but the fix keeps the broken path and adds a regex on top. Cleaner: delete `_search_issue_referencing_prs_rest` and have `find_prs_for_issue` short-circuit on GraphQL alone.

- No-PAT CLI users get an empty list instead of a wrong one — wrong data is worse than no data for an inspection command.
- PAT users on GraphQL-empty get correct empty (the cross-reference set is genuinely empty, which is what GraphQL just told us).
- PAT users on GraphQL-error get empty after the existing retry/backoff in `execute_graphql_query` / `get_github_graphql_query` (max_attempts=8 in the latter) — no need for a second data source.

Issue body confirms scoring isn't affected (`find_solver_from_cross_references` uses GraphQL only and never falls through), which bounds the blast radius. Drop the regex helper, the filter, and the two new test classes; keep the existing `find_prs_for_issue` tests and confirm they still pass under the simpler cascade.

Optional UX improvement: surface a "set GITTENSOR_MINER_PAT" hint in `gitt issues submissions` when no PAT is set, so the empty-list case is actionable instead of silent.

mkdev5:
> The premise is right (REST bare-token search returns false positives) but the fix keeps the broken path and adds a regex on top. Cleaner: delete `_search_issue_referencing_prs_rest` and have `find_prs_for_issue` short-circuit on GraphQL alone.
> 
> * No-PAT CLI users get an empty list instead of a wrong one — wrong data is worse than no data for an inspection command.
> * PAT users on GraphQL-empty get correct empty (the cross-reference set is genuinely empty, which is what GraphQL just told us).
> * PAT users on GraphQL-error get empty after the existing retry/backoff in `execute_graphql_query` / `get_github_graphql_query` (max_attempts=8 in the latter) — no need for a second data source.
> 
> Issue body confirms scoring isn't affected (`find_solver_from_cross_references` uses GraphQL only and never falls through), which bounds the blast radius. Drop the regex helper, the filter, and the two new test classes; keep the existing `find_prs_for_issue` tests and confirm they still pass under the simpler cascade.
> 
> Optional UX improvement: surface a "set GITTENSOR_MINER_PAT" hint in `gitt issues submissions` when no PAT is set, so the empty-list case is actionable instead of silent.

just pushed the update, can you please review again?

anderdc (approved):

PR #778, opened 2026-04-24
fix: make store_evaluation atomic by deferring sub-write commits to outer transaction 

### LINKED ISSUE #749: [Bug] store_evaluation is non-atomic: sub-writes commit before outer rollback
## Summary

`DatabaseStorage.store_evaluation` ([`gittensor/validator/utils/storage.py:47-89`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/utils/storage.py#L47)) sets `autocommit = False` and wraps a sequence of `Repository` bulk writes in what looks like a single transaction, but each sub-write commits independently. The outer `rollback()` cannot undo any earlier sub-write, so a mid-evaluation failure leaves a partial write in the database while the caller sees `success=False`.

## Where the commits happen

Methods called inside the supposed-transaction block, each calling `self.db.commit()` before returning:

- `set_miner` at [`repository.py:107`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/storage/repository.py#L107) (via `execute_command` at line 64)
- `store_pull_requests_bulk` at [`repository.py:209`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/storage/repository.py#L209)
- `store_issues_bulk` at [`repository.py:264`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/storage/repository.py#L264)
- `store_file_changes_bulk` at [`repository.py:313`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/storage/repository.py#L313)
- `set_miner_evaluation` at [`repository.py:369`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/storage/repository.py#L369)

## Steps to reproduce

1. Run a validator with `STORE_DB_RESULTS=true` against a Postgres backing store.
2. Force any sub-write inside `store_evaluation` to raise. Easiest is to insert a conflicting `issues` row out of band so the next `BULK_UPSERT_ISSUES` violates a constraint the upsert clause does not resolve.
3. Observe `store_evaluation` returns `StorageResult(success=False, errors=['Failed to store evaluation data for UID X: ...'])`.
4. Query the DB: rows from the earlier `store_pull_requests_bulk` call are still present even though the round was reported as a failure.

## Expected

When `store_evaluation` reports failure, the database state should match its state at function entry. None of the sub-writes from the failing round should remain.

## Actual

The PR rows from the earlier sub-write are committed. The matching `miner_evaluations` row is absent. Subsequent rounds see orphaned PRs and inconsistent counts.

## Additional context

PR [#668](https://github.com/entrius/gittensor/pull/668) attempted a fix and was closed with the request to file a tracking issue first.

Tet-9:
Closes #749

## Problem

`DatabaseStorage.store_evaluation` (`gittensor/validator/utils/storage.py:47-89`) sets `autocommit = False` and wraps a sequence of `Repository` bulk writes in what appears to be a single atomic transaction. However, each sub-write method commits independently to the database before returning, making the outer `rollback()` completely ineffective against any earlier sub-write.

The following methods each call `self.db.commit()` internally before returning control to `store_evaluation`:

- `set_miner` — `repository.py:107` (via `execute_command` at line 64)
- `store_pull_requests_bulk` — `repository.py:209`
- `store_issues_bulk` — `repository.py:264`
- `store_file_changes_bulk` — `repository.py:313`
- `set_miner_evaluation` — `repository.py:369`

This means if any sub-write raises mid-evaluation, the outer `except` block calls `rollback()` — but all previously committed sub-writes are already durably flushed to disk. The caller receives `StorageResult(success=False)` while the database is left in a partially written state: orphaned PR rows exist without a matching `miner_evaluations` row, causing inconsistent counts in subsequent rounds.

## Fix

Removed all internal `self.db.commit()` calls from the five `Repository` sub-write methods. Transaction control is now owned exclusively by `store_evaluation`, which issues a single `commit()` on full success or a single `rollback()` on any failure. This restores true atomicity — either all sub-writes for a round land together or none do.

## Impact

- A mid-evaluation failure now leaves the database exactly as it was before `store_evaluation` was called
- Subsequent rounds no longer see orphaned PR rows or inconsistent miner evaluation counts
- Behavior matches the documented contract: `StorageResult(success=False)` means no state was written

## Files Changed

- `gittensor/validator/utils/storage.py` — single `commit()` on success, single `rollback()` on failure
- `gittensor/validator/storage/repository.py` — removed premature `commit()` calls from `set_miner`, `store_pull_requests_bulk`, `store_issues_bulk`, `store_file_changes_bulk`, `set_miner_evaluation`

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -49,13 +49,14 @@ class BaseRepository:
         finally:
             cursor.close()
 
-    def execute_command(self, query: str, params: tuple = ()) -> bool:
+    def execute_command(self, query: str, params: tuple = (), commit: bool = True) -> bool:
         """
         Execute an INSERT, UPDATE, or DELETE command.
 
         Args:
             query: SQL command string
             params: Query parameters tuple
+            commit: Whether to commit after execution (default True)
 
         Returns:
             True if successful, False otherwise
@@ -63,25 +64,28 @@ class BaseRepository:
         try:
             with self.get_cursor() as cursor:
                 cursor.execute(query, params)
-                self.db.commit()
+                if commit:
+                    self.db.commit()
                 return True
         except Exception as e:
-            self.db.rollback()
+            if commit:
+                self.db.rollback()
             self.logger.error(f'Error executing command: {e}')
             return False
 
-    def set_entity(self, query: str, params: tuple) -> bool:
+    def set_entity(self, query: str, params: tuple, commit: bool = True) -> bool:
         """
         Insert or update an entity using the provided query.
 
         Args:
             query: SQL INSERT/UPDATE query with ON DUPLICATE KEY UPDATE
             params: Query parameters tuple
+            commit: Whether to commit after execution (default True)
 
         Returns:
             True if successful, False otherwise
         """
-        return self.execute_command(query, params)
+        return self.execute_command(query, params, commit=commit)
 
 
 class Repository(BaseRepository):
@@ -93,18 +97,19 @@ class Repository(BaseRepository):
     def __init__(self, db_connection):
         super().__init__(db_connection)
 
-    def set_miner(self, miner: Miner) -> bool:
+    def set_miner(self, miner: Miner, commit: bool = True) -> bool:
         """
         Insert a miner (ignore conflicts)
 
         Args:
             miner: Miner object to store
+            commit: Whether to commit after execution (default True)
 
         Returns:
             True if successful, False otherwise
         """
         params = (miner.uid, miner.hotkey, miner.github_id)
-        return self.set_entity(SET_MINER, params)
+        return self.set_entity(SET_MINER, params, commit=commit)
 
     def cleanup_stale_miner_data(self, evaluation: MinerEvaluation) -> None:
         """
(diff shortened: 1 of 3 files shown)


anderdc:
Fix conflicts.

Tet-9:
> Fix conflicts.

It's done now🫡

Tet-9:
@anderdc , it's review time🫡

anderdc (changes requested):
Premise and most of the fix are right, but `cleanup_stale_miner_data` was missed. `store_evaluation` (storage.py:84) calls it without `commit=False`, and that function's four `execute_command` calls (repository.py:132, 133, 138, 139) still default to `commit=True`. The first one commits the in-flight transaction, flushing the prior bulk PR/issue/file-change writes that were just queued — same orphans scenario as #749 describes, just shifted past cleanup.

Thread `commit: bool = True` through `cleanup_stale_miner_data` and its four `execute_command` calls, then call it as `cleanup_stale_miner_data(miner_eval, commit=False)` in `store_evaluation`. Add a test that forces a failure between cleanup and `set_miner_evaluation` so the regression is pinned.

Tet-9:
@anderdc, it's on for review now🫡

Tet-9:
Fixed — threaded commit=False through cleanup_stale_miner_data and its four execute_command calls, and added two regression tests that pin the failure scenario.

anderdc (approved):

PR #800, opened 2026-04-25
fix: handle null repository owner in parse_repo_name to prevent TypeError on deleted fork-owner accounts

### LINKED ISSUE #794: [Bug] parse_repo_name` crashes with TypeError when fork repository's owner account was deleted (null `owner`)
## Summary
`parse_repo_name` in `gittensor/utils/utils.py` dereferences `repo_data["owner"]["login"]` without a null check. When a PR's repository has `owner: null` — which GitHub's GraphQL API returns when the fork owner's account was deleted (ghost-user scenario) — this raises `TypeError: 'NoneType' object is not subscriptable`. The crash occurs at the top of `PullRequest.from_graphql_response`, before any per-PR try/except, so the entire PR is silently dropped from scoring.

## Location
https://github.com/entrius/gittensor/blob/test/gittensor/utils/utils.py#L9-L11

​```python
def parse_repo_name(repo_data: Dict):
    """Normalizes and converts repository name from dict"""
    return f'{repo_data["owner"]["login"]}/{repo_data["name"]}'.lower()
​```

## Impact
Honest-miner harm: any miner whose PR history includes a fork whose owner later deleted their GitHub account has those PRs (and potentially the entire scoring pass for that miner, depending on call site) silently dropped.

**Six call sites** (all unguarded):

| File | Line | Risk |
|---|---|---|
| `gittensor/classes.py` | 239 | **Primary crash path** — top of `from_graphql_response`, drops PR from scoring |
| `gittensor/classes.py` | 422, 430, 436 | Log-formatting — shielded in practice by earlier :239 failure |
| `gittensor/utils/github_api_tools.py` | 887 | `head_repo and parse_repo_name(head_repo)` — the `head_repo` truthy check only guards `head_repo = None`, NOT `head_repo['owner'] = None`, so still crashes |
| `gittensor/utils/github_api_tools.py` | 983 | Inside a `try:` — caught but noisy |

## Reproduction
A PR whose `headRepository.owner` (or `repository.owner`) is `null` in GitHub's GraphQL response. This happens when the fork owner's GitHub account has been deleted; GitHub keeps the PR visible but returns `null` for the repository's `owner` field.

## Why this isn't covered by #537
[#537](https://github.com/entrius/gittensor/issues/537)'s null-author fix guarded `pr_raw['author']` only. The nested `repository.owner` / `headRepository.owner` paths were not touched. Same ghost-user pattern, different field.

## Suggested fix (~3 lines)

​```python
from typing import Optional

def parse_repo_name(repo_data: Dict) -> Optional[str]:
    """Normalizes and converts repository name from dict.
    Returns None if owner is missing (e.g. deleted fork-owner account)."""
    owner = (repo_data.get('owner') or {}).get('login')
    name = repo_data.get('name')
    return f'{owner}/{name}'.lower() if owner and name else None
​```

Call sites need a `None`-guard on the result — skip the PR / log and continue rather than propagating `None` into downstream string comparisons. Happy to open a PR if useful.

Tet-9:
Fixes #794

## Problem
`parse_repo_name` in `gittensor/utils/utils.py` dereferences
`repo_data["owner"]["login"]` without a null check. When a PR's
repository has `owner: null` — which GitHub's GraphQL API returns when
the fork owner's account was deleted — this raises
`TypeError: 'NoneType' object is not subscriptable`.

The crash occurs at the top of `PullRequest.from_graphql_response`
before any per-PR try/except, so the entire PR is silently dropped
from scoring. Honest miners whose PR history includes a fork whose
owner later deleted their GitHub account have those PRs dropped.

### `gittensor/utils/utils.py`
Changed `parse_repo_name` to return `Optional[str]` — returns `None`
when `owner` or `name` is missing, instead of crashing.

### `gittensor/classes.py`
- Line 239 (`from_graphql_response`): raise `ValueError` on `None`
  result so the caller's try/except logs and skips the PR cleanly
- Lines 422, 430, 436 (log call sites): guard with `or '<unknown>'`
  so logging never crashes even if called before the ValueError path

### `gittensor/utils/github_api_tools.py`
- Line 869: guard `parse_repo_name(head_repo)` with `or ''` so the
  equality check against `repository_full_name` never crashes
- Line 965: add explicit `None` check after `parse_repo_name` call,
  log a warning and `continue` to skip the PR cleanly

## Changes
- `gittensor/utils/utils.py` — `parse_repo_name` returns `Optional[str]`
- `gittensor/classes.py` — null guards at all 4 call sites
- `gittensor/utils/github_api_tools.py` — null guards at both call sites

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -237,6 +237,8 @@ class PullRequest:
         from gittensor.validator.utils.datetime_utils import parse_github_timestamp_to_cst
 
         repository_full_name = parse_repo_name(pr_data['repository'])
+        if repository_full_name is None:
+            raise ValueError(f'PR #{pr_data.get("number")} has null repository owner — skipping')
         pr_state = PRState(pr_data['state'])
         is_merged = pr_state == PRState.MERGED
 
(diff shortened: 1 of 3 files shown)


Tet-9:
@anderdc , review!!

PR #804, opened 2026-04-26
feat(cli): add `gitt config unset <key>` subcommand

ebios-star:
## Summary

Completes the standard CLI config triplet. \`gitt config\` already shows all values and \`gitt config set <k> <v>\` writes one, but there was no way to remove a stale key short of editing the JSON by hand. Add \`gitt config unset <key>\` to fill the gap, mirroring the convention from \`git config --unset\`, \`kubectl config unset\`, \`gh config clear-cache\`, etc.

\`\`\`
$ gitt config unset wallet
Unset wallet: alice

$ gitt config unset not_set; echo \$?
Error: Key 'not_set' is not set in config
1
\`\`\`

Implementation notes:
- Removes KEY from \`~/.gittensor/config.json\` and persists the trimmed config; prints the old value so the operator can see what was cleared.
- Exits non-zero with a clear error when the config file is missing, the JSON is invalid, or KEY is not set — so shell scripts can detect no-op invocations.
- Leaves the file untouched on the invalid-JSON path, instead of silently clobbering unparsable content (verified by a test).
- Lives next to \`config_set\` in [gittensor/cli/main.py](gittensor/cli/main.py) and slots into the existing \`config_group\`, so \`gitt config --help\` lists \`unset\` automatically.

## Related Issues

N/A.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- New \`tests/cli/test_config_unset.py\` covers the four exit paths (key removed, key missing, file missing, invalid JSON).
- \`uv run --extra dev pytest tests/\` — all 447 tests pass (443 baseline + 4 new).

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/cli/main.py
+++ b/gittensor/cli/main.py
@@ -138,6 +138,39 @@ def config_set(key: str, value: str):
         console.print(f'[green]Set {key}:[/green] {value}')
 
 
+@config_group.command('unset')
+@click.argument('key', type=str)
+def config_unset(key: str):
+    """Remove a configuration value.
+
+    [dim]Deletes KEY from `~/.gittensor/config.json`. Exits non-zero if the
+    config file is missing, the JSON is invalid, or KEY is not set, so
+    shell scripts can detect the missing key.[/dim]
+
+    [dim]Examples:
+        $ gitt config unset wallet
+        $ gitt config unset contract_address
+    [/dim]
+    """
+    if not CONFIG_FILE.exists():
+        console.print('[red]Error: No config file found at ~/.gittensor/config.json[/red]')
+        raise SystemExit(1)
+
+    try:
+        config = json.loads(CONFIG_FILE.read_text())
+    except json.JSONDecodeError:
+        console.print('[red]Error: Invalid JSON in config file[/red]')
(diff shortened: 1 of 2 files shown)


ebios-star:
@anderdc @landyndev — friendly ping for review when you have a moment.

Adds `gitt config unset <key>` to complete the standard config CLI triplet (alongside `gitt config` and `gitt config set`). Mirrors `git config --unset` / `kubectl config unset`. New tests cover all four exit paths (key removed, key missing, file missing, invalid JSON) including the on-disk-preservation invariant for the corrupt-file path.

Complementary to (but independent of) #813's `gitt config set` whitelist.

anderdc:
`gitt config set <key> <new_value>` already covers the realistic case (changing a stale value). The only thing `unset` adds is removing the key entirely from the JSON, which is cosmetic — downstream code reads via `config.get(key)` so empty-string and missing-key behave the same. The git/kubectl/gh precedent doesn't transfer cleanly because those tools have structurally-distinct empty-vs-missing semantics this config doesn't. Closing.

PR #805, opened 2026-04-26
fix: reject unrecognised --network values in gitt miner commands

### LINKED ISSUE #795: Incomplete fix of #694: `gitt miner` commands still forward typo'd `--network` strings as endpoints
## Summary
Issue #694 ("Typos in `--network` were treated as custom RPC/WebSocket URLs") was closed, but the fix is only in effect for `issue_commands` — and only because those commands declare `type=NETWORK_CHOICE` on their Click option, so invalid names are rejected at parse time. The underlying `resolve_network()` function and the miner-side `_resolve_endpoint()` still pass unknown strings straight through.

PR #695 (closed-unmerged) would have closed the gap on both paths; without it, `gitt miner post` / `gitt miner check` happily accept a typo like `--network finny` and feed it to `bt.Subtensor(network='finny')`.

## Location
- [`gittensor/cli/miner_commands/helpers.py:45-57`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/miner_commands/helpers.py#L45-L57) — `_resolve_endpoint` uses `NETWORK_MAP.get(network.lower(), network)`, falling back to the raw string.
- [`gittensor/cli/miner_commands/post.py:38`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/miner_commands/post.py#L38) and [`check.py:32`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/miner_commands/check.py#L32) — `--network` is a plain string option, **no `type=NETWORK_CHOICE`**, so Click does not reject typos.

For contrast, [`issue_commands/helpers.py:58`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/issue_commands/helpers.py#L58) defines `NETWORK_CHOICE = click.Choice(['finney','test','local'])` and applies it to all issue-command `--network` options — which is what saves that path.

## Impact
- `gitt miner post --network finny` fails with an obscure `bt.Subtensor` connection error instead of a clean "unknown network" message.
- Same for config-file typos (`gitt config set network finny` then `gitt miner post`), since the config path bypasses Click entirely.
- The PAT is validated locally before network setup and `_connect_bittensor` fails fast, so **the PAT is not actually sent anywhere**. The harm is UX/debuggability, not credential leakage.

## Reproduction
```bash
gitt config set network finny        # typo
gitt miner post --pat ghp_xxxx       # or: gitt miner post --network finny --pat ...
# → "Failed to initialize bittensor: ..." with confusing cause

Tet-9:
Fixes #795

## Problem
`gitt miner post` and `gitt miner check` accepted any string for
`--network` without validation. A typo like `--network finny` was
silently forwarded to `bt.Subtensor(network='finny')`, which produced
an obscure connection error instead of a clear "unknown network"
message.

The root cause was two-fold:
1. `--network` in `post.py` and `check.py` used a plain `str` Click
   option with no `type=` constraint, so Click accepted any value
2. `_resolve_endpoint` fell back to the raw string via
   `NETWORK_MAP.get(network.lower(), network)` when the name was
   unrecognised

By contrast, `issue_commands` already defined `NETWORK_CHOICE =
click.Choice(...)` and applied it to all `--network` options — which
is why that path was safe.

The same gap existed for config-file typos: `gitt config set network
finny` followed by `gitt miner post` bypassed Click entirely and hit
the same fallback in `_resolve_endpoint`.

### `gittensor/cli/miner_commands/helpers.py`
- Added `NETWORK_CHOICE = click.Choice(['finney', 'test', 'local'],
  case_sensitive=False)` mirroring `issue_commands/helpers.py`
- Rewrote `_resolve_endpoint` to raise `click.BadParameter` when the
  network name is not in `NETWORK_MAP` and is not a WebSocket URL
  (`ws://` or `wss://`). Raw WebSocket URLs are still accepted for
  custom devnet setups

### `gittensor/cli/miner_commands/post.py` and `check.py`
- Added `NETWORK_CHOICE` to imports
- Applied `type=NETWORK_CHOICE` to the `--network` Click option so
  Click rejects typos at parse time before any code runs

## Changes
- `gittensor/cli/miner_commands/helpers.py` — added `NETWORK_CHOICE`,
  rewrote `_resolve_endpoint` with strict validation
- `gittensor/cli/miner_commands/post.py` — `type=NETWORK_CHOICE` on
  `--network`
- `gittensor/cli/miner_commands/check.py` — `type=NETWORK_CHOICE` on
  `--network`

### DIFF
--- a/gittensor/cli/miner_commands/check.py
+++ b/gittensor/cli/miner_commands/check.py
@@ -12,6 +12,7 @@ from rich.table import Table
 
 from .helpers import (
     NETUID_DEFAULT,
+    NETWORK_CHOICE,
     _connect_bittensor,
     _error,
     _load_config_value,
@@ -29,7 +30,7 @@ console = Console()
 @click.option('--wallet', 'wallet_name', default=None, help='Bittensor wallet name.')
 @click.option('--hotkey', 'wallet_hotkey', default=None, help='Bittensor hotkey name.')
 @click.option('--netuid', type=int, default=NETUID_DEFAULT, help='Subnet UID.', show_default=True)
-@click.option('--network', default=None, help='Network name (local, test, finney).')
+@click.option('--network', default=None, type=NETWORK_CHOICE, help='Network name (local, test, finney).')
 @click.option('--rpc-url', default=None, help='Subtensor RPC endpoint URL (overrides --network).')
 @click.option('--json-output', 'json_mode', is_flag=True, default=False, help='Output results as JSON.')
 def miner_check(wallet_name, wallet_hotkey, netuid, network, rpc_url, json_mode):
--- a/gittensor/cli/miner_commands/helpers.py
+++ b/gittensor/cli/miner_commands/helpers.py
@@ -14,6 +14,8 @@ from rich.console import Console
 
 from gittensor.constants import NETWORK_MAP
 
+NETWORK_CHOICE = click.Choice(['finney', 'test', 'local'], case_sensitive=False)
+
 console = Console()
 
 NETUID_DEFAULT = 74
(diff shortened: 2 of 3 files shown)


anderdc:
See #795 — issue closed as not worth doing. Closing.

PR #724, opened 2026-04-22
fix: json mode and human mode disagree on exit codes and aggregates

andriypolanski:
## Summary

1. **`gitt issues list --id`** — Missing on-chain issue: JSON already exited non-zero with a structured error; human mode exited **0** after a soft warning. Both paths now use the same helper as other issue commands (`handle_exception`) so exit code and error shape stay consistent.

2. **`gitt miner check --json-output`** — The top-level **`invalid`** count mixed “no PAT” (`has_pat is False`) with “invalid PAT” rows. JSON now exposes **`no_pat`**, **`invalid_pat`**, and **`no_response`** (plus **`valid`**), matching the human table semantics. **`invalid` is removed** from the JSON payload (breaking change for scripts that read that field).

## Related Issues
Closes https://github.com/entrius/gittensor/issues/723

## Type of Changes

- [x]  Bug Fix

## Tests

- `tests/cli/test_issues_list_json.py` — **`test_issues_list_human_missing_issue_exits_non_zero`**
- `tests/cli/test_miner_commands.py` — **`TestPatCheckAggregateCounts`** (imports `_pat_check_aggregate_counts` from miner `helpers`)

## Migration notes (miner check JSON)

- Replace reads of **`invalid`** with **`no_pat`** + **`invalid_pat`**, or derive from **`results`**.
- **`no_response`** still means rows classified as no response (including cases that do not match the other three categories).

## Test plan

```bash
pytest tests/cli/test_issues_list_json.py -q
pytest tests/cli/test_miner_commands.py::TestPatCheckAggregateCounts -q
```

Issue #542, opened 2026-04-17
miner check/post exit 0 when zero validators accept PAT


aviu16:
## Description

Both `miner post` and `miner check` exit with code 0 after the broadcast/check phase even when `accepted_count == 0` (post) or `valid_count == 0` (check).

Exit code 1 is only set for pre-flight errors (invalid PAT, hotkey not registered, no reachable validators). But the operational failure case — all validators reachable, none accepted — exits 0.

`miner_post` at `cli/miner_commands/post.py:168` falls through to exit 0 after printing results. The JSON output does include `"success": false` when `accepted_count == 0`, but the exit code contradicts it.

`miner_check` at `cli/miner_commands/check.py:114` also exits 0 regardless of results. Its JSON output doesn't even include a `success` field (inconsistent with `miner_post`).

## Steps to Reproduce

1. Register a hotkey on subnet 74
2. Run `gitt miner post` with a valid PAT when all validators happen to reject (e.g. account age < 180 days)
3. Observe exit code: `echo $?` → `0`
4. Same for `gitt miner check` when no validators have the PAT stored

## Impact

CI/CD pipelines checking exit codes treat failed broadcasts as success. Automation that gates on `gitt miner post && deploy` proceeds even when no validator accepted the PAT.

anderdc:
I wouldn't call that an operational failure, it's simply the output, no one accepted the PAT.

PR #807, opened 2026-04-26
fix(cli): miner post JSON drops no_response into rejected bucket

### LINKED ISSUE #806: gitt miner post: JSON output collapses no_response into rejected
## Summary

`gitt miner post` (`gittensor/cli/miner_commands/post.py`) renders three response categories in human mode but only two in JSON mode, silently collapsing `no_response` into `rejected`. This is the same JSON/human-parity bug class that #724 just fixed in the sibling `gitt miner check` command — the same shared-helper pattern applies directly.

## The mismatch

The synapse exposes three distinct response states: `accepted=True`, `accepted=False`, and `accepted=None` (validator unreachable / no response).

**Human mode** (`post.py:158-164`) renders them as three separate categories:

```python
if r['accepted'] is True:    status = '[green]✓[/green]'      # accepted
elif r['accepted'] is False: status = '[red]✗[/red]'         # rejected
else:                        status = '[yellow]—[/yellow]'   # no response
```

**JSON mode** (`post.py:138-150`) emits only two buckets:

```python
{
    'success': accepted_count > 0,
    'total_validators': len(results),
    'accepted': accepted_count,
    'rejected': len(results) - accepted_count,   # silently includes no_response
    'results': results,
}
```

So a run with 2 accepts, 1 reject, and 2 no-responses shows `2 ✓ / 1 ✗ / 2 —` to humans but reports `accepted: 2, rejected: 3` in JSON. JSON consumers cannot distinguish "validator rejected my PAT" from "validator unreachable" — two operationally different conditions that need different follow-up.

## Suggested fix (mirror of #724)

#724 solved the identical bug class in `gitt miner check` by extracting two shared helpers in `gittensor/cli/miner_commands/helpers.py`:

- `_pat_check_row_category(row)` — single source of truth for category classification
- `_pat_check_aggregate_counts(results)` — Counter-based bucketing both modes consume

The same shape applies here:

1. Add `_pat_post_row_category(row)` with three buckets: `accepted`, `rejected`, `no_response`.
2. Add `_pat_post_aggregate_counts(results)` returning `{'accepted': N, 'rejected': N, 'no_response': N}`.
3. Update `post.py` so both renderings consume the helpers (`**counts` spread in JSON; `_PAT_POST_STATUS_MARKUP[category]` lookup in the table).
4. Add unit tests in `tests/cli/test_miner_commands.py` mirroring `TestPatCheckAggregateCounts`.

After the fix the JSON output shape becomes:

```python
{
    'success': counts['accepted'] > 0,
    'total_validators': len(results),
    **counts,                              # accepted / rejected / no_response
    'results': results,
}
```

statxc:
## Summary

Mirror of #724 for the sibling `gitt miner post` command. Three response states (`accepted=True` / `False` / `None`) are rendered as three distinct buckets in human mode, but JSON mode emits only `accepted` and `rejected`, silently bucketing `no_response` into `rejected` via `rejected = total - accepted`.

Fixes #806 

## Problem

`gittensor/cli/miner_commands/post.py` had two divergent classifications of the same per-validator response:

- **Human table** (3 buckets): `[green]✓[/green]` / `[red]✗[/red]` / `[yellow]—[/yellow]`
- **JSON output** (2 buckets): `accepted` and `rejected = total - accepted` — silently includes no-responses

Example with 5 validators (2 accept, 1 reject, 2 unreachable):

```
human: 2 ✓ / 1 ✗ / 2 —
JSON:  {"accepted": 2, "rejected": 3}     ← no_response folded into rejected
```

JSON consumers cannot distinguish "validator rejected my PAT" from "validator unreachable" — operationally distinct conditions that need different follow-up.

## Fix

Mirror the structure #724 used for `gitt miner check`:

- Add `_pat_post_row_category(row) -> {'accepted'|'rejected'|'no_response'}` and `_pat_post_aggregate_counts(results) -> {accepted, rejected, no_response}` to `gittensor/cli/miner_commands/helpers.py`, sibling to the existing `_pat_check_*` pair.
- Replace the inline if/elif/else in `post.py`'s human-mode loop with a `_PAT_POST_STATUS_MARKUP[category]` lookup driven by the same row-category helper.
- Spread `**counts` into the JSON envelope so all three buckets surface in the output.

After the fix:

```
JSON: {"accepted": 2, "rejected": 1, "no_response": 2}
```

## Out of scope

This PR fixes only the JSON/human classification mismatch. Exit-code behavior on zero acceptances is intentionally not changed — the maintainer's position on that case is settled by #542 ("I wouldn't call that an operational failure, it's simply the output, no one accepted the PAT") and by the prior PR attempts #559, #550, #679.


statxc:
Hi @anderdc how are you? could you review this?

anderdc:
fix conflicts we just added some changes to those files

statxc:
> fix conflicts we just added some changes to those files

@anderdc I've fixed conflicts. Thanks.

anderdc (changes requested):
post.py:163-164 still has explicit 'accepted' and 'rejected' keys after the **counts spread, so rejected = len(results) - accepted_count overrides the correct value from counts and still bundles no_response. Drop those two keys (same shape as #724's change to check.py:107-114). Add an envelope-level test asserting accepted + rejected + no_response == total_validators.

statxc:
> post.py:163-164 still has explicit 'accepted' and 'rejected' keys after the **counts spread, so rejected = len(results) - accepted_count overrides the correct value from counts and still bundles no_response. Drop those two keys (same shape as #724's change to check.py:107-114). Add an envelope-level test asserting accepted + rejected + no_response == total_validators.

Thanks @anderdc  I’ve updated this following you.

anderdc (approved):

statxc:
Thanks @anderdc

PR #809, opened 2026-04-27
fix: reject PATs whose GraphQL viewer is null and guard empty errors list

### LINKED ISSUE #751: [Bug] _test_pat_against_repo accepts viewer:null and crashes on empty errors list
## Summary

`_test_pat_against_repo` ([`gittensor/validator/pat_handler.py:161-182`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/pat_handler.py#L161)) accepts a PAT as valid whenever the GraphQL response is HTTP 200 and lacks an `errors` field. It does not verify that the `viewer` query actually returned data.

A PAT that authenticates but lacks the `read:user` scope (fine-grained PATs without `Public Repositories (read-only)`) responds with `{"data": {"viewer": null}}` - HTTP 200, no `errors` key - and is accepted. The validator stores it as valid. On the next scoring round, `get_github_id` ([`github_api_tools.py:221-238`](https://github.com/entrius/gittensor/blob/test/gittensor/utils/github_api_tools.py#L221)) sees the same `viewer: null` shape and returns `None`. The miner is then dropped with `failed_reason="No Github id found for miner X's PAT"`. The miner sees `pat_valid=True` from `gitt miner check` for the same PAT.

A second smaller defect in the same function: `data["errors"][0].get("message", "unknown")` at [line 179](https://github.com/entrius/gittensor/blob/test/gittensor/validator/pat_handler.py#L179) raises `IndexError` when `errors` is an empty list. Some upstream proxies return `{"errors": []}`. The `IndexError` is not caught and propagates out of the synapse handler.

## Steps to reproduce

For the viewer-null path:

1. Generate a fine-grained PAT with no permissions selected (or scoped to a private repo with no `read:user`).
2. Run `gitt miner post --pat <pat>` against a validator.
3. The validator's `handle_pat_broadcast` accepts the PAT.
4. Wait one full scoring round.
5. The validator drops the miner with `"No Github id found for miner X's PAT"`. The miner's local `gitt miner check` reports `pat_valid=True`.

For the empty-errors-list path:

1. Mock `requests.post` to return `{"data": null, "errors": []}`.
2. `_test_pat_against_repo` reaches `data["errors"][0]` and raises `IndexError`.

## Expected

A PAT whose viewer query returns `null` is rejected at PAT-broadcast time with a message pointing at the missing scope (similar to the message the miner-side CLI prints in `cli/miner_commands/post.py:189-191`). An empty `errors` list is handled without an `IndexError`.

## Actual

The PAT is accepted; the failure surfaces a round later as an evaluation drop, indistinguishable from a miner who never broadcast a PAT. The empty-errors edge case crashes the handler.

MilosM348:
```
## Summary

Closes #751.

`_test_pat_against_repo` (gittensor/validator/pat_handler.py) accepts
a PAT as valid whenever the GraphQL response is HTTP 200 and lacks an
`errors` field. It does not verify that the `viewer` query actually
returned data. Two distinct defects fall out of that:

1. A fine-grained PAT that authenticates but lacks the
   `Public Repositories (read-only)` scope responds with
   `{"data": {"viewer": null}}` — HTTP 200, no `errors` key. The
   current code passes through to `return None` (success). The
   validator stores the PAT as valid; the next scoring round drops
   the miner with `"No Github id found for miner X's PAT"`. The miner
   sees `pat_valid=True` from `gitt miner check` for the same PAT and
   has no actionable signal to fix the scope before scoring bites.

2. `data["errors"][0].get("message", "unknown")` raises `IndexError`
   when `errors` is `[]`. Some upstream proxies return `{"errors": []}`
   (a non-canonical but observed shape). The `IndexError` is not
   caught and propagates out of the synapse handler.

## Fix

Tighten the success condition to require all of:

  1. HTTP 200 (existing).
  2. `errors` is empty / absent. The check now uses a truthy test
     (`if errors:`) so an empty list and `None` are both treated as
     "no errors" — fixes the `IndexError` AND matches the spec
     semantics. The first-error extraction also guards against
     non-dict entries.
  3. `data.viewer` is non-null. A new branch returns a descriptive
     error message that points the miner at the missing scope, so
     `gitt miner post` and `gitt miner check` surface a clear reason
     at broadcast time instead of leaking it as an opaque evaluation
     drop a round later.

The error message wording mirrors the one already printed by
`cli/miner_commands/post.py:189-191` so the validator-side rejection
is consistent with what miners see locally.

## Tests

Added `TestTestPatAgainstRepo` to
`tests/validator/test_pat_handler.py` with six unit tests directly
invoking `_test_pat_against_repo` against a `requests.post` mock:

  - viewer with login → accepted
  - viewer == null → rejected, error message mentions "viewer"
  - missing `data` key → rejected
  - empty errors list + valid viewer → accepted, no IndexError
  - non-empty errors → first error message surfaced
  - non-200 status → status code surfaced

The existing eight tests that mock `_test_pat_against_repo` itself
continue to pass because the function's signature and contract are
unchanged (still `Optional[str]`, error-string-or-None semantics
preserved).

## Reproduction

For the viewer-null defect (from the issue body):

1. Generate a fine-grained PAT with no permissions selected (or
   scoped to a private repo with no `read:user`).
2. `gitt miner post --pat <pat>` against a validator.
3. Before: `handle_pat_broadcast` accepts; one round later the miner
   is dropped with `"No Github id found for miner X's PAT"` while
   `gitt miner check` still reports `pat_valid=True`.
4. After: broadcast is rejected immediately with the new descriptive
   error, no broken state stored.

For the empty-errors-list defect:

1. Mock `requests.post` to return `{"data": null, "errors": []}`.
2. Before: `_test_pat_against_repo` reaches `data["errors"][0]` and
   raises `IndexError`, which propagates out of the handler.
3. After: empty list is treated as "no errors", function falls
   through to the viewer check and returns the viewer-null error.

## Closes

Closes #751.
```

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -163,6 +163,16 @@ def _test_pat_against_repo(pat: str) -> Optional[str]:
 
     Scoring uses the GraphQL API to fetch miner PRs, so this mirrors the real path.
     Returns an error string on failure, None on success.
+
+    The PAT is considered valid only when:
+      1. The HTTP response is 200.
+      2. The body has no non-empty `errors` list.
+      3. The `data.viewer` payload is non-null. A fine-grained PAT that
+         authenticates but lacks the `Public Repositories (read-only)` scope
+         responds with `{"data": {"viewer": null}}` — HTTP 200, no `errors`
+         key — which would otherwise be silently accepted and only surface
+         a round later when scoring drops the miner with
+         `"No Github id found for miner X's PAT"`.
     """
     headers = {'Authorization': f'Bearer {pat}', 'Accept': 'application/json'}
     try:
@@ -175,8 +185,20 @@ def _test_pat_against_repo(pat: str) -> Optional[str]:
         if response.status_code != 200:
             return f'GitHub GraphQL API returned {response.status_code}'
         data = response.json()
-        if 'errors' in data:
-            return f'GraphQL error: {data["errors"][0].get("message", "unknown")}'
+        # Truthy check — covers both `[]` (empty list returned by some
+        # upstream proxies, which is "no errors") and `None`. Avoids the
+        # IndexError that `data["errors"][0]` would raise on `[]`.
+        errors = data.get('errors')
+        if errors:
+            first = errors[0] if isinstance(errors[0], dict) else {}
+            return f'GraphQL error: {first.get("message", "unknown")}'
+        viewer = (data.get('data') or {}).get('viewer')
+        if viewer is None:
+            return (
+                'PAT lacks GraphQL viewer access. Fine-grained PATs need '
+                '"Public Repositories (read-only)" permission; classic PATs '
+                'need the read:user scope.'
+            )
         return None
     except requests.RequestException as e:
         return str(e)
--- a/tests/validator/test_pat_handler.py
+++ b/tests/validator/test_pat_handler.py
@@ -11,6 +11,7 @@ from bittensor.core.synapse import TerminalInfo
 from gittensor.synapses import PatBroadcastSynapse, PatCheckSynapse
 from gittensor.validator import pat_storage
 from gittensor.validator.pat_handler import (
+    _test_pat_against_repo,
     blacklist_pat_broadcast,
     blacklist_pat_check,
     handle_pat_broadcast,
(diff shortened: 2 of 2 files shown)


anderdc:
Duplicate of #764 which is better scoped/implemented. Closing.

PR #811, opened 2026-04-27
refactor: extract duplicate backoff delay calc into _backoff_seconds helper

### LINKED ISSUE #810: Duplicated exponential-backoff formula across 9 call sites in github_api_tools.py
## Description

`gittensor/utils/github_api_tools.py` open-codes the same exponential-backoff sleep formula `min(5 * (2 ** attempt), 30)` (plus two near-identical variants with different base/cap values) at 9 separate call sites across
`get_merge_base_sha`, `get_pull_request_file_changes`, `execute_graphql_query`, and `get_github_graphql_query`.

This is a copy-paste smell:
- Tuning the backoff curve (base delay, cap, jitter) requires touching every call site and trusting the reader to spot drift.
- The two variants (e.g. resource-limit retries using `base=2, cap=15`) are visually almost identical to the main formula, making it easy to "fix" one and silently leave the other behind.
- There is no single place to add observability (e.g. log the chosen sleep duration) without 9 edits.

## Steps to Reproduce

1. `grep -n "min(.*\*\*attempt" gittensor/utils/github_api_tools.py` — observe 9 hits with subtle differences in the constants.
2. Note that retry policy is implicit at each call site rather than expressed as a named helper.

## Expected Behavior

A single private helper, e.g.

```python
def _backoff_seconds(attempt: int, base: int = 5, cap: int = 30) -> int:
    return min(base * (2 ** attempt), cap)
```

## Actual Behavior

The formula is inlined at every retry site, with two parallel variants
(base=5, cap=30 and base=2, cap=15) that must be kept in sync by hand.

wdeveloper16:
## Summary

- The expression `min(5 * (2**attempt), 30)` (and two variants) was  copy-pasted across 9 call sites in `github_api_tools.py`
- Extracts a single private `_backoff_seconds(attempt, base=5, cap=30)`  helper that encodes the formula once
- All callers (`get_merge_base_sha`, `get_pull_request_file_changes`,  `execute_graphql_query`, `get_github_graphql_query`) updated —  no behaviour change

## Test plan
- [ ] Confirm existing tests pass
- [ ] Verify backoff values unchanged: attempt 0→5s, 1→10s, 2→20s, 3→30s (capped)
- [ ] Confirm resource-limit variant (base=2, cap=15) still correct

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -9,7 +9,7 @@ from datetime import datetime, timedelta, timezone
 from math import ceil
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
 
-from gittensor.utils.utils import parse_repo_name
+from gittensor.utils.utils import backoff_seconds, parse_repo_name
 
 if TYPE_CHECKING:
     from gittensor.classes import FileChange as FileChangeType
@@ -299,7 +299,7 @@ def get_merge_base_sha(repository: str, base_sha: str, head_sha: str, token: str
                 return None
 
             if attempt < max_attempts - 1:
-                backoff_delay = min(5 * (2 ** (attempt)), 30)
+                backoff_delay = backoff_seconds(attempt)
                 bt.logging.warning(
                     f'Compare API for {repository} failed with status {response.status_code} '
                     f'(attempt {attempt + 1}/{max_attempts}), retrying in {backoff_delay}s...'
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
Same formula lives in `gittensor/utils/mirror/client.py:106` and `:123` — both `min(5 * (2**attempt), 30)`, identical to most call sites here. Pull those in too. Move the helper to `gittensor/utils/utils.py` (next to `parse_repo_name`) and drop the leading underscore so both files import it from there — avoids a mirror→github_api_tools dependency and keeps it where shared utilities already live.

The linear `2 * (attempt + 1)` at github_api_tools.py:567 and the `time.sleep(2**attempt)` in neurons/base/neuron.py:147 are different shapes — leave alone.

wdeveloper16:
Thanks for your review, Fair point.
Will fix it right away.

wdeveloper16:
Just done!
I'd be happy if you'd review the PR again.
Thanks

anderdc (approved):

PR #820, opened 2026-04-27
fix: alert on mirror service outage instead of silently zeroing scores

wdeveloper16:
## Summary

Follow-up to #796. When `mirror.gittensor.io` is unavailable, fetch failures were either logged at `WARNING` level (issue discovery) or without scoring-impact context (OSS path), making it impossible to distinguish a service outage from isolated miner failures by watching error logs.

- Upgrade per-miner issue-discovery fetch failure from `WARNING` → `ERROR` (consistent with OSS path severity)
- Improve OSS per-miner error message to include hotkey prefix and explicit consequence ("scores will be 0 this round for this miner")
- Add aggregate outage detection in `run_mirror_issue_discovery`: emits `ERROR` when ≥50% of eligible miners fail **and** `total_eligible >= 5` (the guard suppresses false positives in small/dev environments)
- Add 4 tests covering: total failure fires alert, 50% boundary fires, 20% rate does not fire, small-population guard suppresses

## No behaviour changes

Scoring logic is untouched — this is purely observability. All 51 existing tests continue to pass.

## Test plan

- [ ] `uv run --with pytest pytest tests/validator/oss_contributions/mirror/test_load.py tests/validator/issue_discovery/test_mirror_scan.py -v` → 51 passed

### DIFF
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -150,7 +150,10 @@ async def run_mirror_issue_discovery(
         try:
             response = client.get_miner_issues(evaluation.github_id, since=lookback_date)
         except MirrorRequestError as e:
-            bt.logging.warning(f'├─ UID {uid}: mirror issue fetch failed ({e}) — skipped this miner')
+            bt.logging.error(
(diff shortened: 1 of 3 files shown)


anderdc:
Needless logging change. Closing

PR #822, opened 2026-04-27
fix: enforce one-issue-per-PR across miners in mirror issue discovery

### LINKED ISSUE #821: [Bug] Mirror issue discovery one-issue-per-PR rule is enforced per-miner rather than per-round
## Description

PR #796 ported the issue-discovery one-issue-per-PR rule into the new mirror path but moved the deduplication set from round-scope into per-miner scope. A solving PR that closes issues authored by multiple miners now pays *each* miner a full discovery score instead of the documented "earliest-created qualifying issue wins" behavior — silently inflating emissions on `closes #N closes #M` PRs and enabling cheap two-account collusion.

## Steps to Reproduce

1. Two registered miners A and B, both eligible (≥ 7 valid solved issues).
2. A files issue `#50` and B files issue `#51` in the same mirror-enabled repo.
3. A third party merges a PR with body `closes #50 closes #51`.
4. Validator runs a scan round.

Both A and B receive the full discovery score for that single PR. Only one should.

The pytest version of step 4 (committed on the companion fix branch) fails on `origin/test @ 41673c6603`:

```
assert e_a.issue_discovery_score > e_b.issue_discovery_score > 0
E   AssertionError: assert 88.97 > 88.97
```

## Expected Behavior

One discovery score per solving PR per round. The earliest-created qualifying issue wins (tie-break: `issue_number`, then `uid`); other miners' issues on the same PR are credibility-only. Both miners still gain credibility from the closed issue regardless of who scores.

## Actual Behavior

The dedup set lives inside the per-miner scorer ([`mirror_scan.py:243` on `origin/test`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/issue_discovery/mirror_scan.py#L243)):

```python
def _score_miner_mirror_issues(...):
    ...
    pr_scored_keys: Set[Tuple[str, int]] = set()  # ← rebuilt on every call
```

`_score_miner_mirror_issues` is called once per miner, so the set never sees more than one miner's issues. Within a single miner the rule still works; only the cross-miner case leaks.

For comparison, the pre-#796 legacy path placed the same set at function scope of `_collect_issues_from_prs` ([`scoring.py:193-194` at `41673c6603^`](https://github.com/entrius/gittensor/blob/41673c6603%5E/gittensor/validator/issue_discovery/scoring.py#L193-L194)), which iterated all miners in a single pass — round-global by construction.

## Environment

- OS: Ubuntu 24.04 (x86_64)
- Python: 3.12.3
- Commit: `origin/test @ 41673c6603` (merge commit of PR #796)

## Additional Context

**Why this matters:**

- `closes #N closes #M` is a common GitHub pattern when maintainers fold duplicate reports into one PR. Each duplicate authored by a different miner gets paid in full.
- Two-account collusion: register a second miner, mirror the first's issues, double the discovery pay on any third-party fix that closes both. Cost: one extra hotkey.
- Logs don't flag the overlap — each per-miner log line looks normal.

A fix branch (`fix/mirror-cross-miner-one-issue-per-pr`) with regression tests is ready and will be opened with `Closes #<this issue>` once this is filed.

mkdev5:
## Summary

PR #796 inlined the one-issue-per-PR dedup set inside `_score_miner_mirror_issues` (called once per miner), shrinking the rule's lifetime from "round" to "single miner." A solving PR closing issues authored by multiple miners then paid each one — silently inflating discovery emissions on `closes #N closes #M` PRs and enabling two-account collusion.

This PR restores legacy round-global semantics by:

1. **Deferred scoring**: `run_mirror_issue_discovery` collects every miner's `(evaluation, filtered_issues, open_issue_count)` before scoring, so the one-issue-per-PR decision sees all miners at once.
2. **Cross-miner canonical map**: A new `_build_canonical_pr_owners` helper produces `(repo, pr_number) → (created_at, issue_number, uid)` for the earliest-created qualifying issue across all miners.
3. **Marker comparison replaces per-miner set**: In `_score_miner_mirror_issues`, the per-miner `pr_scored_keys` set and its `pr_key in pr_scored_keys` / `pr_scored_keys.add(pr_key)` pair are replaced with an exact marker-tuple comparison against the canonical map.

Same-account issues (discoverer == solver) never claim the slot, mirroring the legacy `pr_scored.add` ordering. `_FAR_FUTURE` is hoisted to module scope so the canonical builder and scorer share it. The scorer's signature gains only `canonical_pr_owners`; its classification path and counters are unchanged. Public API, log shapes, and cache stats are unchanged — only which miner pockets the score on a shared PR changes.

## Related Issues

Closes #821

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -19,9 +19,10 @@ Anti-gaming gates (all applied):
 - issue.author_github_id != solving_pr.author_github_id (anti-self-issue)
 
 Same-account ("solver is also discoverer") gives credibility only — no
-discovery score. One-issue-per-PR rule prevents a single solving PR from
-generating multiple discovery scores when it closes more than one issue;
-the earliest-created qualifying issue wins.
+discovery score. One-issue-per-PR rule is round-global: a single solving PR
+awards at most one discovery score across the entire validator round, even
+when it closes issues authored by different miners. The earliest-created
+qualifying issue across all miners wins; the rest are credibility only.
 
 Base-score resolution uses a per-cycle cross-miner cache. Most solving PRs on
 mirror-enabled repos will be miners' own PRs that OSS scoring already tokenized;
@@ -96,6 +97,9 @@ class _CacheStats:
     fetch_failures: int = 0
 
 
+_FAR_FUTURE = datetime.max.replace(tzinfo=timezone.utc)
+
+
 async def run_mirror_issue_discovery(
     miner_evaluations: Dict[int, MinerEvaluation],
     mirror_repos: Dict[str, RepositoryConfig],
@@ -134,12 +138,14 @@ async def run_mirror_issue_discovery(
         f'{sum(len(ev.mirror_merged_prs) for ev in miner_evaluations.values())} scored mirror PRs'
     )
 
-    processed = 0
     skipped_no_gh = 0
     skipped_failed = 0
     fetch_errors = 0
     no_issues = 0
 
+    # Phase 1: fetch each miner's issues.  The one-issue-per-PR rule is round-
+    # global, so scoring is deferred until every miner's batch is in hand.
+    pending: List[Tuple[MinerEvaluation, List[MirrorIssue], int]] = []
     for uid, evaluation in miner_evaluations.items():
         if not evaluation.github_id or evaluation.github_id == '0':
             skipped_no_gh += 1
@@ -166,8 +172,10 @@ async def run_mirror_issue_discovery(
         # mirror-scoped state (the legacy GraphQL global open-issue count was
         # never the right signal for the gate).
         open_issue_count = sum(1 for i in filtered if i.state == 'OPEN')
+        pending.append((evaluation, filtered, open_issue_count))
 
-        processed += 1
+    canonical_pr_owners = _build_canonical_pr_owners(pending)
+    for evaluation, filtered, open_issue_count in pending:
         _score_miner_mirror_issues(
             evaluation,
             filtered,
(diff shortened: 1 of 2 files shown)


anderdc (changes requested):
Functionally correct, but more re-architecting than the bug needs. The minimum fix is: hoist the dedup decision to round scope. Everything else (the NamedTuple, the phase-1 partition) is incidental.

Concrete scope-down ask:

1. Drop `_MinerBatch`. Phase 1 just collects `pending: List[Tuple[MinerEvaluation, List[MirrorIssue], int]]` (eval, filtered, open_issue_count) — same as today, minus the inline classify loop.

2. `_build_canonical_pr_owners(pending)` classifies inline:

       for evaluation, filtered, _ in pending:
           for issue in filtered:
               if _classify_issue(issue) != 'solved':
                   continue
               sp = issue.solving_pr
               if issue.author_github_id == sp.author_github_id:
                   continue
               ...

   Yes, \`_classify_issue\` runs twice per issue (here + in the scorer). It's attribute checks — microseconds at miner scale. Not worth a NamedTuple.

3. \`_score_miner_mirror_issues\` keeps its current signature and inline classification path. The only change is replacing \`pr_scored_keys: Set[Tuple[str, int]] = set()\` and the \`pr_key in pr_scored_keys\` check with the marker comparison against \`canonical_pr_owners\`.

4. Pull \`_FAR_FUTURE = datetime.max.replace(tzinfo=timezone.utc)\` to module scope so the scorer and canonical builder share it instead of redefining.

Net diff should be ~50 lines instead of ~140, and the scorer barely changes — which is what you want for a scoring-economy fix where the diff itself is the audit surface.

Tests stay as-is; they target \`_build_canonical_pr_owners\` and end-to-end behavior, both unchanged by the refactor.

mkdev5:
@anderdc please review the update

anderdc (approved):

PR #786, opened 2026-04-24
perf(cache): replace deepcopy in MinerEvaluationCache with shallow copy

FairyPigDev:
Closes: #785

## Summary
- `MinerEvaluationCache.get` and `create_lightweight_copy` were running `copy.deepcopy` over the full evaluation tree (PRs + thousands of FileChanges + Issues) on every cache hit and every cache store. For miners with hundreds of PRs this cost hundreds of ms per UID per round.
- Cached evals only feed the GitHub-fetch-failure fallback path (issue bounties + issue discovery scoring), which never reads `file_changes` — so `store` now drops `file_changes` at cache time, and both `store` and `get` shallow-copy the evaluation + PRs while creating fresh `Issue` instances to isolate the cache from downstream `discovery_*` mutations.
- Net: copy cost drops from O(FileChanges) (with full patch bodies) to O(PRs + Issues).

## What changed
- `gittensor/classes.py`: replaced `deepcopy` in `MinerEvaluationCache` with two targeted helpers (`_build_cache_entry`, `_isolate_for_downstream`) plus PR-level shallow-copy helpers. Removed the now-unused `create_lightweight_copy` public method.
- `tests/validator/test_validator_cache_fallback.py`: added `TestCacheIsolation` locking down the three invariants the new strategy relies on — cache drops `file_changes`/`github_pat`, `get()` results are isolated from each other, and post-`store` mutations on the source eval don't leak into the cache.

## Test plan
- [x] `uv run pytest tests/validator/test_validator_cache_fallback.py` (6 passed, incl. 3 new)
- [x] `uv run pytest tests/` (444 passed)
- [x] `uv run ruff check` + `ruff format --check` on touched files

PR #824, opened 2026-04-27
fix: strip mirror file contents in MinerEvaluationCache

### LINKED ISSUE #823: [Bug]:`MinerEvaluationCache` doesn't strip mirror file contents — unbounded memory growth
## Problem

`MinerEvaluationCache.store()` drops `patch` strings from legacy PRs (via `_pr_for_cache`), but the equivalent mirror lists (`mirror_merged_prs`, `mirror_open_prs`, `mirror_closed_prs`) are stored as-is. The `head_content` and `base_content` blobs on `ScoredMirrorPR.files` — whole head/base file contents fetched from the mirror — stay in memory for the lifetime of the cached entry.

## Where

`gittensor/classes.py` — `MinerEvaluationCache._build_cache_entry` only iterates the legacy `*_pull_requests` lists. The three `mirror_*_prs` lists fall through `copy.copy(evaluation)` as shared references, so nothing strips their file payloads.

## Impact

- Cached evaluations grow with every mirror-scored PR and never shrink.
- Worst on validators scoring high-traffic mirror-enabled repos with large diffs (one big PR can pin multiple MB).
- The cache is per-UID and survives across rounds, so the leak compounds the longer the validator runs.

## Expected

After `cache.store(evaluation)`:

- Every `ScoredMirrorPR.files` on the cached copy is `None` (or each `MirrorFile.head_content` / `base_content` is `None`) across all three mirror buckets.
- The original `evaluation` passed in is untouched — DB storage downstream still needs the full file payloads.
- Legacy `FileChange.patch` stripping behavior is unchanged.

## Repro

1. Run a validator against a repo with `mirror_enabled: true` in `master_repositories.json`.
2. Score a miner that has at least one non-trivial mirror PR on that repo.
3. Inspect `validator.evaluation_cache._cache[uid].evaluation.mirror_merged_prs[0].files[0].head_content` — still populated with the full file contents.

carlos4s:
## Summary

`MinerEvaluationCache` got shallow-copy helpers in #786, but they only cover the legacy `PullRequest` lists. The three mirror lists (`mirror_merged_prs`, `mirror_open_prs`, `mirror_closed_prs`) still get shared by reference, which causes two real problems:

- **Memory:** cached evaluations keep `ScoredMirrorPR.files[*].head_content` and `base_content` (whole head/base file blobs). Cache memory grows with every mirror-scored PR and never shrinks (#823).
- **Aliasing:** mutations on the source eval after `store()` leak into the cache, and back-to-back `get()` calls hand back the same `ScoredMirrorPR` / `MirrorPullRequest` / `MirrorLinkedIssue` / `MirrorLabel` objects.

This PR extends the same shallow-copy strategy from #786 to mirror PRs, via four small private helpers that mirror the existing `_pr_for_cache` / `_pr_with_fresh_issues` pair:

- `_build_cache_entry`: drop `ScoredMirrorPR.files` and shallow-copy the nested `MirrorPullRequest` (`review_summary`, `labels`, `linked_issues` + each linked issue's `labels`).
- `_isolate_for_downstream`: same nested-metadata copy on the way out, so successive `get()` calls return fully independent snapshots.

The cache-fallback path only runs issue competitions and issue discovery scoring, neither of which reads file contents — so dropping `files` outright is enough, no need to keep stripped `MirrorFile` placeholders.

## Related Issues

Closes #823

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Three new tests in `tests/validator/test_validator_cache_fallback.py::TestMirrorCacheIsolation`, one per invariant:

- `test_cache_drops_mirror_files_without_mutating_source` — `files` is `None` on the cached copy across all three mirror buckets, and the source eval is untouched.
- `test_store_isolates_cached_mirror_metadata_from_source_mutations` — mutating `source.mirror_merged_prs[i].pr.{title, review_summary, labels, linked_issues}` after `store()` does not affect the cached copy.
- `test_get_returns_isolated_mirror_metadata` — two consecutive `get()` calls return independent snapshots.

```
$ uv run pytest tests/validator/test_validator_cache_fallback.py -v
9 passed

$ uv run pytest tests/
all green

$ uv run ruff check && uv run ruff format --check
clean

$ uv run pyright gittensor/classes.py tests/validator/test_validator_cache_fallback.py
0 errors
```

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -708,6 +708,9 @@ class MinerEvaluationCache:
         cached.merged_pull_requests = [_pr_for_cache(pr) for pr in evaluation.merged_pull_requests]
         cached.open_pull_requests = [_pr_for_cache(pr) for pr in evaluation.open_pull_requests]
         cached.closed_pull_requests = [_pr_for_cache(pr) for pr in evaluation.closed_pull_requests]
+        cached.mirror_merged_prs = [_scored_mirror_pr_for_cache(pr) for pr in evaluation.mirror_merged_prs]
+        cached.mirror_open_prs = [_scored_mirror_pr_for_cache(pr) for pr in evaluation.mirror_open_prs]
+        cached.mirror_closed_prs = [_scored_mirror_pr_for_cache(pr) for pr in evaluation.mirror_closed_prs]
         return cached
 
     @staticmethod
@@ -735,3 +738,9 @@ def _pr_with_fresh_issues(pr: 'PullRequest') -> 'PullRequest':
     if pr.issues is not None:
         pr_copy.issues = [copy.copy(issue) for issue in pr.issues]
     return pr_copy
+
+
+def _scored_mirror_pr_for_cache(scored: 'ScoredMirrorPR') -> 'ScoredMirrorPR':
+    scored_copy = copy.copy(scored)
+    scored_copy.files = None
+    return scored_copy
--- a/tests/validator/test_validator_cache_fallback.py
+++ b/tests/validator/test_validator_cache_fallback.py
@@ -4,6 +4,8 @@ from datetime import datetime, timezone
 from typing import cast
 
 from gittensor.classes import FileChange, Issue, MinerEvaluation, MinerEvaluationCache, PRState, PullRequest
+from gittensor.utils.mirror.models import MirrorFile, MirrorPullRequest, MirrorReviewSummary
+from gittensor.validator.oss_contributions.mirror.scored_pr import ScoredMirrorPR
 from neurons.validator import Validator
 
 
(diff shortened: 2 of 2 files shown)


carlos4s:
@anderdc Please review this PR. I'd appreciate any feedbacks. Thanks

anderdc (changes requested):
Scope this down to the actual leak. The file-strip change in `_build_cache_entry` plus a `_scored_mirror_pr_for_cache` helper that nulls `files` is all #823 asks for.

Drop `_scored_mirror_pr_with_fresh_pr`, `_mirror_pr_with_fresh_metadata`, `_mirror_linked_issue_with_fresh_labels`, the `_isolate_for_downstream` mirror branch, and the two metadata-isolation tests (`test_store_isolates_cached_mirror_metadata_from_source_mutations`, `test_get_returns_isolated_mirror_metadata`) — those guard against mutations no caller on the cache-fallback path actually makes. The legacy `_pr_with_fresh_issues` deliberately only deep-copies `issues` for the same reason; see the comment at `classes.py:723-725` ("PR metadata is read-only on the cache-fallback path, so we can share it"). The mirror path feeds the same downstream (issue competitions + issue discovery), so the same logic applies.

Keep `test_cache_drops_mirror_files_without_mutating_source` — that one matches the actual issue invariant.

If aliasing on metadata ever becomes a real concern, land it in its own PR with the offending caller cited.

carlos4s:
@anderdc pushed the update. plz review again.

anderdc (approved):

PR #825, opened 2026-04-27
fix: add request-level rate limiting to MirrorClient to respect Cloudflare 50 req/10s limit

wdeveloper16:
## Summary

Follow-up to #796. `MirrorClient` documented that callers are responsible for respecting Cloudflare's 50 req/10s rate limit, but neither `load.py` nor `mirror_scan.py` implemented any throttle. Under load with many miners, this risks 429 responses that the retry loop alone cannot absorb efficiently.

- Add `_throttle()` to `MirrorClient`: enforces a minimum interval between  outgoing requests using `time.monotonic` + `threading.Lock` (thread-safe)
- Default interval is `MIRROR_MIN_REQUEST_INTERVAL_SECONDS = 0.2s` (5 req/s),  a 10× safety margin under the Cloudflare limit
- Expose `min_request_interval` as a constructor param; test helpers pass `0`  to disable throttling so existing retry tests are unaffected
- Add `TestThrottle` with two cases: sleep fires when interval hasn't elapsed,  and `min_request_interval=0` never sleeps

## No behaviour changes for non-throttled paths

All existing retry/backoff logic is untouched. On retries the backoff sleep (5–30s) far exceeds the 0.2s interval so `_throttle()` adds no extra delay during retry sequences.

## Test plan

- [ ] `uv run --with pytest pytest tests/utils/test_mirror_client.py -v`  → 23 passed

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -35,6 +35,8 @@ GITTENSOR_MIRROR_DEFAULT_URL = 'https://mirror.gittensor.io'
 # File endpoint returns head/base blob contents; allow more time than plain GitHub calls.
 MIRROR_HTTP_TIMEOUT_SECONDS = 30
 MIRROR_MAX_ATTEMPTS = 3
+# Cloudflare rate limit is 50 req/10s per IP. 200ms floor = 5 req/s = 10x safety margin.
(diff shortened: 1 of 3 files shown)


wdeveloper16:
Hey, @anderdc 
What is the reason to close the PR?

PR #828, opened 2026-04-28
fix(cli): emit consensus_threshold=0 in vote list JSON when whitelist is empty

### LINKED ISSUE #753: [Bug] gitt vote list --json reports consensus_threshold=1 when validator whitelist is empty
## Summary

`gitt vote list --json` ([`gittensor/cli/issue_commands/vote.py:236-249`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/issue_commands/vote.py#L236)) computes `required = (n // 2) + 1` where `n = len(validators)`, then emits the value as `consensus_threshold` regardless of whether any validators are whitelisted.

When the whitelist is empty, the JSON payload reports `{"success": true, "count": 0, "consensus_threshold": 1}`. There is no integer number of votes that satisfies a zero-validator quorum, so the field is mathematically meaningless in this case but is still typed as a positive integer.

The human-mode branch (lines 251-264) prints `"No validators whitelisted."` and does not show the threshold, so the inconsistency is JSON-only.

## Steps to reproduce

1. Point at a test contract with no validators on the whitelist (none added, or all removed).
2. Run `gitt vote list --json --network test`.
3. Output:
   ```json
   {"success": true, "validators": [], "count": 0, "consensus_threshold": 1}
   ```

## Expected behavior

When `count == 0`, `consensus_threshold` should not silently report a positive integer. Either the field is omitted / null / 0, or `success: false` is set with an explanatory error.

## Actual behavior

`consensus_threshold: 1` with `success: true`. JSON consumers (dashboards, monitoring) cannot detect the empty-whitelist edge case from the field.

## Additional context

Same family as [#735](https://github.com/entrius/gittensor/issues/735) (`gitt vote list --json` JSON shape inconsistency), which was already merged. The empty-whitelist edge case is not covered by that fix.

ebios-star:
## Summary

Closes #753.

\`gitt vote list --json\` ([cli/issue_commands/vote.py:248-249](gittensor/cli/issue_commands/vote.py#L248-L249)) computed \`required = (n // 2) + 1\` unconditionally, so an empty validator whitelist produced \`{\"count\": 0, \"consensus_threshold\": 1}\` — a 1-of-0 threshold that no integer vote count can satisfy.

JSON consumers (dashboards, monitoring scripts) could not detect the empty-whitelist edge case from this field alone, even though the human-mode branch already prints \"No validators whitelisted.\" and skips the threshold line entirely.

Now: \`consensus_threshold == 0\` when no validators are whitelisted; the existing \`(n // 2) + 1\` majority is preserved for non-empty whitelists.

## Related Issues

Closes #753.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- New \`tests/cli/test_vote_list_consensus.py\` covers three cases:
  - empty whitelist → \`consensus_threshold == 0\`
  - 3 validators → \`consensus_threshold == 2\`
  - 1 validator → \`consensus_threshold == 1\`
- \`uv run --extra dev pytest tests/\` — 446 tests pass (443 baseline + 3 new).

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/cli/issue_commands/vote.py
+++ b/gittensor/cli/issue_commands/vote.py
@@ -246,7 +246,7 @@ def vote_list_validators(network: str, rpc_url: str, contract: str, as_json: boo
             validators = client.get_validators()
 
         n = len(validators)
-        required = (n // 2) + 1
+        required = (n // 2) + 1 if n > 0 else 0
 
         if as_json:
             emit_json(


ebios-star:
@anderdc @landyndev — friendly ping for review when you have a moment.

Closes #753. When the validator whitelist is empty, `gitt vote list --json` previously emitted `{"count": 0, "consensus_threshold": 1}` — a 1-of-0 threshold that no integer vote count can satisfy. JSON consumers (dashboards, monitoring) now see `consensus_threshold: 0` for empty whitelists. Existing `(n // 2) + 1` majority preserved for non-empty cases. 3 new tests, full test suite passes.

Happy to rebase if needed.

anderdc (changes requested):
remove the tests you added (disproportionate to the change) and remove the comments you added, one liner doesn't need all those comments

ebios-star:
Done @anderdc — thanks for the review.

Stripped the test file and the inline comment; the change is now a single one-line diff:

```diff
-        required = (n // 2) + 1
+        required = (n // 2) + 1 if n > 0 else 0
```

Branch rebased onto current `test` and force-pushed (`--force-with-lease`).

anderdc (approved):

PR #516, opened 2026-04-17
fix: prevent stale cache fallback for legitimate zero-PR miners

MkDev11:
## Summary

This PR fixes a cache fallback bug where miners with legitimate zero PRs could be assigned stale cached evaluations.

Changes:
- Added explicit `github_pr_fetch_failed` signal to `MinerEvaluation`.
- Updated `load_miners_prs` to set this flag only on actual fetch-failure/error paths.
- Updated validator cache logic to:
  - store cache only for successful non-failed fetches with PRs
  - fallback to cache only when fetch failed and no PRs were loaded
- Added regression tests for fetch-failure signaling and cache fallback behavior.

## Related Issues

Closes #515

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

Commands run:
- `uv run pytest tests/utils/test_github_api_tools.py::TestLoadMinersPrsFetchFailureSignal -v`
- `uv run pytest tests/validator/test_validator_cache_fallback.py -v`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pyright`
- `uv run pytest tests/ -v`

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #830, opened 2026-04-28
fix: hybrid-merge cached mirror PRs when only mirror fetch fails

### LINKED ISSUE #681: Skip scoring CLOSED PRs in `score_miner_prs` to save GitHub API budget
## Summary

`score_miner_prs` calls `score_pull_request` for every CLOSED PR, triggering 2+ GitHub API calls each (file changes + file contents + merge-base). The computed `base_score` for closed PRs is never used in reward calculation — only `len(closed_prs)` matters for credibility. Skipping scoring for closed PRs would save API budget with no effect on miner rewards.

## Motivation

Each scored closed PR costs 2–3 GitHub API calls:

1. `get_pull_request_file_changes` — REST call for the diff
2. `fetch_file_contents_for_pr` — GraphQL batch for base + head file contents
3. Merge-base resolution (implicit)

These scores are unused in rewards:

- **Reward aggregation** (Phase 3, lines 406–408) only sums `earned_score` from `merged_pull_requests`
- **Credibility** (`check_eligibility`) only uses `len(closed_prs)` — not their scores
- **Eligibility gate** counts closed PRs for the credibility ratio but never reads `base_score`

The only consumer is the DB via `store_pull_requests_bulk`, which persists them for analytics.

**Proposed fix** — filter `pr_groups` in `scoring.py:65-69`:

```python
pr_groups = [
    ('MERGED', miner_eval.merged_pull_requests),
    ('OPEN', miner_eval.open_pull_requests),
]
```

Closed PRs would still be loaded, counted for credibility, and stored — they just wouldn't be scored (`base_score` stays `0.0`).

**Trade-off:** Closed-PR `base_score` and `file_changes` will no longer be populated in the DB. If dashboards or analytics rely on closed-PR scores, they would need updating. If that data is not consumed, this is a clean API-budget win.

mkdev5:
## Summary

Mixed legacy + mirror miners are silently zeroed when `mirror.gittensor.io` is down. `combine` OR's `mirror_eval.fetch_failed` into the legacy `github_pr_fetch_failed`, but `should_use_cache_fallback` still requires `total_prs == 0` — so a successful legacy fetch (1+ PR) blocks the fallback and the validator finalizes a partial evaluation missing every mirror PR. Reach: any miner with PRs in both an `entrius/*` mirror repo and any non-mirror repo, during a mirror outage. Drops past `MIN_VALID_MERGED_PRS` → 0 OSS reward.

Adds a per-source `mirror_pr_fetch_failed` flag (mirror is binary — atomic load), wires it through `combine`, and routes mixed mirror-only failures through a hybrid merge in `store_or_use_cached_evaluation`: keep the fresh legacy data, splice cached mirror PRs back in, and re-cache the result. Cached `ScoredMirrorPR` reuse is safe because `base_score`/`token_score` are deterministic over the merged diff; credibility and other cross-PR multipliers get recomputed in `finalize_miner_scores`.

The pre-mirror legacy partial-pagination semantic (#516) is preserved — that path uses the existing `should_use_cache_fallback` gate untouched. Pure-mirror miners and full-failure cases also keep the existing full-eval fallback.

## Related Issues

Closes #681

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Checklist

- [x] Self-review completed


anderdc (changes requested):
The bug is real and the diagnosis is correct, but the hybrid-merge approach is more invasive than the situation warrants given the legacy path is on the chopping block.

The whole `combine` module is documented as going away on delete-day ("On delete-day this whole module goes away — at that point MirrorMinerEvaluation becomes the canonical container"). Once legacy is gone:

- `mirror_pr_fetch_failed` collapses back into a single failure flag
- The mixed-source problem stops existing entirely
- A mirror outage just means "no data this round, use cached eval" — same shape as legacy works today

Everything the hybrid splice adds (`_splice_cached_mirror_into`, the dedicated conditional block, the re-cache step) is transition-only scaffolding that we'll then have to delete. And during the transition it produces a mixed-snapshot eval (fresh legacy PRs + stale-cached mirror PRs evaluated together) where credibility and other cross-PR multipliers get recomputed across a Frankenstein view of the miner's state.

Suggested simpler fix:

1. Keep the `mirror_pr_fetch_failed` flag and the `combine` wiring — that part is necessary signal.

2. In `store_or_use_cached_evaluation`, just let mirror failures through the existing cache-fallback path instead of building a new branch. Roughly:

       if not miner_eval.should_use_cache_fallback and not miner_eval.mirror_pr_fetch_failed:
           bt.logging.warning(...)
           continue

       cached_eval = self.evaluation_cache.get(...)
       if cached_eval is not None:
           bt.logging.info(...)
           miner_evaluations[uid] = cached_eval
           cached_uids.add(uid)

3. Drop `_splice_cached_mirror_into` and the dedicated hybrid block in `neurons/validator.py` entirely.

The trade-off: during a mirror outage, mixed-source miners get evaluated against a coherent one-round-stale snapshot instead of a fresh-legacy + stale-mirror hybrid. They lose the freshness of *this round's* legacy PRs for one cycle. Given mirror outages are transient and we're heading toward mirror-only anyway, that's the right trade — less code, no cross-source contamination, and nothing extra to delete on delete-day.

Tests: `test_combine.py` assertions stay; the validator-side tests can be simplified to assert "mirror failure with cached eval available → swapped to cached eval, returned in cached_uids."

mkdev5:
pushed the update

anderdc (approved):

PR #50, opened 2025-12-02
Fix issue multiplier to use validated issues list

NikolaPantel:
Bug: calculate_issue_multiplier() was using the unfiltered pr.issues list instead of the filtered valid_issues list, allowing invalid issues (self-created, wrong timing, etc.) to be scored.

Impact: Miners could game the system by adding self-created issues to their PRs and still receive issue bonus multipliers.

Fix: Changed 2 lines to use valid_issues instead of pr.issues:
- Line 190: num_issues = min(len(valid_issues), ...)
- Line 195: issue = valid_issues[i]

This ensures only issues that pass _is_valid_issue() validation are scored, preventing gaming and maintaining fairness.

Contribution by Gittensor, learn more at https://gittensor.io/

PR #832, opened 2026-04-28
fix: enforce one-issue-per-PR across miners in mirror issue discovery

### LINKED ISSUE #821: [Bug] Mirror issue discovery one-issue-per-PR rule is enforced per-miner rather than per-round

ClaudeCenter:
## Summary
- Closes #821
- Restores round-global one-issue-per-PR enforcement that #796 narrowed
  to per-miner scope, closing the two-account collusion vector on
  multi-issue solving PRs
- Alternative implementation to #822 — same observable result, different
  internal mechanics

## What was broken
PR #796 moved the `pr_scored_keys` dedup set into `_score_miner_mirror_issues`,
where it only ever sees a single miner's issues. A solving PR closing issues
authored by miners A and B (e.g., a third party merges a fix that closes
both A's #50 and B's #51) awarded full discovery score to *each* miner.
Documented behavior: one score per solving PR per round, earliest-created
qualifying issue wins.

## How this fixes it
- **Phase 1** fetches each miner's issues and partitions them by
  `_classify_issue` into a `_MinerBatch(evaluation, solved_issues,
  closed_count, open_issue_count)` `NamedTuple`. `_classify_issue` runs
  once per issue across the cycle.
- **`_build_canonical_pr_owners`** walks every miner's `solved_issues`
  and produces `(repo, pr_number) → (created_at, issue_number, uid)` for
  the earliest-created qualifying issue across all miners. Same-account
  issues (discoverer == solver) are excluded — they never claim the slot.
- **`_score_miner_mirror_issues`** consumes the batch directly, comparing
  each issue's marker against `canonical_pr_owners[pr_key]`. Mismatches
  drop to credibility-only, matching legacy pre-#796 semantics.

### DIFF
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -18,9 +18,10 @@ Anti-gaming gates (all applied):
 - issue.author_github_id != solving_pr.author_github_id (anti-self-issue)
 
 Same-account ("solver is also discoverer") gives credibility only — no
-discovery score. One-issue-per-PR rule prevents a single solving PR from
-generating multiple discovery scores when it closes more than one issue;
-the earliest-created qualifying issue wins.
+discovery score. One-issue-per-PR rule is round-global: a single solving PR
+awards at most one discovery score across the entire validator round, even
+when it closes issues authored by different miners. The earliest-created
+qualifying issue across all miners wins; the rest are credibility only.
 
 Base-score resolution uses a per-cycle cross-miner cache. Most solving PRs on
 mirror-enabled repos will be miners' own PRs that OSS scoring already tokenized;
@@ -32,7 +33,7 @@ the cache so sibling discoveries benefit.
 
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, List, NamedTuple, Optional, Set, Tuple
 
 import bittensor as bt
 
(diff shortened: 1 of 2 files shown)


anderdc:
Duplicate of #822 which is better scoped/implemented. Closing.

PR #834, opened 2026-04-28
feat: support HTTP ETag cache for GitHub REST GETs

### LINKED ISSUE #833: feat: support HTTP ETag cache for GitHub REST GETs
## Summary

Add an HTTP ETag cache for GitHub REST GETs in `gittensor/utils/github_api_tools.py` so repeat round-to-round calls return `304 Not Modified` instead of doing requests from scratch every time and consuming rate-limit quota.

## Motivation

A validator scoring round makes thousands REST calls through `gittensor/utils/github_api_tools.py`. The bulk of these are highly repetitive across rounds:

- `get_merge_base_sha` for the same `(repo, base_sha, head_sha)` triples (immutable for any merged PR)
- `get_pull_request_file_changes` for the same PRs whose `head_sha` has not changed
- `check_github_issue_closed` for the same active bounty issues
- `_search_issue_referencing_prs_rest` for the same issues across CLI invocations

Today, none of these are cached at the HTTP layer:

- No `ETag` / `If-None-Match` / `If-Modified-Since` handling
- No on-disk response cache
- Storage (`gittensor/validator/storage/repository.py`) is write-only - no read-back path either

GitHub's primary rate-limit docs guarantee that `304 Not Modified` responses do not consume quota: https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2026-03-10#use-conditional-requests-if-appropriate

Additionally, re-using the cache must be faster than performing the actual request.

By persisting `(etag, body)` per request URL and sending `If-None-Match` on subsequent calls, we can convert the majority of these repeated REST calls into free and fast 304s.

seroperson:
Closes #833

## Summary

Add an on-disk ETag cache for GitHub REST GETs so repeat round-to-round calls return `304 Not Modified` (which [does not consume primary rate quota](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api#use-conditional-requests-if-appropriate)). Wraps four hot-path REST functions and emits per-round hit/miss telemetry from the validator forward loop.

Cache is implemented using [Apache DiskCache](https://github.com/grantjenks/python-diskcache) library

## Before / After

**Before**: every REST call from `oss_contributions/scoring.py` and `issue_competitions/forward.py` consumes rate-limit quota even if the underlying resource has not changed since the last round.

**After**: first call stores the ETag in `~/.gittensor/cache/github_etag/`; subsequent calls send `If-None-Match`, GitHub returns 304, and the cached body is replayed via a synthesized `requests.Response` (with `X-Gittensor-Cache: hit`). Each scoring round logs `etag cache: hits=X misses=Y hit_rate=Z%` so the savings are observable.

## Tests

- `tests/utils/test_github_etag_cache.py` covers store/lookup/TTL/disabled-mode/counters
- `TestCachedGithubGet` in `tests/utils/test_github_api_tools.py` covers 200+ETag store, 304 hit, no-ETag no-cache, 5xx falls through
- `tests/utils/conftest.py` autouse fixture isolates the cache to `tmp_path`.

## Design notes

- Always-revalidate behavior: every call sends `If-None-Match` if there is a stored ETag; 304 → return cached body, 200 → refresh. The cache only converts 200s into 304s and never serves stale data.
- Disabling: `GITTENSOR_ETAG_CACHE_DISABLED=1` short-circuits to no-op (the SQLite DB is never opened). Used for debugging and scoring-parity verification. `GITTENSOR_ETAG_CACHE_PATH` overrides the default location.
- Storage: a single `diskcache.Cache` directory at `~/.gittensor/cache/github_etag/` (SQLite + WAL). For a hard reset: `rm -rf ~/.gittensor/cache/github_etag/`.

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -27,6 +27,9 @@ GRAPHQL_VIEWER_QUERY = '{ viewer { login } }'
 MAX_FILE_SIZE_BYTES = 1_000_000
 # Too many object lookups in one GraphQL query can trigger 502 errors and lose all results.
 MAX_FILES_PER_GRAPHQL_BATCH = 50
+# On-disk ETag cache for GitHub REST GETs - 304 responses do not consume rate quota.
+ETAG_CACHE_TTL_DAYS = 30
+ETAG_CACHE_DEFAULT_PATH = '~/.gittensor/cache/github_etag'
 
 # =============================================================================
 # das-github-mirror (https://mirror.gittensor.io)
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 
 import bittensor as bt
 import requests
+from requests.structures import CaseInsensitiveDict
 
 from gittensor.classes import (
     FileChange,
@@ -30,6 +31,7 @@ from gittensor.constants import (
     PR_LOOKBACK_DAYS,
     REVIEW_PENALTY_RATE,
 )
+from gittensor.utils.github_etag_cache import build_request_key, get_default_cache, record_hit, record_miss
 from gittensor.utils.models import PRInfo
 from gittensor.validator.utils.datetime_utils import parse_github_iso_to_utc
 from gittensor.validator.utils.load_weights import RepositoryConfig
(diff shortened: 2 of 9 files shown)


anderdc:
Once everything moves to mirror repos, this goes away — REST traffic to api.github.com isn't the long-term shape. In the meantime, the uncacheable GraphQL calls are the bigger share of quota usage, not the wrapped REST endpoints, and per-miner PAT utilization isn't high enough for the proportional gain to justify a 475-line on-disk cache + new dep + env vars. Closing.

PR #835, opened 2026-04-28
fix: exclude self-filed-and-self-solved issues from mirror valid-solved gate

polyjeff529:
```
## Summary

The mirror issue-discovery path increments `valid_solved_count` before the
same-account (self-issue) check in `_score_miner_mirror_issues`
(`gittensor/validator/issue_discovery/mirror_scan.py:266-298`). A miner
who files an issue from their own GitHub account and solves it with a
PR from the same account currently has that solve count toward the
`MIN_VALID_SOLVED_ISSUES = 7` eligibility gate, even though the
discovery-score path correctly skips it (gives credibility only, no
score).

Result: a miner with six legitimate cross-account discoveries plus a
single self-filed-and-self-solved issue clears the gate on padding
rather than work — six real cross-account solves alone would fail it.
The seventh "solve" proves no cross-account discovery activity, which is
exactly what the gate exists to verify per the existing docstring of
`check_issue_eligibility` ("low-quality solves don't carry the miner
past the minimum").

## Fix

`mirror_scan.py:285-301`

Compute `is_self_solved` once and gate `valid_solved_count++` on it.
The same flag is reused for the existing same-account credibility-only
branch (no change to that branch's behaviour).

```python
is_self_solved = issue.author_github_id == solving_pr.author_github_id

# Self-issues are excluded from valid_solved — the gate exists to verify
# cross-account discovery work.
if not is_self_solved and cached.token_score >= MIN_TOKEN_SCORE_FOR_BASE_SCORE:
    valid_solved_count += 1

if is_self_solved:
    bt.logging.debug(...)
    continue
```

### DIFF
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -285,12 +285,19 @@ def _score_miner_mirror_issues(
             )
             continue
 
+        is_self_solved = issue.author_github_id == solving_pr.author_github_id
+
         # Valid-solved gate (legacy parity): solving PR must meet the token threshold.
-        if cached.token_score >= MIN_TOKEN_SCORE_FOR_BASE_SCORE:
+        # Self-issues are excluded — the gate exists to verify the miner did real
+        # cross-account discovery work, and a self-filed-and-self-solved issue is
+        # not evidence of that. Without this exclusion, a miner with 6 real
+        # cross-account discoveries plus a single self-filed-and-self-solved issue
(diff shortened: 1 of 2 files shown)


anderdc:
Premise doesn't hold. Self-solved issues count toward eligibility/credibility because they are real solving activity — the system already handles their lower economic value correctly at the score path (credibility only, no reward), so excluding them from the gate too would double-penalize the same activity at a surface that wasn't designed to do that work. The 'padding' concern doesn't survive incentive analysis: a miner gets zero score from a self-solved issue, so there's no economic gain to manufacture one, and detect_and_penalize_miners_sharing_github already covers Sybil-style cross-account farming. Closing.

PR #859, opened 2026-04-29
fix: close all sessions in session_scope even when one close() raises (follow-up of #687)

### LINKED ISSUE #858: fix: session_scope leaks remaining sessions when close() raises
## Bug
Introduced in #687 (`perf: pool http connections per miner scoring iteration`).

In `session_scope()`'s `finally` block, if the first `session.close()` raises an exception, the loop aborts and all remaining sessions in the cache are never closed:

```python
finally:
    cache = _session_cache
    _session_cache = None
    for s in cache.values():
        s.close()  # if this raises, the rest are leaked
```

## Expected behaviour

All sessions should be attempted regardless of individual failures. The first exception, if any, should be re-raised only after all sessions have been closed.

wdeveloper16:
Follow-up to #687. Closes #858.

## Problem

In `session_scope()`'s `finally` block, if the first `session.close()` raises, the loop aborts and every remaining session in the cache is leaked:

```python
for s in cache.values():
    s.close()  # raises here → remaining sessions never closed
```

## Fix

Wrap each `close()` individually so all sessions are always attempted. The first exception, if any, is re-raised after the full iteration:

```python
first_exc: Optional[Exception] = None
for s in cache.values():
    try:
        s.close()
    except Exception as exc:
        if first_exc is None:
            first_exc = exc
if first_exc is not None:
    raise first_exc
```

## Test

Adds `test_all_sessions_closed_even_when_first_close_raises` to `TestSessionScope`: two sessions are created, the first `close()` is made to raise, and the test asserts both sessions had `close()` called and `_session_cache` is `None`.

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -202,8 +202,15 @@ def session_scope() -> Iterator[None]:
     finally:
         cache = _session_cache
         _session_cache = None
+        first_exc: Optional[Exception] = None
         for s in cache.values():
-            s.close()
+            try:
+                s.close()
+            except Exception as exc:
+                if first_exc is None:
+                    first_exc = exc
+        if first_exc is not None:
+            raise first_exc
 
 
 def _build_session(token: str) -> requests.Session:
--- a/tests/utils/test_github_api_tools.py
+++ b/tests/utils/test_github_api_tools.py
@@ -1645,6 +1645,28 @@ class TestSessionScope:
 
         assert github_api_tools._session_cache is None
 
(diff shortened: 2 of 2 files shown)


wdeveloper16:
Hi, @anderdc 
I'd appreciate your feedback with this.

wdeveloper16:
Hi, @anderdc 
Could you please review the PR and share me your feedback?

anderdc:
`requests.Session.close()` doesn't raise in the standard urllib3 stack. On the off chance it did, `_session_cache = None` drops the references right before the loop and GC reclaims the rest. Defensive guard for a scenario the underlying library doesn't produce. Closing.

PR #863, opened 2026-04-29
refactor: simplify get_all_uids — drop dead exclude param and redundant set() wrap

### LINKED ISSUE #861: [Refactor] get_all_uids has dead exclude parameter and redundant set() wrap
## Description

\`get_all_uids\` in [gittensor/utils/uids.py](gittensor/utils/uids.py) has three small issues that compound into a confusing API:

\`\`\`python
def get_all_uids(self, exclude: List[int] = []) -> set[int]:
    available_miner_uids = {uid for uid in range(self.metagraph.n.item()) if uid not in exclude}
    available_miner_uids.add(0)
    return set(available_miner_uids)
\`\`\`

1. **`exclude: List[int] = []`** — mutable default argument (PEP 8 anti-pattern). \`grep -rn get_all_uids\` confirms exactly one caller (\`gittensor/validator/forward.py:61\`) and it never passes \`exclude\`, so the parameter is dead.

2. **The \`exclude\` parameter is semantically broken even when used.** Line 18 unconditionally re-adds UID 0 to the set, so \`get_all_uids(self, exclude=[0])\` still returns a set containing 0. Callers wanting to truly exclude a UID would be silently betrayed for UID 0.

3. **Redundant \`set()\` wrap.** \`available_miner_uids\` is already a set built from a set comprehension (line 15) plus \`.add(0)\`. The outer \`set(...)\` call on line 20 reconstructs an identical set.

## Suggested refactor

\`\`\`python
def get_all_uids(self) -> set[int]:
    \"\"\"Return all miner UIDs for scoring.

    UID 0 is always included (subnet requirement).
    \"\"\"
    available_miner_uids = set(range(self.metagraph.n.item()))
    available_miner_uids.add(0)
    return available_miner_uids
\`\`\`

\`set(range(N))\` includes 0 for any \`N >= 1\`; the explicit \`.add(0)\` only matters at network bootstrap when \`N == 0\`, which is preserved.

## Impact

Cosmetic / API-hygiene. No behavior change for the existing caller. Eliminates a mutable-default smell, a misleading parameter, and a redundant collection allocation. Verifiable in one place — the single caller in \`forward.py\` does not pass \`exclude\`.

ebios-star:
## Summary

Closes #861.

\`get_all_uids\` in [gittensor/utils/uids.py](gittensor/utils/uids.py) had three small but compounding issues:

1. **Mutable default argument** — \`exclude: List[int] = []\`. \`grep -rn get_all_uids\` confirms the lone caller (\`validator/forward.py\`) never passes \`exclude\`, so the parameter is dead.
2. **The exclude parameter is semantically broken even when used.** The function unconditionally re-adds UID 0 to the set, so a caller asking to exclude UID 0 would be silently betrayed.
3. **Redundant \`set()\` wrap.** The return value re-wraps an already-set value built from a set comprehension plus \`.add(0)\`.

After this change \`get_all_uids(self)\` returns \`set(range(n)) | {0}\`. The bootstrap invariant (\`n == 0\` still yields \`{0}\`) is preserved by an explicit \`.add(0)\`.

Net: **-16 / +5 lines** in the source file, plus a fresh \`tests/utils/test_uids.py\` that pins the contract.

## Related Issues

Closes #861.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

\`tests/utils/test_uids.py\` covers four cases:
- \`n=5\` → returns \`{0..4}\` (standard range).
- \`n=0\` → returns \`{0}\` (bootstrap invariant preserved).
- \`n=1\` → returns \`{0}\` (single-UID).
- regression guard ensuring \`exclude=\` is no longer accepted (\`TypeError\`).

\`uv run --extra dev pytest tests/\` — all 686 tests pass (682 baseline + 4 new).

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/uids.py
+++ b/gittensor/utils/uids.py
@@ -1,20 +1,11 @@
-from typing import List
-
-
-def get_all_uids(self, exclude: List[int] = []) -> set[int]:
-    """Return all eligible miner UIDs for scoring.
-
-    Args:
-        exclude: UIDs to omit from the returned set.
+def get_all_uids(self) -> set[int]:
+    """Return all miner UIDs for scoring.
 
     Returns:
-        Set of miner UIDs that are serving and within the validator-permit TAO limit.
-        UID ``0`` is always included.
+        Set of miner UIDs in the metagraph. UID ``0`` is always included
+        (subnet requirement) so the empty-metagraph bootstrap case still
+        yields ``{0}`` rather than ``set()``.
     """
-    # Get all available miner UIDs, excluding specified ones
-    available_miner_uids = {uid for uid in range(self.metagraph.n.item()) if uid not in exclude}
-
-    # Ensure miner UID 0 is always included (subnet requirement)
+    available_miner_uids = set(range(self.metagraph.n.item()))
     available_miner_uids.add(0)
-
-    return set(available_miner_uids)
+    return available_miner_uids
--- a/tests/utils/test_uids.py
+++ b/tests/utils/test_uids.py
@@ -1,0 +1,47 @@
+# The MIT License (MIT)
+# Copyright © 2025 Entrius
+
+"""Tests for gittensor.utils.uids.get_all_uids."""
+
+from types import SimpleNamespace
+
+import pytest
+
+from gittensor.utils.uids import get_all_uids
+
+
+class _FakeMetagraphSize:
+    """Mimic the bittensor metagraph's `n.item()` interface."""
+
+    def __init__(self, n: int):
+        self._n = n
+
+    def item(self) -> int:
+        return self._n
+
+
+def _make_self(n: int) -> SimpleNamespace:
+    return SimpleNamespace(metagraph=SimpleNamespace(n=_FakeMetagraphSize(n)))
+
+
+def test_get_all_uids_returns_full_range():
+    result = get_all_uids(_make_self(5))
+    assert result == {0, 1, 2, 3, 4}
+    assert isinstance(result, set)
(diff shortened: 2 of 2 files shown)


ebios-star:
Hi @anderdc @landyndev — friendly review request whenever you have a moment.

This PR closes #861. `get_all_uids` had a mutable-default `exclude: List[int] = []` parameter that the single caller never passes, and that the function's own logic silently overrides for UID 0 anyway. Removing the dead parameter and the redundant `set(...)` wrap drops 16 lines and tightens the signature without changing behaviour.

Tests: 4 new cases in `tests/utils/test_uids.py` covering the standard range, the empty-metagraph bootstrap invariant, the single-UID case, and a guard that the dead parameter cannot reappear. Full suite (686) passes.

Happy to address any feedback or rebase if needed.

anderdc:
Too much churn for a debatable hygiene win. The exclude parameter is dead — never called, never mutated — so the 'mutable default' and 'silently betrayed for UID 0' concerns describe call shapes that don't exist. If something were actively being typed wrong or misused at a real call site, that'd be worth the refactor; tidying an unused signature isn't. Closing.

ebios-star:
Thanks @anderdc — that's a fair call and a useful framing. "Active misuse at a real call site" vs. "tidying an unused signature" is a clean line; I'll calibrate future refactors to that bar. Closing #861 alongside this PR.

PR #870, opened 2026-04-29
refactor: inline single-use _zero_for_duplicate_penalty

### LINKED ISSUE #869: [Refactor] Inline `_zero_for_duplicate_penalty` (single call site, fragile field list)
## Summary

`gittensor/validator/oss_contributions/inspections.py:_zero_for_duplicate_penalty` is called from exactly one site:
`detect_and_penalize_miners_sharing_github`. The helper exists for readability, but it hides the field-list. Inlining keeps the field-list visible at the call site and makes #13 (the "forgot a new score field" risk) self-evident in code review.

The alternative — keep the helper but drive its body from an explicit `MinerEvaluation.SCORE_BEARING_FIELDS` tuple — is mentioned in #13.

## Motivation

PR #784 already touched this area to address the EMA bleed; the function's field list is the kind of code that bit-rots silently.
Either inline or table-drive, but pick one.

bitcompass:
## Summary

Inlines `_zero_for_duplicate_penalty` into its single call site in `detect_and_penalize_miners_sharing_github`
([gittensor/validator/oss_contributions/inspections.py][1]) and removes the helper.

The helper existed for readability, but it had exactly one caller and the field-list lived behind a function jump. The function's own docstring already warned: "when adding new score-bearing fields to MinerEvaluation, update this function so the penalty stays comprehensive — a forgotten field would let a penalized miner retain partial credit in that dimension." Inlining makes that risk self-evident at code review time: adding a new score field to `MinerEvaluation` now lands next to the zeroing block in the same diff hunk.

The reminder comment is preserved next to the inlined block.

No behavioural change. Same fields zeroed in the same order.

[1]: gittensor/validator/oss_contributions/inspections.py

## Related Issues

Fixes #869

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [ ] Tests added/updated
- [x] Manually tested

The existing test `tests/validator/test_duplicate_penalty_propagation.py` exercises `detect_and_penalize_miners_sharing_github` end-to-end and covers the zeroing block via its public entry point. No new tests are
needed because the refactor is mechanical (function body moved verbatim into the loop body); test surface and assertions are
unchanged.

Verified locally:

```bash
uv run pre-commit run --all-files
uv run pre-commit run --all-files --hook-stage pre-push
uv run pytest tests/validator/test_duplicate_penalty_propagation.py -v


anderdc:
This doesn't meaningfully improve the repository. Closing.

PR #874, opened 2026-04-30
fix: detect PascalCase compiled-language tests and root-level conftest.py

### LINKED ISSUE #865: [Bug] is_test_file misclassifies pytest conftest.py outside tests/ directories
## Description

\`is_test_file\` in [gittensor/classes.py:76-103](https://github.com/entrius/gittensor/blob/test/gittensor/classes.py#L76-L103) detects \`conftest.py\` only when it lives under a path that already matches \`(^|/)tests?/\`. pytest auto-discovers \`conftest.py\` at any directory level — root, package directories, sub-packages — and a \`conftest.py\` is purely test infrastructure (fixtures, hooks, plugins) regardless of where it sits. Today every \`conftest.py\` outside a \`tests/\` directory earns full source-code score.

Verified directly:

\`\`\`python
>>> from gittensor.classes import FileChange
>>> for fname in ['conftest.py', 'tests/conftest.py', 'src/conftest.py', 'pkg/sub/conftest.py']:
...     fc = FileChange(pr_number=1, repository_full_name='a/b', filename=fname,
...                     changes=0, additions=0, deletions=0, status='modified')
...     print(fname, fc.is_test_file())
conftest.py            False    ← should be True
tests/conftest.py      True
src/conftest.py        False    ← should be True
pkg/sub/conftest.py    False    ← should be True
\`\`\`

## Steps to Reproduce

1. Submit a merged PR that touches a \`conftest.py\` at the repository root (or under \`src/\`, \`pkg/\`, etc. — anywhere except a \`tests/\` directory).
2. Note that the file is scored as ordinary source code instead of test infrastructure.

## Expected Behavior

Any file whose basename equals \`conftest.py\` is recognised as a test file. This is independent of directory: pytest treats every \`conftest.py\` on the discovery path as test infrastructure.

## Actual Behavior

\`is_test_file\` returns \`False\` for \`conftest.py\` unless the filename also matches the \`(^|/)tests?/\` directory pattern. Two basename-level patterns (\`r'^conftest\\.py$'\` would suffice) close the gap.

## Impact

A miner whose PR touches multiple \`conftest.py\` files outside dedicated test directories accumulates undeserved \`base_score\` from test-infrastructure changes. Same bug class as #765 (Android \`androidTest/\` + \`*IT.java\`/\`*IT.kt\`) and #790 (RSpec \`*_spec.<ext>\` + \`spec/\` paths) — both narrow ecosystem gaps fixed by extending the existing pattern lists.

## Suggested fix

Add \`r'^conftest\\.py$'\` to the basename \`test_patterns\` list at \`gittensor/classes.py:90\` and a regression test alongside the existing Android/Ruby cases in \`tests/test_classes.py\`. The fix is a 1-line regex add plus 1 test case.

### LINKED ISSUE #866: [Bug] is_test_file misses Swift/Kotlin/Java/C# PascalCase test files (FooTests.swift, FooTest.kt, TestFoo.java)
## Description

\`is_test_file\` in [gittensor/classes.py:76-103](https://github.com/entrius/gittensor/blob/test/gittensor/classes.py#L76-L103) only matches snake-case test naming (\`_test.<ext>\`, \`_tests.<ext>\`, \`test_…\`, \`spec_…\`, \`.test.<ext>\`, …). The PascalCase / camelCase conventions used by XCTest (Swift), JUnit/Kotest (Kotlin), JUnit (Java), and xUnit (C#) are missed:

\`\`\`python
>>> from gittensor.classes import FileChange
>>> for fname in ['MyClassTests.swift', 'FooTest.swift',
...               'FooTest.kt', 'BarTests.kt',
...               'TestFoo.java', 'fooTest.cs']:
...     fc = FileChange(pr_number=1, repository_full_name='a/b', filename=fname,
...                     changes=0, additions=0, deletions=0, status='modified')
...     print(fname, fc.is_test_file())
MyClassTests.swift   False    ← XCTest convention
FooTest.swift        False    ← XCTest convention
FooTest.kt           False    ← JUnit5/Kotest convention
BarTests.kt          False    ← Kotest convention
TestFoo.java         False    ← JUnit prefix convention
fooTest.cs           False    ← xUnit/.NET convention
\`\`\`

\`is_test_file\` correctly catches these only when they live under a \`tests?/\` or \`spec/\` directory. Modular-monolith projects that co-locate the test next to the source file (a common Swift/Kotlin pattern when there's no separate \`Tests/\` target) silently leak full \`base_score\` for test-infrastructure changes.

## Steps to Reproduce

1. Submit a merged PR that touches a Swift / Kotlin / Java test file co-located with source (no \`Tests/\` or \`tests/\` directory in the path), e.g. \`MyClass.swift\` and \`MyClassTests.swift\` in the same folder.
2. Note that the test file is scored as ordinary source code.

## Expected Behavior

Files whose basename matches the \`Test/Tests\` PascalCase or \`testFoo\`/\`TestFoo\` prefix convention are recognised as test files independent of directory.

## Actual Behavior

The basename pattern list at \`classes.py:90\` only checks snake-case + dot-separated test conventions. PascalCase and camelCase are absent.

## Impact

Same class as #765 (Android \`androidTest/\` + \`*IT.java\`/\`*IT.kt\`) and #790 (RSpec \`*_spec.<ext>\` + \`spec/\` paths) — narrow ecosystem gaps. A miner whose merged PR touches \`MyClassTests.swift\` next to \`MyClass.swift\` gets credit they shouldn't, and the issue affects four mainstream ecosystems.

## Suggested fix

Add basename patterns to the list at \`classes.py:90\` covering the four conventions:

- \`r'tests?\\.[^.]+$'\` — already present in lowercased form for snake-case but doesn't fire for \`footests.swift\` (because there's no underscore or dot delimiter); a more permissive variant \`r'(?<![a-z_.])tests?\\.[^.]+$'\` matches \`footest.swift\` while keeping \`mytestclass.py\` out.
- Or, more conservatively, language-targeted: \`r'tests?\\.(swift|kt|java|cs|fs|vb|scala)$'\`.
- For the Java/C# \`TestFoo\` prefix convention: \`r'^test[a-z][^.]*\\.(java|cs|scala|kt)$'\`.

Pick whichever shape fits the existing test_patterns list. Add regression cases in \`tests/test_classes.py\` alongside the Android/Ruby cases for each of the four ecosystems.

polyjeff529:
## Summary

`FileChange.is_test_file()` missed two real-world test naming conventions, so legitimate test files were being scored as production source.

### #866 — PascalCase compiled-language tests
JUnit / Kotlin / MSTest / XCTest / ScalaTest commonly use `<ClassName>Test.<ext>` or `<ClassName>Tests.<ext>`. Examples that previously slipped through:
- `app/src/main/java/com/example/AccountServiceTest.java`
- `app/src/main/kotlin/com/example/UserRepositoryTests.kt`
- `MyApp/Tests/AuthFlowTests.swift`
- `src/MyProject.Tests/CheckoutTests.cs`
- `core/src/main/scala/com/example/PaymentTest.scala`

The existing `_test\.[ext]$` patterns only fire on snake_case, so `Foo_test.kt` matched but `FooTest.kt` did not.

The matcher needs to distinguish `FooTest.kt` (test) from `Latest.kt` / `Manifest.java` (not test). The capital `T` is the disambiguator, but the existing implementation lowercases the basename before matching, which erases it. The new pattern uses the original-case basename and requires `[A-Z][A-Za-z0-9]*Tests?\.(swift|kt|java|cs|scala)$`.

### #865 — root-level / package-level `conftest.py`
pytest's fixture file is treated as a test by pytest regardless of where it lives in the tree. Previously only `conftest.py` files inside a `tests/` directory were classified as tests; ones at the project root or alongside a package (a common pytest layout) were scored as production code.

Added an early return on `basename == 'conftest.py'`.

## Tests

Three new parametrized blocks in `tests/test_classes.py`:
- `test_is_test_file_detects_pascalcase_compiled_language_tests` — 9 cases across Java / Kotlin / Swift / C# / Scala
- `test_is_test_file_rejects_pascalcase_lookalikes` — guards against `Latest.kt`, `Manifest.java`, `latest.md`, `contest.css`
- `test_is_test_file_detects_conftest_at_any_depth` — root, `tests/`, package-level, deeply nested

Existing detection cases (gradle source sets, rspec, non-test lookalikes, snake_case `_test`/`_spec`/`.test`) all still pass.

## Test plan

- [x] `pytest tests/test_classes.py` — 42 passed
- [x] `ruff check gittensor/classes.py tests/test_classes.py` — clean
- [x] `ruff format --check` — clean

Closes #866
Closes #865

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -76,6 +76,11 @@ class FileChange:
     def is_test_file(self) -> bool:
         filename_lower = self.filename.lower()
         basename = filename_lower.split('/')[-1]
+        basename_original = self.filename.split('/')[-1]
+
+        # pytest fixture file — applies project-wide regardless of directory
+        if basename == 'conftest.py':
+            return True
 
         test_dir_patterns = [
             r'(^|/)tests?/',
@@ -100,7 +105,14 @@ class FileChange:
             r'^tests\.[^.]+$',
         ]
 
-        return any(re.search(pattern, basename) for pattern in test_patterns)
+        if any(re.search(pattern, basename) for pattern in test_patterns):
+            return True
+
+        # PascalCase test conventions in compiled languages: `FooTest.kt`,
+        # `FooTests.swift`, `AccountServiceTests.cs`. The capital `T` boundary
+        # is what lets us distinguish these from incidental endings like
+        # `Latest.kt`, so we match against the original-case basename.
+        return bool(re.search(r'[A-Z][A-Za-z0-9]*Tests?\.(swift|kt|java|cs|scala)$', basename_original))
 
     @classmethod
     def from_github_response(cls, pr_number: int, repository_full_name: str, file_diff: DefaultDict) -> 'FileChange':
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -76,11 +76,6 @@ class FileChange:
     def is_test_file(self) -> bool:
         filename_lower = self.filename.lower()
         basename = filename_lower.split('/')[-1]
-        basename_original = self.filename.split('/')[-1]
-
-        # pytest fixture file — applies project-wide regardless of directory
-        if basename == 'conftest.py':
-            return True
 
         test_dir_patterns = [
             r'(^|/)tests?/',
@@ -88,6 +83,7 @@ class FileChange:
             r'(^|/)androidtest[a-z]*/',
             r'(^|/)integrationtest/',
             r'(^|/)spec/',
+            r'\.tests?/',  # .NET MyProject.Tests/FooTests.cs
         ]
         if any(re.search(pattern, filename_lower) for pattern in test_dir_patterns):
             return True
@@ -103,16 +99,10 @@ class FileChange:
             r'\.spec\.[^.]+$',
             r'^test\.[^.]+$',
             r'^tests\.[^.]+$',
+            r'^conftest\.py$',
         ]
 
-        if any(re.search(pattern, basename) for pattern in test_patterns):
-            return True
-
-        # PascalCase test conventions in compiled languages: `FooTest.kt`,
-        # `FooTests.swift`, `AccountServiceTests.cs`. The capital `T` boundary
-        # is what lets us distinguish these from incidental endings like
-        # `Latest.kt`, so we match against the original-case basename.
-        return bool(re.search(r'[A-Z][A-Za-z0-9]*Tests?\.(swift|kt|java|cs|scala)$', basename_original))
+        return any(re.search(pattern, basename) for pattern in test_patterns)
 
     @classmethod
     def from_github_response(cls, pr_number: int, repository_full_name: str, file_diff: DefaultDict) -> 'FileChange':
--- a/tests/test_classes.py
+++ b/tests/test_classes.py
@@ -75,32 +75,12 @@ def test_is_test_file_preserves_existing_test_conventions():
 @pytest.mark.parametrize(
     'filename',
     [
-        'app/src/main/java/com/example/FooTest.java',
-        'app/src/main/java/com/example/FooTests.java',
-        'app/src/main/kotlin/com/example/FooTest.kt',
-        'app/src/main/kotlin/com/example/FooTests.kt',
-        'MyApp/Tests/FooTests.swift',
-        'MyApp/Tests/FooTest.swift',
         'src/MyProject.Tests/AccountServiceTests.cs',
         'src/MyProject.Tests/AccountServiceTest.cs',
-        'core/src/main/scala/com/example/FooTest.scala',
     ],
 )
-def test_is_test_file_detects_pascalcase_compiled_language_tests(filename):
+def test_is_test_file_detects_dotnet_dotted_tests_directory(filename):
     assert _file_change(filename).is_test_file() is True
-
-
-@pytest.mark.parametrize(
-    'filename',
-    [
-        'docs/latest.md',
-        'src/main/kotlin/com/example/Latest.kt',
-        'app/src/main/java/com/example/Manifest.java',
-        'src/styles/contest.css',
-    ],
-)
-def test_is_test_file_rejects_pascalcase_lookalikes(filename):
-    assert _file_change(filename).is_test_file() is False
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


anderdc (changes requested):
Scope this down to two pattern-list additions:

```python
test_dir_patterns = [
    ...
    r'\.tests?/',           # .NET MyProject.Tests/FooTests.cs
]
test_patterns = [
    ...
    r'^conftest\.py$',
]
```

Drop the `basename_original` variable and the `[A-Z][A-Za-z0-9]*Tests?\.(swift|kt|java|cs|scala)$` regex. The `\.tests?/` directory pattern covers the meaningful #866 case (.NET test-project layout — `MyProject.Tests/FooTests.cs` doesn't match the existing `(^|/)tests?/` because the segment is `MyProject.Tests`, not `tests`). The remaining co-located `FooTest.kt` layout the basename regex catches is rare enough not to justify a second case-sensitivity rule in the function — and the leading `[A-Z]` requirement excludes `fooTest.cs`, which #866 explicitly lists as a missed case.

Tests can shrink to match: keep conftest-at-any-depth and `.Tests/` directory cases; the co-located PascalCase fixtures aren't needed.

polyjeff529:
> Scope this down to two pattern-list additions:
> 
> ```python
> test_dir_patterns = [
>     ...
>     r'\.tests?/',           # .NET MyProject.Tests/FooTests.cs
> ]
> test_patterns = [
>     ...
>     r'^conftest\.py$',
> ]
> ```
> 
> Drop the `basename_original` variable and the `[A-Z][A-Za-z0-9]*Tests?\.(swift|kt|java|cs|scala)$` regex. The `\.tests?/` directory pattern covers the meaningful #866 case (.NET test-project layout — `MyProject.Tests/FooTests.cs` doesn't match the existing `(^|/)tests?/` because the segment is `MyProject.Tests`, not `tests`). The remaining co-located `FooTest.kt` layout the basename regex catches is rare enough not to justify a second case-sensitivity rule in the function — and the leading `[A-Z]` requirement excludes `fooTest.cs`, which #866 explicitly lists as a missed case.
> 
> Tests can shrink to match: keep conftest-at-any-depth and `.Tests/` directory cases; the co-located PascalCase fixtures aren't needed.

I updated, Could you check it?

anderdc (approved):

PR #875, opened 2026-04-30
fix: isolate malformed master_repositories entries instead of collapsing the loader

### LINKED ISSUE #750: [Bug] load_master_repo_weights collapses to {} when any entry is not a dict
## Summary

`load_master_repo_weights` ([`gittensor/validator/utils/load_weights.py:93-138`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/utils/load_weights.py#L93)) assumes every value in `master_repositories.json` is a dict and accesses it via `metadata.get('weight', 0.01)`. The per-entry `try/except` only catches `(ValueError, TypeError)`, so an `AttributeError` raised by a non-dict entry escapes to the function-level `except Exception` (line 136) and the whole loader returns `{}`.

A `{}` return causes `load_miners_prs` ([`github_api_tools.py:986-988`](https://github.com/entrius/gittensor/blob/test/gittensor/utils/github_api_tools.py#L986)) to skip every PR as `"... not in master repositories"`. Every miner in the round scores zero. The only signal is one log line: `Unexpected error loading repository weights: 'float' object has no attribute 'get'`.

The companion loader `load_programming_language_weights` (lines 141-185) does guard with `if isinstance(config, dict)` and supports plain-float back-compat (line 169). The repo loader does not.

## Steps to reproduce

1. Edit `gittensor/validator/weights/master_repositories.json` and insert one non-dict entry, e.g. `"owner/repo": 0.1`.
2. Call `load_master_repo_weights()` from a REPL or run the validator forward pass.
3. Function returns `{}`. Logs contain a single `Unexpected error loading repository weights` line referencing `AttributeError`.
4. The next scoring round logs `Skipping PR #N in <repo> - ineligible repo` for every PR and produces an all-zero `oss_rewards` array.

## Expected

A malformed single entry is skipped with a per-entry warning. The remaining well-formed entries load. Behavior matches `load_programming_language_weights`.

## Actual

One bad entry collapses the entire repo weights map. The validator round silently produces zero OSS scores with no warning that names the offending repo.

polyjeff529:
## Summary

A single malformed entry in `master_repositories.json` zeroes out the entire scoring round.

The per-entry `try/except` in `load_master_repo_weights` only catches `(ValueError, TypeError)`, so an `AttributeError` raised by `metadata.get('weight', 0.01)` against a non-dict entry escapes to the function-level `except Exception:` and the loader returns `{}`. `load_miners_prs` then logs `Skipping PR #N in <repo> - ineligible repo` for *every* PR; `oss_rewards` ends up all zero. The only signal is one log line — `Unexpected error loading repository weights: 'float' object has no attribute 'get'`.

The sibling loader `load_programming_language_weights` already gets this right (`isinstance(config, dict)` guard with a plain-float back-compat path). This PR mirrors that pattern in the repo loader.

## What changes

- `isinstance(metadata, dict)` guard around the dict-form parsing.
- Plain-float back-compat (`'owner/repo': 0.1` → `RepositoryConfig(weight=0.1, …defaults)`), matching `load_programming_language_weights`.
- The per-entry `except` now skips the malformed entry with a warning that includes the offending value, instead of silently retrying `metadata.get('weight', 0.01)` — which itself `AttributeError`'d on the non-dict path that triggered this in the first place.
- The function-level `except Exception` block stays as the JSON-decode / file-not-found / truly-unexpected catch-all; it's no longer reached on a per-entry typo.

No behavior change for well-formed JSON. The live `master_repositories.json` continues to load identically (verified by the existing `test_master_repositories_not_empty` and `test_mirror_enabled_field_present_on_live_configs` tests).

## Tests

New `TestLoadMasterRepoWeightsMalformedEntries` class in `tests/validator/test_load_weights.py` — 5 cases:

- `test_non_dict_entry_isolated_other_entries_load` — list entry skipped, dict siblings load.
- `test_plain_float_entry_back_compat` — plain-float weight loads with default `mirror_enabled=False` / `inactive_at=None`.
- `test_unparseable_weight_skipped_others_survive` — `{'weight': 'not-a-number'}` is skipped, not promoted to default.
- `test_string_entry_isolated` — string entry skipped cleanly.
- `test_only_malformed_entries_returns_empty_dict_not_outer_except` — regression: an all-malformed file still returns `{}`, but via the per-entry skip path, not the outer `except Exception`.

All 33 tests in `tests/validator/test_load_weights.py` pass; ruff check + format clean.

## Test plan

- [x] `pytest tests/validator/test_load_weights.py` — 33 passed (5 new)
- [x] `ruff check gittensor/validator/utils/load_weights.py tests/validator/test_load_weights.py` — clean
- [x] `ruff format --check` — clean
- [x] Live `master_repositories.json` continues to load with the same entry count (existing `test_master_repositories_not_empty` is the regression sentinel)

Closes #750

### DIFF
--- a/gittensor/validator/utils/load_weights.py
+++ b/gittensor/validator/utils/load_weights.py
@@ -113,21 +113,29 @@ def load_master_repo_weights() -> Dict[str, RepositoryConfig]:
             bt.logging.error(f'Expected dict from {weights_file}, got {type(data)}')
             return {}
 
-        # Parse JSON data into RepositoryConfig objects
+        # Parse JSON data into RepositoryConfig objects.
+        # Companion loader load_programming_language_weights() also accepts a
+        # plain-float weight for back-compat; mirror that here so a malformed
+        # entry (or one written in the legacy short form) is isolated rather
+        # than collapsing the whole loader to {}.
         normalized_data: Dict[str, RepositoryConfig] = {}
         for repo_name, metadata in data.items():
             try:
-                config = RepositoryConfig(
-                    weight=float(metadata.get('weight', 0.01)),
-                    inactive_at=metadata.get('inactive_at'),
-                    additional_acceptable_branches=metadata.get('additional_acceptable_branches'),
-                    mirror_enabled=bool(metadata.get('mirror_enabled', False)),
-                )
+                if isinstance(metadata, dict):
+                    config = RepositoryConfig(
+                        weight=float(metadata.get('weight', 0.01)),
+                        inactive_at=metadata.get('inactive_at'),
+                        additional_acceptable_branches=metadata.get('additional_acceptable_branches'),
+                        mirror_enabled=bool(metadata.get('mirror_enabled', False)),
+                    )
+                else:
+                    # Plain-float weight: matches the back-compat path in
+                    # load_programming_language_weights().
+                    config = RepositoryConfig(weight=float(metadata))
                 normalized_data[repo_name.lower()] = config
             except (ValueError, TypeError) as e:
-                bt.logging.warning(f'Could not parse config for {repo_name}: {e}, using defaults')
-                # Create config with defaults if parsing fails
-                normalized_data[repo_name.lower()] = RepositoryConfig(weight=float(metadata.get('weight', 0.01)))
+                bt.logging.warning(f'Could not parse config for {repo_name}: {metadata!r} - {e}; skipping entry')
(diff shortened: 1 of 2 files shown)


anderdc:
PR not needed — overly defensive. Existing tests catch this preflight (`test_master_repositories_not_empty` fails CI if the loader collapses to `{}`).

PR #880, opened 2026-04-30
fix: signal PR files fetch failures so miners aren't silently zero-scored

### LINKED ISSUE #879: [Bug] `get_pull_request_file_changes` returns `[]` on retry exhaustion, indistinguishable from "PR has no files"
## Description

`gittensor/utils/github_api_tools.py:361` — when the REST `/repos/{repo}/pulls/{n}/files` request fails on every retry attempt, the function returns `[]`. The same return value is used when the PR genuinely has zero file changes. Callers in `oss_contributions/scoring.py:90` (`score_pull_request`) treat `[]` as "no file changes found, skip" and silently award `base_score=0`, so a transient GitHub 5xx is indistinguishable from a real empty PR.

The function's docstring says "returns `Optional[List[FileChange]]`", but the implementation never actually returns `None` on the failure path — every exit returns either a populated list or `[]`. The type hint is therefore a lie.

## Steps to Reproduce

1. Pick any merged PR that should produce a non-zero score (e.g. one with  ~10 file changes). Run a normal scoring round and confirm the  `base_score` is non-zero.
2. Patch `requests.get` (or use a network-fault injector / `httpretty`) to  return HTTP 502 for every call to `/repos/.../pulls/.../files` for the  duration of one round.
3. Re-run scoring for that miner.

## Expected Behavior

The function should return `None` (matching its declared `Optional` return
type) when all retries fail. The scoring layer should detect `None` and
either skip scoring this PR (preserving the prior cached score) or mark
`miner_eval.github_pr_fetch_failed = True` so the round is recognised as
incomplete.

## Actual Behavior

`get_pull_request_file_changes` returns `[]`. `score_pull_request` logs
`"No file changes found."`, returns without setting `pr.base_score`, and
the PR contributes 0 to the miner's `total_score`. The miner is silently
penalised for a GitHub-side outage.

## Environment

- OS: Linux (any)
- Python version: 3.10+
- Commit/Version: `d8ba803` (refactor: drop get_contract_address re-export)

## Additional Context

- Failure-handling parity exists on the GraphQL path (`load_miners_prs` sets `github_pr_fetch_failed=True`), but the REST path doesn't have it.
- Related issue: `MinerEvaluationCache` cache fallback (PR #531/IS #515) already exists for the GraphQL path; this is the missing REST companion.
- Suggested patch shape: `return None` from the failure branch and update the `if not file_changes:` check at scoring.py:91 to `if file_changes is None:`
  → log + skip; `if not file_changes:` (i.e. `[]`) → log + accept zero base
  score.

since-2017-hub:
## Summary

`get_pull_request_file_changes` was declared as `Optional[List[FileChange]]` but its retry-exhausted failure path returned
`[]`, identical to a PR that genuinely has zero file changes. The caller in `score_pull_request` did `if not file_changes:` and skipped, leaving `pr.base_score` at 0 — so a transient GitHub 5xx during a scoring round was indistinguishable from a real empty PR and silently penalised the miner.

This PR makes the failure path honest:

- `gittensor/utils/github_api_tools.py`: return `None` on retry   exhaustion (matches the docstring and declared return type).
- `gittensor/validator/oss_contributions/scoring.py`: distinguish  `file_changes is None` (fetch failed → set
  `miner_eval.github_pr_fetch_failed = True` and skip, so the existing `should_use_cache_fallback` path in `validator.py` engages) from `not file_changes` (legit empty PR → skip without flagging).
- Tests updated to assert `result is None` on retry exhaustion.

This brings the REST file-changes path to parity with the GraphQL path `load_miners_prs`), which already sets `github_pr_fetch_failed = True` on fetch failure.

## Related Issues

Fixes #879 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Updated three existing tests in `tests/utils/test_github_api_tools.py` that asserted `result == []` on retry exhaustion to assert
`result is None` instead:

- `TestFileChangesRetryLogic::test_gives_up_after_three_attempts`
- `TestFileChangesRetryLogic::test_gives_up_after_three_connection_errors`
- `TestFileChangesPagination::test_all_pages_fail_returns_none` (renamed
  from `..._returns_empty_list`)

The `test_retry_on_connection_error_then_success` case still asserts `[]` because in that scenario the retry succeeds with a legitimately empty list of files — which is the correct, distinct return value post-fix.

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -83,6 +83,7 @@ class FileChange:
             r'(^|/)androidtest[a-z]*/',
             r'(^|/)integrationtest/',
             r'(^|/)spec/',
+            r'\.tests?/',  # .NET MyProject.Tests/FooTests.cs
         ]
         if any(re.search(pattern, filename_lower) for pattern in test_dir_patterns):
             return True
@@ -98,6 +99,7 @@ class FileChange:
             r'\.spec\.[^.]+$',
             r'^test\.[^.]+$',
             r'^tests\.[^.]+$',
+            r'^conftest\.py$',
         ]
 
         return any(re.search(pattern, basename) for pattern in test_patterns)
@@ -305,12 +307,8 @@ class PullRequest:
         last_edited_at = parse_github_timestamp_to_cst(raw_edited_at) if isinstance(raw_edited_at, str) else None
         merged_at = parse_github_timestamp_to_cst(pr_data['mergedAt']) if is_merged else None
 
-        changes_requested_count = 0
-        if is_merged:
-            cr_reviews = pr_data.get('changesRequestedReviews', {}).get('nodes', [])
-            changes_requested_count = sum(
-                1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS
-            )
+        cr_reviews = (pr_data.get('changesRequestedReviews') or {}).get('nodes') or []
+        changes_requested_count = sum(1 for r in cr_reviews if r.get('authorAssociation') in MAINTAINER_ASSOCIATIONS)
 
         current = {(n.get('name') or '').lower() for n in (pr_data.get('labels') or {}).get('nodes') or [] if n}
         label: Optional[str] = None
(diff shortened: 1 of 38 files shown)


anderdc (changes requested):
Just make this PR a one-liner to adjust the response type to not have `Optional`

anderdc (approved):

PR #891, opened 2026-04-30
perf(storage): bump execute_values page_size to reduce PG round-trips

### LINKED ISSUE #890: [Perf] BULK_UPSERT page_size=100 splits large bulk inserts into 10× more PG round-trips than needed
## Description

`Repository.store_pull_requests_bulk`, `store_issues_bulk`, and `store_file_changes_bulk` in [`gittensor/validator/storage/repository.py:212,267,316`](gittensor/validator/storage/repository.py#L212) all call psycopg2's `execute_values(..., page_size=100)`. Each `page_size` chunk is one PostgreSQL round-trip — a separate `INSERT ... VALUES (...) ... ON CONFLICT ...` statement built from up to 100 rows.

For a typical full-miner-set scoring round, a single validator can produce:
- **PRs:** ~200 miners × ~50 PRs (lookback window) = ~10,000 PR upserts
- **Issues:** comparable order via `closingIssuesReferences`
- **file_changes:** ~10,000 PRs × ~10-30 files each = 100,000–300,000 file_change upserts

At `page_size=100` that's ~100 round-trips for PRs/issues each and ~1000–3000 round-trips for file_changes per round. On a managed PostgreSQL where each round-trip costs ~1-3 ms of network latency, file_changes ingestion alone can spend 5-10s+ on round-trip overhead — pure waste, since each row's payload is tiny relative to the protocol cost.

`execute_values` is specifically designed to avoid this by inlining all rows in `page_size` chunks into a single SQL statement; the `page_size=100` default in this code is the lower end of what's reasonable.

## Why not just bump every call site to `page_size=1000`?

The `file_changes` path is different from PRs/issues because every row carries the `FileChange.patch` column — the raw diff text from GitHub. Patches are variable-size (a few hundred bytes to several MB; only **source-file fetches** are capped at `MAX_FILE_SIZE_BYTES = 1_000_000`, the diff itself isn't). At `page_size=1000` for file_changes:

- Typical case (5 KB avg patch): ~5 MB query string per batch — fine.
- Pathological case (large patches): tens of MB query string in a single transient Python `str` — visible memory pressure on validator hosts and risk of `statement_timeout` on managed PG with short defaults.

PRs and issues have bounded per-row sizes (titles, names, datetimes, score scalars) and are safe at `page_size=1000`.

## Fix

Three-line change in [`repository.py`](gittensor/validator/storage/repository.py):

- `store_pull_requests_bulk` — `page_size=100` → `page_size=1000`
- `store_issues_bulk` — `page_size=100` → `page_size=1000`
- `store_file_changes_bulk` — `page_size=100` → `page_size=200` (kept lower because of the unbounded `patch` column; comment added explaining the asymmetry)

```python
# file_changes carries the variable-size patch column, so we keep
# page_size lower than the PR/issue bulk paths to bound the per-batch
# SQL string size when patches are large.
execute_values(..., page_size=200)
```

## Impact

- **MEDIUM-HIGH** for PRs/issues: ~10× fewer round-trips → roughly 1-3s saved per round on PR+issue ingestion at typical PG network latencies.
- **MEDIUM** for file_changes: 2× fewer round-trips with the conservative bump.
- No semantic change — same SQL, same data, same `ON CONFLICT` behavior; only the chunking changes.

## Coordination note

Open PR #778 (atomic transactions) modifies the surrounding `commit()`/`rollback()` lines in these same methods. The two changes touch adjacent lines but don't conflict on the same edits — whichever lands first, the other rebases trivially.

## Related

- #522, #556 — added retry logic to these methods, did not touch `page_size`
- #445 — proposed extracting `_bulk_upsert` helper, did not touch `page_size`
- psycopg2 `execute_values` docs recommend `page_size` of "a few hundred or a thousand" for typical workloads

PfanP:
`Repository.store_pull_requests_bulk` and `store_issues_bulk` used `execute_values(..., page_size=100)`. A 10k-row insert needs ~100 PG round-trips at that setting; on managed PostgreSQL (~1-3 ms RTT) that's seconds of pure protocol overhead per scoring round.

Bump page_size to 1000 for the PR/issue paths, which carry only bounded-size scalar columns. Keep page_size=200 for store_file_changes_bulk because each row carries the raw `patch` column (variable, can be MB); 1000 there could build a multi-MB SQL string under heavy diffs.

No semantic change — same SQL, same data, same ON CONFLICT behavior; only the chunking changes.

Fixes #890

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -209,7 +209,7 @@ class Repository(BaseRepository):
                     BULK_UPSERT_PULL_REQUESTS.replace('VALUES %s', 'VALUES %s'),
                     values,
                     template=None,
-                    page_size=100,
+                    page_size=1000,
                 )
                 self.db.commit()
                 return len(values)
@@ -264,7 +264,7 @@ class Repository(BaseRepository):
                 from psycopg2.extras import execute_values
 
                 execute_values(
-                    cursor, BULK_UPSERT_ISSUES.replace('VALUES %s', 'VALUES %s'), values, template=None, page_size=100
+                    cursor, BULK_UPSERT_ISSUES.replace('VALUES %s', 'VALUES %s'), values, template=None, page_size=1000
                 )
                 self.db.commit()
                 return len(values)
@@ -308,12 +308,15 @@ class Repository(BaseRepository):
                 # Use psycopg2's execute_values for efficient bulk insert
                 from psycopg2.extras import execute_values
 
+                # file_changes carries the variable-size patch column, so we keep
+                # page_size lower than the PR/issue bulk paths to bound the per-batch
+                # SQL string size when patches are large.
                 execute_values(
                     cursor,
                     BULK_UPSERT_FILE_CHANGES.replace('VALUES %s', 'VALUES %s'),
                     values,
                     template=None,
-                    page_size=100,
+                    page_size=200,
                 )
                 self.db.commit()
                 return len(values)


anderdc:
Speculative — no measurement, no benchmark. Current implementation has not seen extreme latency or issues in production. Closing.

PR #894, opened 2026-04-30
fix: handle malformed mirror JSON responses

### LINKED ISSUE #893: [Bug] MirrorClient malformed 2xx JSON bypasses mirror failure handling and can stop validator rounds
## Description

`MirrorClient._get()` treats any HTTP 2xx response as successful and directly returns `response.json()`. If the mirror, CDN, or proxy returns a 2xx response with malformed JSON, truncated JSON, or an HTML/plaintext error body, `response.json()` raises `JSONDecodeError` / `ValueError` outside the client's retry and `MirrorRequestError` handling.

All validator mirror call sites treat `MirrorRequestError` as the recoverable mirror-failure signal. A raw JSON parse exception therefore bypasses the intended fallback/skip behavior and can abort the current validator scoring path instead of marking the mirror request as failed.

This is a practical reliability issue for validator operators: a single bad mirror response can interrupt OSS mirror scoring or issue discovery, bubble to the validator outer loop, and stop the validator thread.

## Steps to Reproduce

1. Create a `MirrorClient` with a mocked session that returns HTTP 200 but raises from `.json()`:

   ```python
   from unittest.mock import Mock

   from gittensor.utils.mirror.client import MirrorClient, MirrorRequestError

   response = Mock(status_code=200, text="<html>bad gateway</html>")
   response.json.side_effect = ValueError("Expecting value")

   session = Mock()
   session.get.return_value = response

   client = MirrorClient(session=session, max_attempts=1)
   client.get_miner_pulls("123")
   ```

2. Observe that the raw `ValueError` escapes instead of `MirrorRequestError`.

3. Use the same client shape through a validator call site, for example `load_mirror_miner_prs(...)`, `score_mirror_pr(...)`, or `run_mirror_issue_discovery(...)`.

4. Observe that those call sites only catch `MirrorRequestError`, so the malformed 2xx response escapes the normal mirror failure path.

## Expected Behavior

Malformed 2xx mirror bodies should be handled as mirror request failures:

- Retry with the same backoff policy used for transient mirror failures.
- After `max_attempts`, raise `MirrorRequestError`.
- Let existing validator callers continue using their current `MirrorRequestError` handling to skip the affected miner/PR or mark the mirror fetch as failed.

The validator should not crash or stop a scoring round because a mirror endpoint returned a malformed 2xx body.

## Actual Behavior

`MirrorClient._get()` returns `response.json()` directly for 2xx responses. JSON decode errors are not caught, retried, or converted to `MirrorRequestError`.

As a result:

- `load_mirror_miner_prs(...)` catches only `MirrorRequestError`, so malformed JSON from `get_miner_pulls(...)` escapes OSS scoring.
- `score_mirror_pr(...)` catches only `MirrorRequestError`, so malformed JSON from `get_pr_files(...)` escapes per-PR mirror scoring.
- `run_mirror_issue_discovery(...)` catches only `MirrorRequestError`, so malformed JSON from `get_miner_issues(...)` escapes issue discovery.
- The exception can bubble through `forward(...)` to the validator run loop's outer exception handler, which logs the error and exits the validator thread.

## Environment

- OS: Linux 6.8.0-90-generic x86_64 GNU/Linux
- Python version: Python 3.12.3
- Commit/Version: `34ffeb8c4b`

## Additional Context

Relevant code paths:

- `gittensor/utils/mirror/client.py:100-115`: network exceptions are caught, but `response.json()` for 2xx responses is outside that error handling.
- `gittensor/validator/oss_contributions/mirror/load.py:59-64`: mirror PR loading catches only `MirrorRequestError`.
- `gittensor/validator/oss_contributions/mirror/scoring.py:134-138`: mirror PR file scoring catches only `MirrorRequestError`.
- `gittensor/validator/issue_discovery/mirror_scan.py:150-155`: mirror issue fetching catches only `MirrorRequestError`.
- `gittensor/validator/issue_discovery/mirror_scan.py:401-409`: solving-PR file fetch catches only `MirrorRequestError`.
- `gittensor/validator/forward.py:137`: issue discovery is called directly from the scoring forward path.
- `neurons/base/validator.py:167-170`: uncaught forward exceptions are logged at the outer validator loop.
- `neurons/validator.py:200-202`: the main process exits once the validator thread is no longer alive.

Suggested fix:

- Catch JSON decode errors (`ValueError` is sufficient for `requests.Response.json()`) around `response.json()` in `MirrorClient._get()`.
- Treat invalid JSON as a retryable mirror failure, setting `last_error` with a short body preview.
- After all attempts fail, raise `MirrorRequestError`.
- Optionally wrap top-level mirror response schema parse failures in the endpoint methods as `MirrorRequestError` too, so a 2xx JSON error object cannot escape as `KeyError`.

Suggested tests:

- Add a `tests/utils/test_mirror_client.py` case where a 200 invalid-JSON response is followed by a valid response; assert the client retries and succeeds.
- Add a case where all attempts return 200 invalid JSON; assert `MirrorRequestError` is raised after `max_attempts`.
- Add a loader or issue-discovery smoke test proving malformed 2xx mirror output is handled through the existing mirror failure path rather than escaping as a raw exception.

Duplicate check:

- Searched upstream open issues/PRs for `MirrorClient invalid JSON response.json MirrorRequestError`, `invalid JSON`, `response.json`, and malformed mirror client wording.
- No open duplicate found for malformed 2xx JSON bypassing `MirrorRequestError`.
- Related but not duplicate: `#876` covers synchronous `MirrorClient` calls blocking the async validator path; it does not cover malformed 2xx response parsing or fallback behavior.

bitloi:
## Summary

Fix `MirrorClient` so malformed HTTP 2xx mirror responses are treated as mirror request failures instead of escaping as raw JSON parse exceptions.

Previously, `_get()` returned `response.json()` directly for any 2xx response. If the mirror, CDN, or proxy returned HTTP 200 with truncated JSON or an HTML/plaintext error body, `ValueError` could bypass retry handling and the `MirrorRequestError` path used by validator mirror callers.

This PR:
- retries invalid 2xx JSON with the existing mirror backoff policy
- raises `MirrorRequestError` after `max_attempts`
- wraps top-level mirror response schema parse failures as `MirrorRequestError`
- preserves existing validator fallback behavior for failed mirror fetches

## Related Issues

Closes #893

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

Commands run:

- `uv run pytest tests/utils/test_mirror_client.py -q`
- `uv run pytest tests/validator/oss_contributions/mirror/test_load.py -q`
- `uv run pytest tests/validator/issue_discovery/test_mirror_scan.py::TestRunMirrorIssueDiscovery::test_mirror_request_error_does_not_abort_other_miners tests/validator/issue_discovery/test_mirror_scan.py::TestSolvingPrCache::test_fetch_failure_skips_scoring_for_that_issue -q`
- `uv run pytest tests/ -q --ignore=tests/validator/issue_discovery/test_one_issue_per_pr_cross_miner.py --ignore=tests/validator/issue_discovery/test_post_merge_issue_edit.py`
- `uv run pyright`
- `uv run pre-commit run --files gittensor/utils/mirror/client.py tests/utils/test_mirror_client.py tests/validator/oss_contributions/mirror/test_load.py`

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/mirror/client.py
+++ b/gittensor/utils/mirror/client.py
@@ -134,7 +134,7 @@ class MirrorClient:
                 except ValueError as e:
                     last_error = f'invalid JSON: {e}; body={_body_preview(response)!r}'
                     if attempt < self.max_attempts - 1:
-                        backoff = min(5 * (2**attempt), 30)
+                        backoff = backoff_seconds(attempt)
                         bt.logging.warning(
                             f'Mirror GET {path} failed ({last_error}) '
                             f'(attempt {attempt + 1}/{self.max_attempts}), retrying in {backoff}s...'
--- a/gittensor/utils/mirror/client.py
+++ b/gittensor/utils/mirror/client.py
@@ -29,6 +29,13 @@ class MirrorRequestError(RuntimeError):
     exhausts all retries on transient failures."""
 
 
+def _body_preview(response: requests.Response) -> str:
+    text = getattr(response, 'text', '')
+    if not isinstance(text, str):
+        text = str(text)
+    return text[:200]
+
+
 class MirrorClient:
     """Client for https://mirror.gittensor.io scoring endpoints."""
 
@@ -65,7 +72,10 @@ class MirrorClient:
         path = f'/api/v1/miners/{github_id}/pulls'
         params = {'since': since.astimezone(timezone.utc).isoformat()} if since else None
         data = self._get(path, params=params)
-        return MirrorPullRequestsResponse.from_dict(data)
+        try:
+            return MirrorPullRequestsResponse.from_dict(data)
+        except Exception as e:
+            raise MirrorRequestError(f'Mirror GET {path} returned invalid mirror response: {e}') from e
 
     def get_miner_issues(
         self,
@@ -77,7 +87,10 @@ class MirrorClient:
         path = f'/api/v1/miners/{github_id}/issues'
         params = {'since': since.astimezone(timezone.utc).isoformat()} if since else None
         data = self._get(path, params=params)
-        return MirrorIssuesResponse.from_dict(data)
+        try:
+            return MirrorIssuesResponse.from_dict(data)
+        except Exception as e:
+            raise MirrorRequestError(f'Mirror GET {path} returned invalid mirror response: {e}') from e
 
     def get_pr_files(
         self,
@@ -91,7 +104,10 @@ class MirrorClient:
         """
         path = f'/api/v1/pulls/{repo_full_name}/{pr_number}/files'
         data = self._get(path)
-        return MirrorPullRequestFilesResponse.from_dict(data)
+        try:
+            return MirrorPullRequestFilesResponse.from_dict(data)
+        except Exception as e:
+            raise MirrorRequestError(f'Mirror GET {path} returned invalid mirror response: {e}') from e
 
     def _get(self, path: str, params: Optional[dict] = None) -> dict:
         url = f'{self.base_url}{path}'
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
`gittensor/utils/mirror/client.py:134` — use `backoff_seconds(attempt)` instead of inline `min(5 * (2**attempt), 30)`. It's already imported and used by the other two retry tails in this same function; the inline copy drifts.

bitloi:
> `gittensor/utils/mirror/client.py:134` — use `backoff_seconds(attempt)` instead of inline `min(5 * (2**attempt), 30)`. It's already imported and used by the other two retry tails in this same function; the inline copy drifts.

@anderdc I just pushed the update. Would you please check it again?

anderdc (approved):

PR #855, opened 2026-04-29
fix(cli): validate --id range on `gitt i list` to match sibling commands

plind-junior:
## Summary

`gitt i list --id <N>` accepted `0`, negative numbers, and out-of-u32-range values and passed them straight through to on-chain lookup, returning a misleading "Issue N not found on-chain" message. Every sibling command using the same `issue_id` semantics (`i submissions`, `vote solution`, `vote cancel`) already rejects these inputs via `validate_issue_id` and emits a structured `bad_parameter` error. `view.py:issues_list` was the only outlier.

Closes the gap left by **#210 / PR #211**, which prescribed this validation across the CLI but listed only `mutations.py / vote.py / admin.py` — `view.py` was outside that scope.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Reproduction (before this PR)

```bash
$ gitt i submissions --id 0
✗ id must be between 1 and 999999 (got 0)              # ✅ rejected at input

$ gitt vote solution 0 5Hxxx... 5Hyyy... 1
✗ id must be between 1 and 999999 (got 0)              # ✅ rejected at input

$ gitt i list --id 0
Network: finney • Contract: ...
✗ Issue 0 not found on-chain.                          # ❌ silently did an on-chain query

$ gitt i list --id -1
✗ Issue -1 not found on-chain.                         # ❌ negative passed through

$ gitt i list --id 99999999999999
✗ Issue 99999999999999 not found on-chain.             # ❌ way out of u32 range, also passed through
```

Verified on `upstream/test` HEAD `d8ba803`.

## After

```bash
$ gitt i list --id 0
✗ id must be between 1 and 999999 (got 0)              # ✅ matches submissions / vote

$ gitt i list --id 0 --json
{"success": false, "error": {"type": "bad_parameter", "message": "id must be between 1 and 999999 (got 0)"}}

$ gitt i list --id 1                                   # valid IDs unaffected
Network: finney • Contract: ...
✗ Issue 1 not found on-chain.                          # PR #335's structured not-found path preserved

$ gitt i list                                          # no --id, list-all path unchanged
```

## Why this isn't a duplicate

- **#210 (CLOSED COMPLETED)** prescribed the rule: *"`1 <= value < 1_000_000` on all commands that accept an issue ID or number."* The issue's file list named only `mutations.py, vote.py, admin.py`, so PR #211 honored that scope and didn't touch `view.py`. This PR closes the leftover gap.
[...]

PR #654, opened 2026-04-21
fix: preserve repo weight precision by removing 2-decimal rounding

bitloi:
## Summary

Drop `round(repo_config.weight, 2)` at both multiplier assignment sites. Extract [resolve_repo_weight(repo_config)](cci:1://file:///root/74/gold/gittensor/gittensor/validator/utils/load_weights.py:44:0-48:29) helper in [load_weights.py](cci:7://file:///root/74/gold/gittensor/gittensor/validator/utils/load_weights.py:0:0-0:0) so the `None → DEFAULT_REPO_WEIGHT` fallback lives in one place. Preserves full 4-decimal precision from [master_repositories.json](cci:7://file:///root/74/gold/gittensor/gittensor/validator/weights/master_repositories.json:0:0-0:0) through the multiplier chain into `earned_score`.

## Related Issues

Fixes #652

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added [TestResolveRepoWeight](cci:2://file:///root/74/gold/gittensor/tests/validator/test_load_weights.py:163:0-183:96) in [tests/validator/test_load_weights.py](cci:7://file:///root/74/gold/gittensor/tests/validator/test_load_weights.py:0:0-0:0) covering `None` default, full-precision preservation for six real weights, and live-data verification against [master_repositories.json](cci:7://file:///root/74/gold/gittensor/gittensor/validator/weights/master_repositories.json:0:0-0:0) (`antoniorodr/cronboard` 0.0349, `junegunn/fzf` 0.0351).

`uv run pytest tests/` → 60 passed. `uv run ruff check` / `ruff format --check` clean.

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

Issue #652, opened 2026-04-21
[BUG] repo_weight rounded to 2 decimals discards up to 14% of the JSON precision


carlos4s:
## Description

`calculate_pr_multipliers` (`gittensor/validator/oss_contributions/scoring.py:232`) and `_collect_issues_from_prs` (`gittensor/validator/issue_discovery/scoring.py:280`) apply `round(repo_config.weight, 2)` before assigning the repo weight as a multiplier. Weights in `master_repositories.json` are 4-decimal and most sit in `[0.03, 0.15]` — where the 2nd decimal is the most significant digit — so rounding there distorts the multiplier by up to 14% per repo and flows into `earned_score`.

## Steps to Reproduce

1. Run:
   ```python
   from gittensor.validator.utils.load_weights import load_master_repo_weights
   repos = load_master_repo_weights()
   w = repos['antoniorodr/cronboard'].weight  # 0.0349
   print(w, round(w, 2))  # 0.0349 0.03 — 14% loss
   ```
2. Trace through the multiplier chain in `calculate_pr_multipliers` — the rounded 0.03 is what reaches `earned_score`.

## Expected Behavior

`repo_weight_multiplier` preserves the full precision from `master_repositories.json`.

## Actual Behavior

75 of 214 tracked repos distort by >5%, 20 by >10%, worst case −14% (`antoniorodr/cronboard` 0.0349 → 0.03). Two repos with nearly identical weights (`cronboard` 0.0349 vs `fzf` 0.0351) land on opposite sides of the rounding boundary, creating a 33% reward spread for identical work.

## Environment

- OS: Linux
- Python version: 3.12
- Commit/Version: `test` branch @ `bad35f7`

## Additional Context

`issue_multiplier`, `label_multiplier`, `review_quality_multiplier`, and `credibility_multiplier` also use `round(..., 2)`, but their values already sit at 2-decimal boundaries (1.00, 1.25, 1.50, etc.) so the loss is negligible. `repo_weight` is uniquely affected because it clusters near 0.03.

**Suggested fix:** drop the `round(..., 2)` at both assignment sites so the full 4-decimal precision reaches the `earned_score` product. Display/log formatting can round independently.

PR #897, opened 2026-04-30
fix(cli): unify gitt issues list fill_pct math across Panel and table paths

### LINKED ISSUE #896: [Bug] gitt issues list shows different "Fill %" for the same issue depending on --id mode vs table mode (raw float div vs Decimal)
## Description

`gitt issues list` computes "Fill %" two different ways for the same field, in the same function, depending on which render path runs:

[`view.py:95`](gittensor/cli/issue_commands/view.py#L95) — single-issue Panel view (`--id <N>`):
```python
fill_pct = (bounty_raw / target_raw * 100) if target_raw > 0 else 0
```

[`view.py:146`](gittensor/cli/issue_commands/view.py#L146) — table row (no `--id`):
```python
fill_pct = float(Decimal(bounty_val) / Decimal(target_val) * 100)
```

Both branches read the same on-chain fields (`bounty_amount`, `target_bounty`) and render the same percentage in the UI. The Panel path uses raw `int / int * 100` and is subject to ordinary float artifacts. The table path goes through `Decimal` first, which is the well-known fix for those artifacts.

For most values the two produce the same display string, but for bounties whose ratio lands on a binary-float-unfriendly number (e.g., 1/3 of target, 1/7, 5/12), the two render paths print marginally different "Fill %" values for the **same issue**, depending on whether the operator passed `--id` or not. That's a UI consistency bug: the same field, queried two ways within seconds of each other, shows different numbers.

## Reproduction

Pick (or register) any issue where `bounty_amount` and `target_bounty` produce a ratio that isn't representable exactly in binary float. A clean case: bounty = 1, target = 3.

```bash
$ gitt i list
[ ... table shows fill_pct via Decimal: "0/3.0 (33%)" — exact ... ]

$ gitt i list --id <N>
[ ... Panel shows fill_pct via raw float: 33.33333333333333 ... ]
```

The two paths use the same data; only the formula differs.

## Why this isn't a duplicate

- **#652 / #654 / #655** (closed) were about repo-weight rounding to 2 decimals discarding precision in `master_repositories.json` ingestion — different code path (load-time JSON parse), different precision concern.
- **#335 (MERGED)** changed the not-found JSON shape on this same path; it didn't touch the math at lines 95 or 146.
- No filed item touches the `Fill %` formula or the Panel-vs-table inconsistency.

## Proposed fix

Pick one path; the `Decimal` form is correct (deterministic, no float artifacts). Apply it at line 95:

```python
# view.py:93-95
bounty_raw = issue.get('bounty_amount', 0)
target_raw = issue.get('target_bounty', 0)
fill_pct = float(Decimal(bounty_raw) / Decimal(target_raw) * 100) if target_raw > 0 else 0
```

A trivial extraction makes the agreement enforced rather than coincidental:

```python
# view.py module-level
def _fill_percent(bounty: int, target: int) -> float:
    if target <= 0:
        return 0.0
    return float(Decimal(bounty) / Decimal(target) * 100)
```

Then both call sites use `_fill_percent(...)`. ~10 lines diff total, no new dependency.

## Tests

`tests/cli/test_issues_list_json.py` already covers the not-found JSON path from PR #335. Two parametric cases pin the new behavior:

| Bounty | Target | Expected Fill % | Asserts |
|---|---|---|---|
| 1 | 3 | 33.333… | Panel and table strings agree to ≥6 decimals |
| 5 | 12 | 41.666… | same |

These are unit tests on `_fill_percent`; the existing rendering tests remain unchanged.

## Scope

- 1 file (`view.py`), ~10 lines, no API changes, no test fixtures touched.
- Aligns the two render paths on a single computation; no behavior change for ratios that round identically under both formulas.

## Coordination

`view.py:issues_list` is currently being touched by **#855** and **#856** (both validate `--id` at parse time). This fix lives further down in the function body (the math, not the parse gate); whichever lands first, the other rebases trivially.

## Labels

`bug`

PfanP:
## Summary

`gitt issues list` computed `Fill %` two different ways for the same field in the same function:

- [`view.py:95`](gittensor/cli/issue_commands/view.py#L95) — single-issue Panel view (`--id <N>`):
  ```python
  fill_pct = (bounty_raw / target_raw * 100) if target_raw > 0 else 0
  ```
- [`view.py:146`](gittensor/cli/issue_commands/view.py#L146) — table all-issues view:
  ```python
  fill_pct = float(Decimal(bounty_val) / Decimal(target_val) * 100)
  ```

Same field, same on-chain data, two formulas. For ratios that don't round identically under both (e.g., 1/3, 5/12, 1/7), the same issue rendered marginally different `Fill %` values depending on whether the operator used `--id` or not.

## Fix

Extract a shared `_fill_percent(bounty, target)` helper using the Decimal path (correct, deterministic — no float-binary artifacts) and route both render paths through it.

```python
def _fill_percent(bounty: int, target: int) -> float:
    if target <= 0:
        return 0.0
    return float(Decimal(bounty) / Decimal(target) * 100)
```

The agreement between the two render paths is now enforced by construction, not coincidence.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Reproduction (before this PR)

For an issue with `bounty=1, target=3`:

```bash
$ gitt i list
[ ... table renders "33%" via Decimal ... ]

$ gitt i list --id <N>
[ ... Panel renders fill_pct via raw float ... ]
# Fill %: 33.3%
```

The two paths agreed for most ratios but diverged on values where binary float and Decimal disagree at the rounding boundary.

## After

Both render paths call `_fill_percent`. Format precision differs by mode (`{:.0f}%` for table, `{:.1f}%` for Panel), but both formats render the same underlying value.

## Why this isn't a duplicate

- **#652 / #654 / #655** (closed) addressed precision loss in `master_repositories.json` ingestion (rounding to 2 decimals) — different code path, different concern.
- **#335 (MERGED)** changed the not-found JSON shape on this same path; it didn't touch the math at lines 95 or 146. Existing tests from #335 are kept passing.
- No filed item touches the `Fill %` formula or the Panel-vs-table inconsistency.

## Tests

`tests/cli/test_issues_list_fill_pct.py` — 13 new cases:

| Group | Asserts |
|---|---|
| **Unit** (5) | `_fill_percent(0, 0) == 0`, negative target == 0, `(100, 100) == 100`, over-full > 100, parametric ratios (1/3, 5/12, 1/7, 2/11) match expected to 1e-12 |
| **Behavioral** (4) | Table mode shows `33%` for 1/3, Panel mode shows `33.3%` for 1/3, both modes show `33` for the same data, JSON mode passes `bounty_amount`/`target_bounty` through unchanged |

```
$ pytest tests/cli/test_issues_list_fill_pct.py -v
13 passed in 0.28s

$ pytest tests/cli/test_issues_list_json.py -v   # PR #335 regression tests
2 passed in 0.16s
```

## Scope

- 2 files changed: `view.py` (+12 helper, 2 call-site swaps), one new test file.
- No API changes, no schema changes, no JSON contract changes.
- Single function-internal extraction; matches the pattern from PR #498 (extract `validate_issue_id` / `validate_ss58`) and PR #442 (extract `_resolve_contract_and_network`).

## Coordination

`view.py:issues_list` is currently being touched by **#855** and **#856** (both validate `--id` at parse time). This fix lives in the function body (the math, not the parse gate); whichever lands first, the other rebases trivially.

## Related

- #335 (MERGED) — structured `not_found` JSON; not regressed (existing test still passes).
- #498 (MERGED) — pattern of extracting validators/helpers; mirrors that shape here.
- #855 / #856 (OPEN) — adjacent in same function; no overlap.

## Labels

`bug`

Fixes #896


JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W2)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W18)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W3)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W24)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W20)

anderdc (changes requested):
The bug premise doesn't hold — `int / int * 100` and `float(Decimal(int) / Decimal(int) * 100)` collapse to the same float, so this isn't fixing a divergence. The extraction is a fine refactor on its own though, so I'll take it.

Scope down: keep the `_fill_percent` helper and the two call-site swaps in `gittensor/cli/issue_commands/view.py`. Drop `tests/cli/test_issues_list_fill_pct.py` entirely — the unit tests assert Python arithmetic, the "behavioral" tests pin format-string output, and the cross-mode test admits in its own docstring that both formulas yield the same float.

mkdev5:
The implementation is small, but the regression test should use a value that actually fails before this change.

For `bounty=1,target=3`, the reported `33%` vs `33.3%` difference is from the existing table/panel format specs (`{:.0f}` vs `{:.1f}`), not from the math path, so these assertions do not prove the call sites were unified. Please switch the fixture to a rounding-boundary case that visibly differs under the old panel formula, e.g. `bounty=23,target=80`: old panel math renders `28.7%`, while the Decimal helper renders `28.8%`. That would give maintainers a concrete pre-fix failure and make the PR rationale much stronger.

anderdc (changes requested):
Drop tests/cli/test_issues_list_fill_pct.py; keep only the helper + two call-site swaps. The 23/80 rounding-boundary example is real, but the cosmetic impact doesn't warrant 176 lines of tests. Same scope-down as the prior review.

anderdc (approved):

PR #901, opened 2026-04-30
fix: isolate malformed mirror PR child rows

### LINKED ISSUE #900: Bug: malformed mirror PR child rows can drop an otherwise valid PR from OSS scoring
## Description

A malformed optional child row inside a mirror PR payload can cause the entire PR to be skipped during parsing.

In `gittensor/utils/mirror/models.py`, `MirrorPullRequest.from_dict()` parses nested `labels` and `linked_issues` inline. If one nested child row is malformed, `MirrorPullRequestsResponse.from_dict()` skips the whole PR instead of only skipping the bad child row.

This can drop an otherwise valid merged PR from mirror OSS scoring.

## Steps to Reproduce

1. Create a mirror PR payload with valid PR-level fields.
2. Add a malformed nested `linked_issues` item, for example one missing `state`.
3. Call `MirrorPullRequestsResponse.from_dict(payload)` and inspect `pull_requests`.

## Expected Behavior

The parent PR should still be parsed and kept if its required PR-level fields are valid.

Only the malformed optional child row should be skipped with a warning.

## Actual Behavior

The malformed child row raises during PR parsing, and the whole PR is skipped.

Example result:
- warning like `Skipping malformed mirror PR Entrius/Gittensor#123: 'state'`
- parsed PR count is `0` instead of `1`

## Environment

- OS: Ubuntu/Linux
- Python version: 3.12
- Commit/Version: `test` branch as of 2026-04-30

## Additional Context

This is inconsistent with `MirrorPullRequestFilesResponse.from_dict()`, which already isolates malformed child `files` rows and keeps the rest of the parent response.

The impact is larger than the bad child row itself: one malformed optional `labels` or `linked_issues` item can remove the PR from OSS scoring entirely.

Crystora:
## Summary

Fixes mirror PR parsing so malformed optional child rows do not drop an otherwise valid PR from OSS scoring.

This change isolates malformed nested `labels` and `linked_issues` rows during mirror PR parsing:
- keeps the parent `MirrorPullRequest` when required PR-level fields are valid
- skips only the malformed child row
- logs a warning for the skipped child

It also adds regression tests covering malformed PR child `linked_issues` and `labels`.

## Related Issues

Fixes #900

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Commands run:
- `.venv/bin/python -m pytest tests/utils/test_mirror_models.py -q`

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/mirror/models.py
+++ b/gittensor/utils/mirror/models.py
@@ -21,6 +21,34 @@ from gittensor.validator.utils.datetime_utils import (
 )
 
 
+def _parse_mirror_labels(raw_labels: list[dict], parent_identifier: str) -> list['MirrorLabel']:
+    labels: list[MirrorLabel] = []
+    for raw in raw_labels or []:
+        try:
+            labels.append(MirrorLabel.from_dict(raw))
+        except Exception as e:
+            name = raw.get('name', '?') if isinstance(raw, dict) else '?'
+            bt.logging.warning(
+                f'Skipping malformed mirror label {name} on {parent_identifier}: {e}'
+            )
+    return labels
+
+
+def _parse_mirror_linked_issues(
+    raw_issues: list[dict], pr_identifier: str
+) -> list['MirrorLinkedIssue']:
+    linked_issues: list[MirrorLinkedIssue] = []
+    for raw in raw_issues or []:
+        try:
+            linked_issues.append(MirrorLinkedIssue.from_dict(raw))
+        except Exception as e:
+            identifier = raw.get('number', '?') if isinstance(raw, dict) else '?'
+            bt.logging.warning(
+                f'Skipping malformed mirror linked issue {identifier} on {pr_identifier}: {e}'
+            )
+    return linked_issues
+
+
 @dataclass
 class MirrorLabel:
     """A label applied to a PR or issue, with the actor who applied it.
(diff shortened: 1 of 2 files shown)


anderdc:
`MirrorPullRequestsResponse.from_dict` only drops a PR if `MirrorPullRequest.from_dict` raises, which (for child rows) requires a `KeyError` on a missing key — `name` for `MirrorLabel`, `number`/`state` for `MirrorLinkedIssue`. The mirror server emits children via `json_build_object` in [`das-github-mirror/packages/das/src/api/miners/miners.service.ts:62-99`](https://github.com/entrius/das-github-mirror/blob/main/packages/das/src/api/miners/miners.service.ts#L62-L99); the listed keys are always present. Null DB values yield `'name': null`, not absent keys, and `data['name']` returns `None` without raising — no exception, no dropped PR.

The symmetry with `MirrorPullRequestFilesResponse.from_dict` doesn't carry: file rows have many more parse-time failure modes (binary/text boundaries, size limits, status enum drift), file payloads are megabytes each, and dropping all files for one PR has a real scoring cost. Labels (3 fields) and linked_issues (12 fields, all `.get()`-defaulted except `number`/`state`) don't have analogous failure modes against the SQL contract.

Net effect of the change is negative on observability: a genuinely malformed payload today is loud (whole-PR drop, top-level warning, alertable). After this change the same malformed payload would silently keep the parent with empty `labels`/`linked_issues` and one per-child warning — quieter, and it would silently feed degraded data into the label-multiplier and linked-issue gates. Closing.

PR #903, opened 2026-04-30
Migrate psycopg2 -> psycopg 3

### LINKED ISSUE #902: Migrate psycopg2 -> psycopg 3
## Summary and motivation

Currently used `psycopg2` is in maintenance-only mode. `psycopg3` is the actively developed successor and is the recommended driver going forward. Quote from [README.md](https://github.com/psycopg/psycopg2):

> The psycopg2 package is still widely used and actively maintained, but it is not expected to receive new features.
> [Psycopg 3](https://pypi.org/project/psycopg/) is the evolution of psycopg2 and is where [new features are being developed](https://www.psycopg.org/psycopg3/docs/index.html): if you are starting a new project you should probably start from 3!

Migration likely requires mostly nothing, the only changed thing seems is a `execute_values` removal. [v2 -> v3 migration guide](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html) just in case.

seroperson:
Closes #902

## Summary

1. Replace `psycopg2-binary` with `psycopg[binary]>=3.2`
1. Drop `execute_values` (unsupported in v3) and switch each bulk upsert to `cursor.executemany`, with the SQL templates expanded from `VALUES %s` to explicit `VALUES (%s, %s, ...)` row tuples
1. Wrap all per-miner writes inside `db_connection.pipeline()` to batch queries into one round-trip while preserving the existing transaction boundary. From [documentation](https://www.psycopg.org/psycopg3/docs/advanced/pipeline.html):
> "The pipeline mode allows PostgreSQL client applications to send a query without having to read the result of the previously sent query... multiple queries/results can be sent/received in a single network roundtrip."
4. Set `connection.prepare_threshold = 0` so our queries are auto-prepared, instead of old behavior, when it was undefined. From [documentation](https://www.psycopg.org/psycopg3/docs/advanced/prepare.html):
> "A query is prepared automatically after it is executed more than prepare_threshold times on a connection..."
5. No schema change, no public API change on `Repository` or `DatabaseStorage`
6. Use `dbname` arg instead of `database`, like in canonical usage example: https://www.psycopg.org/psycopg3/docs/basic/usage.html#main-objects-in-psycopg-3; Keyword list: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

## Test plan

I haven't added new tests as currently there are no any tests which cover the real database connection, either any migrations to spin-up everything locally. I did some synthetic manual testing and everything seems good.

I can try to implement necessary code here to test it against a real connection and do real tests if you want.

### DIFF
--- a/gittensor/validator/storage/database.py
+++ b/gittensor/validator/storage/database.py
@@ -8,12 +8,12 @@ from typing import Any, Optional
 import bittensor as bt
 
 try:
-    import psycopg2
+    import psycopg
 
     POSTGRES_AVAILABLE = True
 except ImportError:
     POSTGRES_AVAILABLE = False
-    bt.logging.warning('psycopg2 not installed. Database storage features will be disabled.')
+    bt.logging.warning('psycopg not installed. Database storage features will be disabled.')
 
 
 def create_database_connection() -> Optional[Any]:
@@ -24,7 +24,7 @@ def create_database_connection() -> Optional[Any]:
         Database connection if successful, None otherwise
     """
     if not POSTGRES_AVAILABLE:
-        bt.logging.error('Cannot create database connection: psycopg2 not installed')
+        bt.logging.error('Cannot create database connection: psycopg not installed')
         return None
 
     try:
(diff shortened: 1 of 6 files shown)


anderdc (changes requested):
1. Fix conflicts.
2. Pin `psycopg[binary]` to a specific version (e.g. `==3.3.3`) — rest of `pyproject.toml` uses exact pins.

anderdc (approved):

PR #905, opened 2026-04-30
fix: block stale cache fallback on GraphQL PAT auth failures

### LINKED ISSUE #904: [Bug] GraphQL PAT auth failures can restore stale cached PR rewards
## Description

A stored miner PAT can pass REST `/user` validation but fail the PR GraphQL query used during scoring with an auth/scope error, for example:

`Resource not accessible by personal access token`

`load_miners_prs` marks `github_pr_fetch_failed=True`, but does not set `failed_reason`. Because cache fallback only skips miners with `failed_reason`, the validator can restore a previous successful cached evaluation and continue rewarding stale PRs even though the current PAT no longer has the GraphQL access scoring requires.

This is different from transient GitHub/network/resource-limit failures: an auth/scope failure is not self-healing until the miner submits a usable PAT.

## Steps to Reproduce

1. Seed `MinerEvaluationCache` with a previous successful evaluation for a UID/hotkey/github_id.
2. Run scoring for the same UID with a stored PAT that passes REST `/user`.
3. Mock the PR GraphQL query to return:
   `{"errors":[{"message":"Resource not accessible by personal access token"}],"data":null}`
4. Call `store_or_use_cached_evaluation`.

## Expected Behavior

GraphQL auth/scope failures should be treated as non-cacheable validation failures.

The miner should not receive stale cached rewards until they submit a PAT that can perform the GraphQL queries required for scoring.

## Actual Behavior

`failed_reason` remains `None`, `github_pr_fetch_failed=True`, `total_prs=0`, and cache fallback restores the old successful evaluation.

## Environment

- Python version: 3.12
- Branch: current `test`

## Additional Context

Related but not duplicate:

- #515 / #516 fixed valid zero-PR results using stale cache.
- #644 / #645 discussed transient per-PR fetch failures and was closed as not planned.
- This issue is narrower: clear GraphQL PAT auth/scope failures should not be treated like transient GitHub outages.

Likely fix: detect GraphQL auth/scope errors such as `Resource not accessible by personal access token`, set `failed_reason`, and preserve cache fallback for transient network/5xx/resource-limit failures.

bitloi:
## Summary

Fixes stale cache fallback when a stored miner PAT passes REST `/user` validation but fails the PR GraphQL query used during scoring with an auth/scope error.

Previously, `load_miners_prs()` marked these GraphQL failures with `github_pr_fetch_failed=True` but left `failed_reason=None`. Since cache fallback only skips miners with `failed_reason`, validators could restore a previous successful cached evaluation and continue rewarding stale PRs even though the current PAT no longer had the GraphQL access required for scoring.

This change detects clear GraphQL PAT auth/scope errors, such as `Resource not accessible by personal access token`, marks the evaluation with a non-cacheable `failed_reason`, and preserves cache fallback for transient/non-auth GraphQL failures.

## Related Issues

Closes #904

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Commands run:

```bash
uv run pytest tests/utils/test_github_api_tools.py tests/validator/test_validator_cache_fallback.py -q
uv run pytest tests/ -q
uv run ruff check gittensor/utils/github_api_tools.py tests/utils/test_github_api_tools.py tests/validator/test_validator_cache_fallback.py
uv run ruff format --check gittensor/utils/github_api_tools.py tests/utils/test_github_api_tools.py tests/validator/test_validator_cache_fallback.py
uv run pyright gittensor/utils/github_api_tools.py tests/utils/test_github_api_tools.py tests/validator/test_validator_cache_fallback.py
uv run pre-commit run --files gittensor/utils/github_api_tools.py tests/utils/test_github_api_tools.py tests/validator/test_validator_cache_fallback.py
```

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -38,6 +38,15 @@ from gittensor.validator.utils.load_weights import RepositoryConfig
 # Beyond this many CHANGES_REQUESTED reviews the quality multiplier is already 0
 _MAX_CHANGES_REQUESTED_REVIEWS = ceil(1 / REVIEW_PENALTY_RATE)
 
+_GRAPHQL_PAT_AUTH_SCOPE_FAILED_REASON = 'GitHub GraphQL PAT auth/scope failure'
+_GRAPHQL_PAT_AUTH_SCOPE_ERROR_MARKERS = (
+    'resource not accessible by personal access token',
+    'bad credentials',
+    'requires authentication',
+    'not have the required scopes',
+    'not been granted the required scopes',
+)
+
 # core github graphql query
 QUERY = """
     query($userId: ID!, $limit: Int!, $cursor: String, $maxChangesRequestedReviews: Int!) {
@@ -723,6 +732,18 @@ def _has_resource_limit_errors(data: Dict) -> bool:
     return any(isinstance(e, dict) and e.get('type') == 'RESOURCE_LIMITS_EXCEEDED' for e in errors)
 
 
+def _graphql_pat_auth_scope_error_message(errors: List[Dict]) -> Optional[str]:
+    """Return the first GraphQL PAT auth/scope error message, if present."""
+    for error in errors:
+        if not isinstance(error, dict):
+            continue
+        message = str(error.get('message') or '')
+        normalized = message.lower()
+        if any(marker in normalized for marker in _GRAPHQL_PAT_AUTH_SCOPE_ERROR_MARKERS):
+            return message
+    return None
+
+
 def try_add_open_or_closed_pr(
     miner_eval: MinerEvaluation,
     pr_raw: Dict,
@@ -908,7 +929,12 @@ def load_miners_prs(
             if 'errors' in data:
                 non_resource_errors = [e for e in data['errors'] if e.get('type') != 'RESOURCE_LIMITS_EXCEEDED']
                 if non_resource_errors:
-                    bt.logging.error(f'GraphQL errors: {non_resource_errors}')
+                    auth_scope_error = _graphql_pat_auth_scope_error_message(non_resource_errors)
+                    if auth_scope_error:
+                        bt.logging.error(f'GraphQL PAT auth/scope failure: {auth_scope_error}')
+                        miner_eval.failed_reason = _GRAPHQL_PAT_AUTH_SCOPE_FAILED_REASON
+                    else:
+                        bt.logging.error(f'GraphQL errors: {non_resource_errors}')
                     miner_eval.github_pr_fetch_failed = True
                     break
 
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
Right bug, but the marker list is the brittle part — GitHub can rephrase any of those five strings and a real auth failure leaks back into cache fallback.

Simpler and more robust: in the `non_resource_errors` branch at `gittensor/utils/github_api_tools.py:929`, just always set `failed_reason`. Something like:

```python
if non_resource_errors:
    bt.logging.error(f'GraphQL errors: {non_resource_errors}')
    miner_eval.failed_reason = f'GitHub GraphQL error: {non_resource_errors[0].get("message", "unknown")[:200]}'
    break
```

This branch only fires when GitHub returns 200 with errors that aren't `RESOURCE_LIMITS_EXCEEDED`. Every other transient class — network exceptions, 5xx, rate limits — hits a different branch and keeps cache fallback. So the GraphQL-errors-array path is overwhelmingly auth / scope / schema / permission failures, none of which should restore stale cached scores. A genuinely transient `INTERNAL_SERVER_ERROR` from GitHub eats one round at 0, which is recoverable next round.

Drop the marker tuple, the helper, and the auth-specific test. Keep the integration test in `test_validator_cache_fallback.py` but assert it works for any GraphQL error message, not just the auth phrasing.

bitloi:
> Right bug, but the marker list is the brittle part — GitHub can rephrase any of those five strings and a real auth failure leaks back into cache fallback.
> 
> Simpler and more robust: in the `non_resource_errors` branch at `gittensor/utils/github_api_tools.py:929`, just always set `failed_reason`. Something like:
> 
> ```python
> if non_resource_errors:
>     bt.logging.error(f'GraphQL errors: {non_resource_errors}')
>     miner_eval.failed_reason = f'GitHub GraphQL error: {non_resource_errors[0].get("message", "unknown")[:200]}'
>     miner_eval.github_pr_fetch_failed = True
>     break
> ```
> 
> This branch only fires when GitHub returns 200 with errors that aren't `RESOURCE_LIMITS_EXCEEDED`. Every other transient class — network exceptions, 5xx, rate limits — hits a different branch and keeps cache fallback. So the GraphQL-errors-array path is overwhelmingly auth / scope / schema / permission failures, none of which should restore stale cached scores. A genuinely transient `INTERNAL_SERVER_ERROR` from GitHub eats one round at 0, which is recoverable next round.
> 
> Drop the marker tuple, the helper, and the auth-specific test. Keep the integration test in `test_validator_cache_fallback.py` but assert it works for any GraphQL error message, not just the auth phrasing.

@anderdc Updated. Non-resource GraphQL errors now always set `failed_reason`, and tests no longer depend on auth-specific wording.

anderdc (approved):

PR #910, opened 2026-04-30
feat: add --repo filter to gitt issues list

### LINKED ISSUE #892: [Feature] gitt issues list dumps every on-chain issue with no way to filter — make `--repo <owner/name>` filter at the CLI level
## Description

`gitt issues list` reads every issue registered on-chain and renders them all into a single table at [`view.py:114-171`](gittensor/cli/issue_commands/view.py#L114-L171). There is no `--repo` filter, so the only way to see issues for a specific repository is to scroll the table or pipe `--json` output through `jq`.

This command is the **most-frequent read-side entry point** into the contract — operators run it to confirm their bounty registered, to see active issues across repos they contribute to, to spot-check ingestion. Today the experience scales linearly with on-chain issue count.

## Concrete operator workflow that's broken

A miner contributing to one repo (say `kubernetes/kubernetes`) currently has to do one of:

```bash
# Option A: scroll the table and eyeball
$ gitt issues list
[ ... 80+ rows across all tracked repos ... ]

# Option B: pipe to jq and rebuild the rendering
$ gitt issues list --json | jq '.issues[] | select(.repository_full_name == "kubernetes/kubernetes")'
{ ... raw JSON, lost the rich coloring/rendering ... }

# Option C: grep the table output
$ gitt issues list | grep kubernetes/kubernetes
[ no help text, no header row, raw rich-formatted lines that may break with --json off ]
```

None of these is a CLI workflow; they're shell workarounds because the CLI doesn't have the filter operators expect.

## Why this scales worse over time

`master_repositories.json` already lists 100+ tracked repos. Issue registration is open to any contributor on any tracked repo, so on-chain issue count grows ~unboundedly relative to a single operator's interest. Every operator running `gitt issues list` today is paying the cost of every other operator's bounties.

## Proposed fix

Add `--repo <owner/name>` to [`issues_list`](gittensor/cli/issue_commands/view.py#L37-L51):

```python
@click.option(
    '--repo',
    'repo_filter',
    default=None,
    type=str,
    help='Filter issues to a specific repository (owner/name).',
)
```

After the contract read, before rendering:

```python
if repo_filter:
    issues = [i for i in issues if i.get('repository_full_name') == repo_filter]
```

Single filter at function entry. No new helpers. No schema change. Reuses the existing `--id` precedent of "client-side filter on the already-fetched list."

## Before / After

```
$ gitt issues list --repo kubernetes/kubernetes
Available Issues
┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ ID ┃ Repository                 ┃ Issue # ┃ Bounty Pool ┃  Status  ┃
┡━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━┩
│  3 │ kubernetes/kubernetes      │ #11234  │ 12.0/30.0   │  Active  │
│  7 │ kubernetes/kubernetes      │ #11402  │ 8.0/15.0    │  Active  │
└────┴────────────────────────────┴─────────┴─────────────┴──────────┘
Showing 2 issue(s)
```

`--json` mode includes the same filter; `--id` continues to take precedence (return the single issue regardless of `--repo`).

## Out of scope (not proposing here)

- `--status` filter (active/completed/cancelled) — separate proposal if `--repo` lands; tackles a different axis.
- `--limit N` — only adds value once the per-repo filter still produces a long list; defer until that's a real problem.

Keeping this issue narrow on purpose: one filter, one file, one diff.

## Coordination note

`view.py:issues_list` is currently being touched by **#855** and **#856** (both add `validate_issue_id` to the `--id` path). Whichever lands first, the `--repo` patch rebases trivially — no overlap on the validation gate, just adds a separate filter line.

Tet-9:
Fixes #892

## Problem
`gitt issues list` renders every on-chain issue with no way to filter
by repository. With 100+ tracked repos and unbounded issue growth,
operators contributing to one repo are forced to scroll through
unrelated issues or resort to shell workarounds like `jq` or `grep`.

## Fix
Added `--repo <owner/name>` option to `issues_list` in `view.py`.
The filter is applied client-side after the contract read, before
rendering — both in JSON mode and table mode. `--id` takes precedence
over `--repo` (the existing single-issue path is unchanged).
The filter is case-insensitive.

## Changes
- `gittensor/cli/issue_commands/view.py` — added `--repo` option and
  two filter lines (JSON path and table path)

### DIFF
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -42,13 +42,22 @@ from .helpers import (
     type=int,
     help='View a specific issue by ID',
 )
+@click.option(
+    '--repo',
+    'repo_filter',
+    default=None,
+    type=str,
+    help='Filter issues to a specific repository (owner/name).',
+)
 @with_cli_behavior_options(
     include_verbose=True,
     include_json=True,
     verbose_help='Show debug output for contract reads',
 )
 @with_network_contract_options('Contract address (uses default if empty)')
-def issues_list(issue_id: int, network: str, rpc_url: str, contract: str, verbose: bool, as_json: bool):
+def issues_list(
+    issue_id: int, repo_filter: str, network: str, rpc_url: str, contract: str, verbose: bool, as_json: bool
+):
     """List issues or view a specific issue.
 
     [dim]Examples:
@@ -76,6 +85,11 @@ def issues_list(issue_id: int, network: str, rpc_url: str, contract: str, verbos
         for issue in issues:
             issue['bounty_alpha'] = format_alpha(issue.get('bounty_amount', 0), 4)
             issue['target_alpha'] = format_alpha(issue.get('target_bounty', 0), 4)
+
+        # Apply --repo filter before rendering (--id takes precedence)
+        if repo_filter and issue_id is None:
+            issues = [i for i in issues if i.get('repository_full_name', '').lower() == repo_filter.lower()]
+
         if issue_id is not None:
             issue = next((i for i in issues if i['id'] == issue_id), None)
             if issue is None:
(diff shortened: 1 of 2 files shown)


igeabdulrahmanikeoluwa-str:
@anderdc , review time 🫡

anderdc (approved):

PR #914, opened 2026-04-30
enh: decode mirror responses with msgspec

### LINKED ISSUE #915: enh: decode mirror responses with msgspec
## Summary

Replace the `@dataclass` and hand-written `from_dict` boilerplate in `gittensor/utils/mirror/models.py` with `msgspec.Struct` ([msgspec repo](https://github.com/jcrist/msgspec)) automatic schema derivation, and swap `response.json()` for `msgspec.json.decode(response.content)` in `gittensor/utils/mirror/client.py`. `msgspec` parses + validates JSON in C, so everything is also a little bit faster.

## Motivation

Currently we do the whole parsing literally manually and it's really a lot of boilerplate to code and handle. Mostly the problem is the boilerplate, but as a bonus we receive a faster json parsing in C. Yes, it's an additional dependency, but it's time-proven, small, and used in a wide amount of popular projects, like polars, prefect, litestar etc.

seroperson:
Closes #915

## Summary

This PR replace the `@dataclass` and hand-written `from_dict` boilerplate methods with automatic schema derivation and parsing using `msgspec` library. It allows us to get rid of boilerplate + benefit some extra performance boost, as `msgspec` parses JSON in C.

Still there is logic which lowers elements of some arrays, fixes strings sent as ints, and skips nested childs instead if they're malformed. Honestly, as we're managing both mirror and gittensor end, I feel like this should be done at mirror's end, and here, in gittensor we should fail-fast if something goes wrong. But I left everything as-is to preserve compatibility.

It's an additional dependency, but it's time-proven, small, and used in a wide amount of popular projects, like polars, prefect, litestar etc.

## Test plan

The parsing was already covered with tests, so I just added two small e2e cases to ensure JSON decoding works correctly. I can add more parsing cases to ensure that nothing broke, but I feel this too much tests for parsing a response.


anderdc:
Current implementation works fine as is — I don't see the need for the rewrite. Net diff is -12 lines; the boilerplate moves into msgspec defaults, `_coerce_id_fields`, and `__post_init__` hooks rather than going away. Perf delta on KB-scale mirror payloads is in the noise vs the rest of the scoring loop. Per the PR description, the right home for the int->str / lowercasing coercions is the mirror server, not a client-side recursive walker. Closing.

seroperson:
@anderdc It's up to you of course, but let me advocate just a little =)

> Net diff is -12 lines; the boilerplate moves into msgspec defaults, 

Such diff is because I've modified existing lines (to set defaults) and they counted as additions.

> the boilerplate moves into msgspec defaults, _coerce_id_fields, and __post_init__ hooks rather than going away

The most notable boilerplate was duplicating every existing field in `from_dict` with `data.get`. This gone.
We have to set defaults instead, but it's much lesser evil than duplicating every field. After mirror response cleanup this can be removed as well.

> Per the PR description, the right home for the int->str / lowercasing coercions is the mirror server, not a client-side recursive walker

Here I just left the pre-existing logic to ensure everything is backward-compatible. After mirror is stabilized and we have some strong API contract, `_coerce_id_fields`, `__post_init__` and per-field defaults can be dropped.

I can actually send accompanying PR to the mirror to ensure response is clear, and then delete all defaults and coercions in this PR. After such PR land, the only thing left here is the plain response model. What do you think?

Issue #533, opened 2026-04-17
[BUG] Duplicate-GitHub penalty destroys github_id/total_open_issues, diverging DB from in-memory state


dev-miro26:
## Description

`detect_and_penalize_miners_sharing_github` (`gittensor/validator/oss_contributions/inspections.py:40`) replaces `miner_evaluations[uid]` with a brand-new `MinerEvaluation(uid, hotkey)`, losing `github_id` (which resets to `'0'`), `total_open_issues`, and any cached PR data. If the uid is already in `cached_uids` from a fallback earlier in the round, `bulk_store_evaluation` skips the overwrite — so the DB/W&B row retains the previously-cached non-zero score while the in-memory object reports 0. Chain weight and reported score diverge.

## Steps to Reproduce

1. UID 12 scored successfully in a prior round (cached).
2. This round, GitHub API fails for UID 12 → fallback used, `cached_uids.add(12)`.
3. UID 13 is discovered to share github_id with UID 12.
4. Dup detection zeroes both uids in-memory; DB storage skips both because both are in `cached_uids`.
5. W&B and DB report stale non-zero scores; on-chain weight is 0 for both.

## Expected Behavior

Zeroing a duplicate should remove the uid from `cached_uids` and force the overwrite, so in-memory and DB agree. `total_open_issues` should be aggregated (max/sum) across sharing uids before zeroing, so downstream spam gating still applies.

## Actual Behavior

Cached uids are not force-overwritten; stale DB rows survive.

## Additional Context

Suggested fix: extract `_replace_with_zero_eval(uid, hotkey, cached_uids) -> MinerEvaluation` helper that constructs the penalty evaluation *and* removes `uid` from `cached_uids`. Call it from the dup-detection loop. Add a test that asserts DB/in-memory consistency after dup detection.

PR #921, opened 2026-05-01
fix: remove duplicate github_id entries from eligible_miners instead of silently overwriting 

### LINKED ISSUE #895: [Bug] eligible_miners dict in issue_competitions silently collapses duplicate github_id collisions, causing bounty mis-vote or stuck Active issues
## Description

In [`gittensor/validator/issue_competitions/forward.py:65-69`](gittensor/validator/issue_competitions/forward.py#L65-L69), the `eligible_miners` dict is built via a comprehension keyed by `github_id`:

```python
eligible_miners = {
    eval.github_id: eval.hotkey
    for eval in miner_evaluations.values()
    if eval.github_id and eval.github_id != '0' and eval.is_eligible
}
```

If two eligible `MinerEvaluation` entries share the same `github_id`, the second one **silently overwrites the first** with no log, no warning, and a non-deterministic winner (Python dict iteration order over `.values()`).

This dict is then used at line 126 to look up which hotkey to vote-pay for a given GitHub solver:

```python
miner_hotkey = eligible_miners.get(str(solver_github_id))
```

When the surviving hotkey is *not* the actual PR author, validators vote for the wrong miner's hotkey — or fall through to a cancel vote when `get_miner_coldkey()` fails on the surviving hotkey.

## Why this happens despite the duplicate-account penalty

The upstream `detect_and_penalize_miners_sharing_github` system zeroes scores for sharing miners, but this comprehension still fires for any duplicate that:

- Slips past the penalty within a single round (race between PAT broadcast and penalty pass)
- Is preserved across rounds because `is_eligible` and `github_id` are set independently of the penalty zeroing (see `inspections.py` history in #380, #544)
- Has had its DB/in-memory state diverge due to past bugs (#533, #596) — even if those are fixed, the comprehension provides no defensive logging if a future regression reintroduces duplicates

In other words: **this code path silently propagates upstream penalty bugs into bounty mis-payments instead of failing loudly**.

## Steps to Reproduce

1. Set up two `MinerEvaluation` entries in `miner_evaluations` with:
   - Different `uid` and `hotkey`
   - Same `github_id` (e.g., both `'12345'`)
   - Both `is_eligible = True`
2. Trigger the issue_competitions forward pass.
3. Observe: `eligible_miners['12345']` resolves to whichever hotkey was iterated last (non-deterministic).
4. The validator votes for that hotkey on any bounty whose solver matches `github_id 12345` — even if it's not the hotkey that authored the merged PR.

## Failure Modes

| Surviving hotkey | Outcome |
|------------------|---------|
| Wrong miner's hotkey | Bounty pays the wrong miner |
| Hotkey with no registered coldkey | `get_miner_coldkey()` returns `None`; validator skips voting; if all validators do the same → bounty stuck `Active` indefinitely |
| Hotkey is the correct one (lucky) | Works correctly, but undetectably racing |

## Severity

**Medium** — bounty mis-payment or availability lockup. Trigger is rare today (depends on duplicate slipping past upstream penalty), but consequences are silent and uncorrectable on-chain. No log entry exists today to detect when this fires in production.

## Proposed Fix

Replace the silent comprehension with an explicit duplicate-detection loop that:
1. Logs every duplicate `github_id`
2. Removes the colliding entry from the map (rather than picking arbitrarily) so the validator skips voting on bounties for that GitHub account, instead of voting for the wrong hotkey

```python
eligible_miners: dict[str, str] = {}
for eval_ in miner_evaluations.values():
    if not eval_.github_id or eval_.github_id == '0' or not eval_.is_eligible:
        continue
    if eval_.github_id in eligible_miners:
        bt.logging.error(
            f'Duplicate github_id={eval_.github_id} across hotkeys: '
            f'{eligible_miners[eval_.github_id][:12]}... vs {eval_.hotkey[:12]}... '
            f'— removing from eligible_miners to prevent ambiguous bounty vote'
        )
        eligible_miners.pop(eval_.github_id, None)
        continue
    eligible_miners[eval_.github_id] = eval_.hotkey
```

This converts a silent failure into a logged, correctable one. The validator will skip voting on the affected bounty, surfacing the duplicate to the operator without misallocating funds.

## Test Plan

- Unit test: build `miner_evaluations` with two entries sharing `github_id`, run the forward pass through the relevant function, assert `eligible_miners` does not contain that `github_id` and that an `error`-level log was emitted.
- Existing tests should still pass since no current test asserts arbitrary-pick behavior.

## Related Issues

- #533, #596 (upstream duplicate-penalty correctness; fixed but defense-in-depth value remains)
- #380, #544 (preservation of `github_id` during penalty)
- #782, #784 (EMA propagation of the penalty)

This issue is in a separate code path from all of the above — the issue-bounty voting forward, not the OSS scoring forward.

Tet-9:
Fixes #895

## Problem
`eligible_miners` in `issue_competitions/forward.py` was built via a
dict comprehension keyed by `github_id`. When two eligible
`MinerEvaluation` entries share the same `github_id`, the second
silently overwrites the first with a non-deterministic winner (Python
dict iteration order over `.values()`).

The surviving hotkey is then used at line 137 to vote-pay the bounty
solver. If the surviving hotkey is not the actual PR author, validators
vote for the wrong miner — or if `get_miner_coldkey()` fails on the
surviving hotkey, the validator skips voting and the bounty becomes
stuck `Active` indefinitely.

This is a defense-in-depth fix: the upstream
`detect_and_penalize_miners_sharing_github` prevents most duplicates,
but race conditions between PAT broadcast and the penalty pass, plus
historical state-divergence bugs (#533, #596), mean duplicates can
still reach this code path in production.

## Fix
Replaced the silent dict comprehension with an explicit loop that:
1. Logs an `error`-level message identifying both hotkeys when a
   duplicate `github_id` is detected
2. Removes the colliding entry from `eligible_miners` entirely, so
   the validator skips voting on that bounty rather than mis-paying

## Changes
- `gittensor/validator/issue_competitions/forward.py` — replaced
  dict comprehension with duplicate-detecting loop

### DIFF
--- a/gittensor/validator/issue_competitions/forward.py
+++ b/gittensor/validator/issue_competitions/forward.py
@@ -62,11 +62,19 @@ async def issue_competitions(
             bt.logging.success(f'Harvested emissions! Extrinsic: {harvest_result.get("tx_hash", "")}')
 
         # Build mapping of github_id->hotkey for eligible miners only
-        eligible_miners = {
-            eval.github_id: eval.hotkey
-            for eval in miner_evaluations.values()
-            if eval.github_id and eval.github_id != '0' and eval.is_eligible
-        }
+        eligible_miners: dict[str, str] = {}
+        for eval_ in miner_evaluations.values():
+            if not eval_.github_id or eval_.github_id == '0' or not eval_.is_eligible:
+                continue
+            if eval_.github_id in eligible_miners:
+                bt.logging.error(
+                    f'Duplicate github_id={eval_.github_id} across hotkeys: '
+                    f'{eligible_miners[eval_.github_id][:12]}... vs {eval_.hotkey[:12]}... '
+                    f'— removing from eligible_miners to prevent ambiguous bounty vote'
+                )
+                eligible_miners.pop(eval_.github_id, None)
+                continue
+            eligible_miners[eval_.github_id] = eval_.hotkey
         bt.logging.info(f'Issue bounties: {len(eligible_miners)} eligible miners out of {len(miner_evaluations)} total')
         for github_id, hotkey in eligible_miners.items():
             bt.logging.info(f'  Eligible miner: github_id={github_id}, hotkey={hotkey[:12]}...')


igeabdulrahmanikeoluwa-str:
@anderdc , this is on for review 🫡

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W5)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W4)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W1)

anderdc:
`detect_and_penalize_miners_sharing_github` sets `is_eligible=False` for every UID sharing a github_id (`inspections.py:78`) before `issue_competitions` runs (`forward.py:75`), so the `is_eligible` filter on this comprehension already excludes duplicates — there is no path today where two `eligible_miners` entries share a github_id. The 'race between PAT broadcast and penalty pass' doesn't match the synchronous call shape (`forward.py:64-75` are sequential awaits in a single forward). Closing as defensive guard for an impossible scenario.

PR #923, opened 2026-05-01
fix: persist refreshed issue discovery for cached fallback UIDs

### LINKED ISSUE #922: [BUG] Cached fallback UIDs skip persistence after fresh issue discovery refresh
## Summary

When OSS GitHub PR fetching fails, `store_or_use_cached_evaluation()` can replace a miner's current evaluation with a cached evaluation and return that UID in `cached_uids`.

Later in the same forward pass, mirror issue discovery can still run successfully and refresh that cached evaluation's issue-discovery fields. However, final DB storage treats the UID as fully cached and skips it through `bulk_store_evaluation(..., skip_uids=cached_uids)`, so the fresh issue-discovery fields are not persisted.

This can make emissions use current-round issue-discovery results while dashboards / DB rows keep stale issue-discovery values from a previous stored evaluation.

## Root Cause

Cache fallback marks restored UIDs to skip DB storage because their cached OSS PR data was already stored:

- `neurons/validator.py`
- `store_or_use_cached_evaluation()`

The forward pass then runs mirror issue discovery after cache fallback:

- `gittensor/validator/forward.py`
- `issue_discovery()`

Mirror issue discovery mutates top-level `MinerEvaluation` issue fields, including:

- `issue_discovery_score`
- `issue_token_score`
- `issue_credibility`
- `is_issue_eligible`
- `total_solved_issues`
- `total_valid_solved_issues`
- `total_closed_issues`
- `total_open_issues`

But final storage skips the whole UID:

- `gittensor/validator/forward.py`
- `bulk_store_evaluation(..., skip_uids=cached_uids)`

## Expected Behavior

Cached OSS fallback should not prevent successfully refreshed issue-discovery fields from being persisted.

If mirror issue discovery fetches a miner's issues successfully, the aggregate miner evaluation row should be updated even when OSS PR data came from cache.

This should also apply when the fresh issue-discovery result is zero/no issues, because that may need to clear stale nonzero issue-discovery fields in the DB.

## Actual Behavior

A UID restored from cache is skipped entirely during DB storage, even after fresh mirror issue discovery updates its evaluation.

As a result:

- current-round emissions can include fresh issue-discovery rewards
- persisted DB/dashboard state can keep stale issue-discovery fields
- successful zero/no-issue refreshes cannot clear stale issue-discovery values

## Why This Matters

After mirror issue discovery was introduced, issue discovery can run independently of the legacy GitHub PR fetch path.

During a GitHub API outage or PR-fetch failure:

1. OSS scoring may fall back to cached evaluation data.
2. Mirror issue discovery may still succeed and compute current issue-discovery fields.
3. Rewards/emissions use the refreshed in-memory values.
4. DB storage skips the cached UID and never persists those refreshed fields.

This creates a consistency gap between emitted rewards and stored validator results.

This issue is specifically about fresh mirror issue-discovery mutation after cached OSS fallback being skipped from DB persistence.

## Suggested Fix

Separate "skip re-storing cached OSS PR/file data" from "skip refreshing the aggregate miner evaluation row."

One possible approach:

- have mirror issue discovery return UIDs whose issue fetch completed successfully
- intersect those UIDs with `cached_uids`
- for that intersection, persist only the miner identity / aggregate `miner_evaluations` row
- keep skipping cached PR, issue, and file-change payload re-storage
- do not refresh cached UIDs when mirror issue fetch fails or issue discovery did not run

Regression tests should cover:

- cached UID with successful nonzero issue-discovery refresh is persisted
- cached UID with successful zero/no-issue refresh clears stale issue-discovery fields
- cached UID with mirror fetch failure remains skipped
- cached PR/file payloads are not re-stored for evaluation-row-only refresh

Daedalus-Icarus:
## Summary

Closes #922.

Cached OSS fallback UIDs were skipped entirely during DB storage, even when mirror issue discovery later refreshed their top-level issue-discovery fields in the same forward pass.

This change separates cached PR/file payload skipping from aggregate evaluation-row persistence:

- mirror issue discovery now returns UIDs whose issue fetch completed successfully
- successful empty/no-relevant issue responses clear stale issue-discovery fields
- cached UIDs with refreshed issue discovery are written through an evaluation-row-only storage path
- cached PR, issue, and file-change payloads remain skipped

## Why

A GitHub PR-fetch failure can restore a cached evaluation and mark the UID as cached. Mirror issue discovery can still succeed afterward and mutate fields such as:

- `issue_discovery_score`
- `issue_token_score`
- `issue_credibility`
- `is_issue_eligible`
- `total_solved_issues`
- `total_valid_solved_issues`
- `total_closed_issues`
- `total_open_issues`

Before this fix, `bulk_store_evaluation(..., skip_uids=cached_uids)` skipped the whole UID, so emissions could use fresh in-memory issue-discovery values while the DB/dashboard kept stale values.

## Fix

- Thread issue-discovery refreshed UIDs through `forward()`.
- Compute cached UIDs that need only an aggregate evaluation-row refresh.
- Add `DatabaseStorage.store_miner_evaluation()` to persist miner identity + `miner_evaluations` row without re-storing cached PR/file data.
- Extend `bulk_store_evaluation()` with `evaluation_only_uids`.

## Testing

- `uv run pytest tests/ -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pyright`
- `uv run pre-commit run --files gittensor/validator/forward.py gittensor/validator/issue_discovery/mirror_scan.py gittensor/validator/utils/storage.py neurons/validator.py tests/validator/issue_discovery/test_mirror_scan.py tests/validator/test_cached_issue_discovery_storage.py`

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -67,7 +67,7 @@ async def forward(self: 'Validator') -> None:
         )
 
         # 2. Score issue discovery
-        issue_rewards = await issue_discovery(
+        issue_rewards, issue_refreshed_uids = await issue_discovery(
             miner_evaluations, master_repositories, programming_languages, token_config, miner_uids
         )
 
@@ -75,7 +75,12 @@ async def forward(self: 'Validator') -> None:
         await issue_competitions(self, miner_evaluations)
 
         # 4. Store all evaluations to DB (includes issue discovery fields)
-        await self.bulk_store_evaluation(miner_evaluations, skip_uids=cached_uids)
+        evaluation_only_uids = issue_refreshed_cached_uids(cached_uids, issue_refreshed_uids)
+        await self.bulk_store_evaluation(
+            miner_evaluations,
+            skip_uids=cached_uids,
+            evaluation_only_uids=evaluation_only_uids,
+        )
 
         # 5. Blend 4 emission pools into final rewards
         rewards = blend_emission_pools(oss_rewards, issue_rewards, miner_uids)
@@ -117,7 +122,7 @@ async def issue_discovery(
     programming_languages: Dict,
     token_config,
     miner_uids: set[int],
-) -> np.ndarray:
+) -> Tuple[np.ndarray, Set[int]]:
     """Score issue discovery and return normalized rewards array.
 
     Mirror-only path: uses ``MirrorClient.get_miner_issues`` with authoritative
@@ -127,14 +132,18 @@ async def issue_discovery(
     solver attribution on busy issues, and non-mirror repos are deliberately
     left out of issue discovery entirely until their mirror_enabled flag flips.
 
-    Returns numpy array of normalized issue discovery rewards (sorted by UID).
+    Returns normalized issue discovery rewards (sorted by UID) plus UIDs whose
+    issue-discovery fields were refreshed from a successful mirror fetch.
     """
     mirror_repos: Dict[str, RepositoryConfig] = {
         name: cfg for name, cfg in master_repositories.items() if cfg.mirror_enabled
     }
 
+    refreshed_uids: Set[int] = set()
     if mirror_repos:
-        await run_mirror_issue_discovery(miner_evaluations, mirror_repos, programming_languages, token_config)
+        refreshed_uids = await run_mirror_issue_discovery(
+            miner_evaluations, mirror_repos, programming_languages, token_config
+        )
     else:
         bt.logging.info('No mirror-enabled repos — issue discovery skipped for this round')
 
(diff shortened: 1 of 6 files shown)


anderdc:
The divergence here is a one-round dashboard lag for UIDs stuck in OSS cache fallback while mirror succeeds — no on-chain, emissions, or bounty-payment impact (those all read in-memory values). With the move to mirror-enabled repos as the default scoring path, the GitHub-PR-fetch fallback that produces `cached_uids` is on the way out, and this whole code path goes with it. Closing as fix-larger-than-the-problem in a path being decommissioned.

PR #933, opened 2026-05-02
fix: use loading_context in --json commands to prevent spinner crash on non-UTF-8 terminals

### LINKED ISSUE #898: [Bug] gitt commands with --json still run the spinner — crashes on non-UTF-8 terminals, leaks Rich frames into stdout
## What's wrong

## Why it bites

```
$ gitt m check
```

```
```

## The fix

```python
# Before
    ...

# After
    ...
```

## Why it's not a duplicate

## Tests added

## Coordination

## Labels

`bug`

597226617:
## Summary

Fixes #898

## Problem

Five CLI commands accept `--json` but still start the Rich spinner via `console.status()`. The spinner uses Braille Unicode characters (`⠇`, `⠏`, …) which crash on non-UTF-8 terminals:

```
UnicodeEncodeError: 'latin-1' codec can't encode character '\u2807'
```

Additionally, spinner frames leak into stdout when terminal buffer is copied, corrupting JSON output.

## Fix

Route all five commands through the existing `loading_context()` helper, which returns `nullcontext()` in JSON mode (no spinner, no output). Mechanical swap — the helper already exists and is imported from `helpers.py`.

**Changed files:**
- `view.py`: 4 sites (`issues_list`, `bounty_pool`, `pending_harvest`, `admin_info`)
- `vote.py`: 1 site (`vote_list_validators`)

**Unchanged:** `vote solution` and `vote cancel` in `vote.py` (no `--json` flag)

### DIFF
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -25,6 +25,7 @@ from .helpers import (
     colorize_status,
     console,
     emit_json,
+    loading_context,
     format_alpha,
     handle_exception,
     print_network_header,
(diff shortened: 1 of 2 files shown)


JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W6)

JesesePU:
I've analyzed this issue and prepared a solution draft. If a fix is needed, I can submit a PR. - Hermes Worker (W7)

anderdc:

PR #936, opened 2026-05-02
test(cli): cover miner CLI subtensor endpoint resolution

kyan9400:
## Problem

`gitt miner post` / `gitt miner check` resolve the subtensor endpoint via `_resolve_endpoint`, combining explicit `--rpc-url`, `--network`, `~/.gittensor/config.json`, and the default finney endpoint.

That behavior had no direct unit coverage, making regressions easy to miss.

## Change

Adds focused pytest coverage for miner endpoint resolution, including:

- explicit `rpc_url` priority
- named network mapping
- unknown endpoint passthrough
- config file fallback
- config `ws_endpoint` priority
- default finney behavior

The tests use `tmp_path` and `monkeypatch` for `HOME`, and do not require network, wallets, Docker, or chain calls.

## Testing

```bash
uv run ruff check tests/cli/test_miner_endpoint_resolution.py
uv run pytest tests/cli/test_miner_endpoint_resolution.py -q

### DIFF
--- a/tests/cli/test_miner_endpoint_resolution.py
+++ b/tests/cli/test_miner_endpoint_resolution.py
@@ -1,0 +1,76 @@
+"""Unit tests for miner CLI `_resolve_endpoint` (no network, no chain, no wallet)."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+from gittensor.cli.miner_commands.helpers import _resolve_endpoint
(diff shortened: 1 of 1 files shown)


kyan9400:
Hi, just checking if there is anything I should adjust here. This is tests-only and local ruff/pytest passed. I can rebase if preferred.

anderdc:
We do not accept pure test PRs unless they have very good reason. Closing.

PR #939, opened 2026-05-02
fix: transient GitHub /user failures now trigger cache fallback instead of skipping evaluation

### LINKED ISSUE #931: [Bug] Transient GitHub /user failures bypass evaluation cache fallback
## Description

The validator has an evaluation cache for GitHub fetch failures, but transient failures in the initial GitHub identity lookup bypass that fallback.

During scoring, `validate_github_credentials()` calls `get_github_id(pat)`, which queries GitHub `/user`. If `/user` times out, returns 5xx, or returns unparsable JSON after retries, `get_github_id()` returns `None`. That is converted into `failed_reason = "No Github id found for miner {uid}'s PAT"`.

Evaluations with `failed_reason` are skipped by `store_or_use_cached_evaluation()` (`neurons/validator.py`), so a miner with a previous successful cached evaluation can still receive zero OSS score for the round.

This happens before legacy or mirror PR scoring starts, so no current-cycle successful work is being discarded.

## Steps to Reproduce

1. Have a miner with a valid stored PAT and a previous successful evaluation in `MinerEvaluationCache` (so `pat_storage` has a non-zero `github_id` and the in-memory cache has a matching entry for `(uid, hotkey, github_id)`).
2. During the next scoring round, make GitHub `/user` fail transiently for that PAT — for example timeout, 5xx, 429, 403-rate-limited, or unparsable JSON — for all 6 retries inside `get_github_id`.
3. Let `validate_response_and_initialize_miner_evaluation()` run inside `evaluate_miners_pull_requests` in `gittensor/validator/oss_contributions/reward.py`.
4. Observe `miner_eval.failed_reason == "No Github id found for miner {uid}'s PAT"`.
5. In `store_or_use_cached_evaluation()`, the `if miner_eval.failed_reason is not None: continue` branch skips the miner, so `MinerEvaluationCache.get(uid, hotkey, github_id)` is never called.

## Expected Behavior

Transient GitHub `/user` failures should not be treated the same as invalid/revoked PATs.

If the validator has a matching stored `github_id` (recorded when the PAT was originally accepted) and a cached evaluation for the same `(uid, hotkey, github_id)`, the existing cache fallback in `store_or_use_cached_evaluation` should be able to run.

Permanent auth errors (401, 403 non-rate-limited, 404) should still fail closed.

## Actual Behavior

The transient `/user` failure becomes a validation failure. `failed_reason` is populated, `github_pr_fetch_failed` is not set, `should_use_cache_fallback` is never checked, and the miner receives zero OSS score for that round despite having a valid stored `github_id` and a cached evaluation.

## Environment

- OS: N/A
- Python version: N/A (>=3.12 per `pyproject.toml`)
- Commit/Version: current `test` branch (tip `d51c4de`)

## Additional Context

This is different from per-PR file-content fetch failures (#645 / #880) and from GraphQL PR-list PAT scope failures (#904 / #905):

- #645 / #880 are about per-PR REST file-content fetches — downstream of scoring.
- #904 / #905 are about GraphQL PR-list auth/scope errors wrongly *enabling* cache fallback.
- This issue is about REST `/user` transient failures wrongly *blocking* cache fallback, and happens before any PR fetch — so using the existing cache fallback here does not discard any partially scored current-cycle PRs.

A safe fix should still fail closed for permanent auth errors such as 401/403 invalid PATs, while distinguishing transient `/user` failures from revoked credentials. Concretely:

- Split `get_github_id` into an identity lookup that reports `VALID` / `INVALID_AUTH` / `TRANSIENT_FAILURE`.
- On `TRANSIENT_FAILURE` with a stored `github_id` pinned at PAT-broadcast time, reuse that id on the `MinerEvaluation` and set `github_pr_fetch_failed = True` so `should_use_cache_fallback` becomes True and the existing `MinerEvaluationCache.get(uid, hotkey, github_id)` lookup matches.
- Short-circuit `evaluate_miners_pull_requests` before legacy and mirror PR loading on transient identity failure so `total_prs` stays 0.
- Keep returning `failed_reason` for `INVALID_AUTH` (401, 403 non-rate-limited, 404).

597226617:
## Summary

Fixes #931

## Problem

When GitHub `/user` times out, returns 5xx, or produces unparsable JSON, `get_github_id()` returned `None` — indistinguishable from an invalid PAT. `validate_github_credentials` then set `failed_reason`, which caused `store_or_use_cached_evaluation` to skip the miner entirely, even when a valid cached evaluation existed.

Result: a transient network hiccup wipes out a miner's evaluation for that round.

## Fix

1. `get_github_id()` returns `(user_id, is_transient)` tuple to distinguish network/5xx/parse failures from permanent auth failures
2. `validate_github_credentials()` propagates the `is_transient` flag
3. `inspect_miner()`: on transient failure, sets `github_pr_fetch_failed=True` and `should_use_cache_fallback=True` instead of `failed_reason`, activating the existing cache fallback path
4. `pat_handler.py`: adapts to new 3-tuple return (ignores `is_transient` — PAT registration should not use stale cache)

## Testing

All modified files pass syntax validation (`ast.parse`).

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from math import ceil
-from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Tuple
 
 from gittensor.utils.utils import backoff_seconds, parse_repo_name
 
@@ -222,17 +222,20 @@ def get_session(token: str) -> requests.Session:
     return _session_cache[key]
 
 
-def get_github_id(token: str) -> Optional[str]:
+def get_github_id(token: str) -> Tuple[Optional[str], bool]:
     """Get GitHub numeric user id (as string) using a PAT.
 
     Args:
         token (str): GitHub personal access token.
 
     Returns:
-        Optional[str]: Numeric user id as a string, or None if it cannot be determined.
+        Tuple[Optional[str], bool]: (user_id, is_transient_failure).
+        user_id is the numeric GitHub user id as string, or None on failure.
+        is_transient_failure is True when the failure is due to network/5xx/parse
+        errors (not an invalid PAT), so callers can use cached data as fallback.
     """
     if not token:
-        return None
+        return None, False
 
     session = get_session(token)
 
@@ -245,10 +248,10 @@ def get_github_id(token: str) -> Optional[str]:
                     user_data: Dict[str, Any] = response.json()
                 except Exception as e:  # pragma: no cover
                     bt.logging.warning(f'Failed to parse GitHub /user JSON response: {e}')
-                    return None
+                    return None, True
 
                 user_id = user_data.get('id')
-                return str(user_id) if user_id is not None else None
+                return (str(user_id), False) if user_id is not None else (None, False)
 
             bt.logging.warning(
                 f'GitHub /user request failed with status {response.status_code} (attempt {attempt + 1}/6)'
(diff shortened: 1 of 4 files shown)


anderdc:

PR #944, opened 2026-05-03
fix: surface decode failures in _read_issues_from_child_storage instead of silently returning empty list 

### LINKED ISSUE #872: [Bug] `gitt admin info` returns zero issues silently when child storage decoding fails on any item
## Description

`cli/issue_commands/admin.py:info` calls `_read_issues_from_child_storage`, which catches decoder errors
and returns `[]` on the truncation/corruption guard described in #47. The CLI then prints "0 issues found" without distinguishing between "no issues registered" and "decoder failed on all items".

## Steps to Reproduce

1. Cause a single corrupted child-storage entry on chain (test environment).
2. Run `gitt admin info`.
3. Observe "0 issues" — the operator has no way to tell whether the chain genuinely has 0 issues or whether decoding silently failed.

## Expected Behavior

Surface the decode-failure condition: `gitt admin info` should print `WARNING: storage decoder reported X failures; results may be incomplete` when the helper returned an empty list due to errors.

## Actual Behavior

Silent zero; operator misled.

## Additional Context

- Suggested: have `_read_issues_from_child_storage` return a small dataclass `(issues, decode_errors)` so callers can
  distinguish.

Tet-9:
Closes #872

## Problem

`_read_issues_from_child_storage` in `gittensor/cli/issue_commands/helpers.py` catches all decode exceptions per-item and silently continues. When every item fails to decode, the function returns `[]` with no indication of failure. All call sites in `view.py` and `helpers.py` then print "0 issues found" — identical output to a genuinely empty contract. Operators have no way to distinguish between:

- The contract having zero registered issues
- The decoder failing on every storage entry due to corruption or schema mismatch

This silent failure masks real on-chain data and can send operators down the wrong debugging path.

### `gittensor/cli/issue_commands/helpers.py`

- Changed return type of `_read_issues_from_child_storage` from `List[Dict[str, Any]]` to `Tuple[List[Dict[str, Any]], int]` — the second element is the count of decode failures
- All early-exit return paths now return `([], 0)`
- Added `decode_errors` counter that increments on every caught decode exception
- Final return is now `(issues, decode_errors)`
- Updated the internal call site at line 900 to unpack the tuple and print a `WARNING` when `decode_errors > 0`

### `gittensor/cli/issue_commands/view.py`

- Updated both call sites (lines 199 and 255) to unpack `(issues, decode_errors)` and print:
- whenever `decode_errors > 0`

## Behavior After Fix

- Zero issues + zero decode errors → normal "0 issues found" output, no warning
- Zero issues + decode errors → warning printed, operator knows decoder failed
- Some issues + decode errors → issues displayed + warning that results are partial

## Files Changed

- `gittensor/cli/issue_commands/helpers.py` — return tuple, track decode_errors, warn at internal call site
- `gittensor/cli/issue_commands/view.py` — unpack tuple, surface warning at both call sites

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -746,7 +746,9 @@ def _read_contract_packed_storage(substrate, contract_addr: str, verbose: bool =
         return None
 
 
-def _read_issues_from_child_storage(substrate, contract_addr: str, verbose: bool = False) -> List[Dict[str, Any]]:
+def _read_issues_from_child_storage(
+    substrate, contract_addr: str, verbose: bool = False
+) -> Tuple[List[Dict[str, Any]], int]:
     """
     Read all issues from contract child storage.
 
@@ -770,14 +772,14 @@ def _read_issues_from_child_storage(substrate, contract_addr: str, verbose: bool
     if not child_key:
         if verbose:
             console.print(f'[dim]Debug: Cannot read issues - no child storage key for {contract_addr}[/dim]')
-        return []
+        return [], 0
 
     # First, read packed storage to get next_issue_id
     packed_storage = _read_contract_packed_storage(substrate, contract_addr, verbose)
     if not packed_storage:
         if verbose:
             console.print('[dim]Debug: Cannot read issues - packed storage read failed[/dim]')
-        return []
+        return [], 0
 
     next_issue_id = packed_storage.get('next_issue_id', 1)
     if verbose:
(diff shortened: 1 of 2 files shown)


anderdc:
This requires every per-item decode to fail simultaneously — realistically only reachable when the operator's CLI version is behind a contract storage-layout upgrade, which the operator already knows about (they upgraded). In steady state it fires approximately never, and during a rollout window the operator figures out the CLI mismatch from the missing issues themselves within a minute. The permanent tuple-return signature change on a public helper plus three duplicate warning prints isn't worth a one-time ~30-second debugging save. Closing as observability polish for an effectively-unreachable scenario.

PR #947, opened 2026-05-03
feat: Be able to run scoring pipeline locally

### LINKED ISSUE #946: feat: Be able to run scoring pipeline locally
## Summary and motivation

Currently any changes to the validation pipeline aren't verifiable e2e. Currently we only write synthetic unit tests, but during development process you usually want to ensure manually by testing the whole feature e2e.

It would be nice to have an easy ability to run the scoring process locally, excluding the subnet/wallet/db harness. Something like `gitt miner score --pat ...`, which would start a validation pipeline against a solo miner locally. This allows easily test most of scenarios and will open a lot of further opportunities:

- Extremely easy to check how changes to the scoring algorithm affect live results.
- Safer contributions, as you know that it definitely won't break anything.
- Can be extended to support more debugging features (like `--profile` to trace the whole pipeline)

seroperson:
Closes #946

## Summary

This PR introduces a new sub-command, which allows to run scoring pipeline (like validator does) locally, but without subnet/db/etc harness. It's primarily intended for debugging purposes, as currently there is no easy way to ensure your changes work e2e. Unit testing is good, but having no ability to run something manually makes people actually skip this method of testing.

Now you can, for example, do changes to the scoring algorithm, and check "live" results right after.

```sh
# or without --pat, using GITTENSOR_MINER_PAT variable
gitt miner score --pat ... --json-output --log-level info
```

The implementation just runs what validator runs, but with subnet/db/wallet things stubbed.

I've placed this command under `gitt miner score`, but we can also move it somewhere or even hide.

https://github.com/user-attachments/assets/f3384824-9cad-4ed1-8bd0-c84266e9d0f9

On the video we also see how the very first request responds with 502 (and it's reproducible for my PAT), likely some constant should be tweaked or something. Nobody knows what else this feature can also uncover =)

## Test plan

Covered with some tests, but strictly there is no much really new testable code.

### DIFF
--- a/gittensor/cli/miner_commands/__init__.py
+++ b/gittensor/cli/miner_commands/__init__.py
@@ -7,12 +7,14 @@ Command structure:
     gitt miner (alias: m)     - Miner management commands
         post                     Broadcast GitHub PAT to validators
         check                    Check how many validators have your PAT
+        score                    Score GitHub entities through the production pipeline
 """
 
 import click
 
 from .check import miner_check
 from .post import miner_post
+from .score import score_command
 
 
 @click.group(name='miner')
@@ -23,12 +25,14 @@ def miner_group():
     Commands:
         post     Broadcast your GitHub PAT to validators
         check    Check how many validators have your PAT stored
+        score    Run the validator scoring pipeline end-to-end for a single miner
     """
     pass
 
 
 miner_group.add_command(miner_post, name='post')
 miner_group.add_command(miner_check, name='check')
+miner_group.add_command(score_command, name='score')
 
 
 def register_miner_commands(cli):
--- a/gittensor/cli/miner_commands/score.py
+++ b/gittensor/cli/miner_commands/score.py
@@ -8,17 +8,21 @@ Stubs axon, wallet, subtensor and DB.
 from __future__ import annotations
 
 import dataclasses
+import json
+import os
 import sys
-from datetime import datetime
+from contextlib import contextmanager
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Dict, NoReturn, Optional, Set, Tuple, cast
-from unittest.mock import patch
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from types import SimpleNamespace
+from typing import TYPE_CHECKING, Any, Dict, Iterator, NoReturn, Optional, Set, Tuple, cast
 
 import click
 from rich.console import Console
 from rich.table import Table
 
-from gittensor.cli.issue_commands.helpers import emit_json, get_github_pat
+from gittensor.cli.issue_commands.helpers import emit_json
 from gittensor.cli.miner_commands.helpers import _error
 
 if TYPE_CHECKING:
@@ -36,7 +40,7 @@ def _die(msg: str, json_mode: bool) -> NoReturn:
 
 
 def _resolve_pat(cli_pat: Optional[str], json_mode: bool) -> str:
-    pat = cli_pat or get_github_pat()
+    pat = cli_pat or os.environ.get('GITTENSOR_MINER_PAT')
     if not pat:
         _die('--pat flag or GITTENSOR_MINER_PAT environment variable is required.', json_mode)
     return pat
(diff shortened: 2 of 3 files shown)


anderdc (changes requested):
Premise looks good. A few simplifications before merge — none individually large, but together they trim ~80 LOC and remove dead surface.

1. **Drop the dead stub methods.** `gittensor/cli/miner_commands/score.py:55-60` declares `_StubValidator.bulk_store_evaluation` and `update_scores`, but neither is reached. `score_command` calls `oss_contributions` / `issue_discovery` / `blend_emission_pools` directly, never `forward()` — those two are only invoked from inside `forward()` (`gittensor/validator/forward.py:78,83`). The only `self.*` calls in this code path are `self.metagraph.hotkeys[uid]` and `self.store_or_use_cached_evaluation(...)` (`gittensor/validator/oss_contributions/reward.py:126,148`). Trim the stub to those two members.

2. **Drop `_SparseHotkeys`.** `score.py:33-37` plus `_StubMetagraph` at :40-42 together exist to make `metagraph.hotkeys[unknown_uid]` return `'__unused__'`. Nothing iterates anything other than `miner_uids = {_DEV_UID}`, so the fallback is for a case that never occurs. Replace with one line — `self.metagraph = SimpleNamespace(hotkeys={uid: hotkey})` — and delete both classes.

3. **Replace `unittest.mock.patch` with a tempfile override.** `score.py:11` imports `unittest.mock.patch` into a production CLI file purely to reroute `pat_storage.load_all_pats` at :282. Cleaner: write the snapshot to a tempfile and override `pat_storage.PATS_FILE` for the duration of the call (already module-level, no test-utility import). That also exercises the real read path, including the `_read_file` corruption-tolerance branch you'd otherwise skip.

4. **Lean on `emit_json(default=str)` for the trivial coercions.** `score.py:103-110` `_jsonable()` handles `Enum`, `datetime`, and `set`. `emit_json` (`gittensor/cli/issue_commands/helpers.py:209`) already passes `default=str`, which covers all three for JSON serialization. Keep an explicit `Enum -> .value` if you need the underlying value rather than the repr; otherwise the helper can shrink.

5. **The 502 you flagged in the PR body needs a follow-up before merge.** "The very first request responds with 502 (and it's reproducible for my PAT)" — landing this command while it reproduces a 502 on first run means new contributors will hit it on day one. Either name the constant you suspect and tune it here, or open a tracking issue and link it from the command's `--help`.

6. **Consider replacing the introspection-based `_project` with `dataclasses.asdict` + post-process.** `_project` at :112-117 reimplements the recursive walk that `dataclasses.asdict` already does. With `asdict(miner_eval)` you get the nested-dataclass walk for free, then drop `github_pat` / bulky list fields and merge in `_EVAL_PROPERTIES`. Optional — current code works.

7. **`_drain_logs` magic timing.** :235-242 has a 2-second deadline + 0.1s grace tick. That smells like a workaround for bittensor's async log writer. Worth either documenting *why* the grace tick is needed (race between `queue.empty()` and worker write?) or moving the helper to a shared module so other CLI commands can adopt it without copy-paste.

seroperson:
Addressed all points, except:

- p.5: addressed in https://github.com/entrius/gittensor/pull/1094 to not widen the scope here
- p.6: `asdict` still needs to handle enums, still needs `_EVAL_PROPERTIES`, doesn't work for mirror PRs etc etc etc, so I assume it doesn't worth.
- p.7: added docstring. Did not move it to a shared module as no other CLI command currently needs it and it feels like a very specific case actually. I can move if you insist =)

anderdc (changes requested):
Legacy PAT scoring is being deprecated soon — please rescope this to mirror + issue discovery only so the dev tool matches the post-deprecation pipeline (and we don't have to strip legacy plumbing out of it again later).

Drop:
- The legacy/mirror dual walk in `_serialize_evaluation` (score.py:131-145) — render only the mirror lists.
- The `source` tag on `_serialize_pr` (score.py:120-124) and the `Source` column in `_render_table`.
- `mirror_scored_pr_to_legacy_pull_request` import + adapter call (score.py:130-134).
- Filter `master_repositories` to `mirror_enabled=True` before calling `oss_contributions` so the legacy branch in reward.py:89 never fires for this command.

Keep:
- Mirror PR scoring + issue discovery + blend.
- Table / JSON renderers, `_drain_logs`, `_apply_log_level`, `_StubValidator`, `_override_pats_file` (PAT is still needed for github_id resolution via `validate_github_credentials_result`).

While you're in there: add a test that exercises mirror-PR rendering end-to-end. Current `test_miner_score.py` only covers the empty-PR-list path and `failed_reason` — `_serialize_evaluation`'s populated-list branch has zero assertions on it. One test that builds a populated `mirror_merged_prs` (or whatever the post-rescope shape is), runs through `_serialize_evaluation`, and asserts a non-trivial PR row in the JSON / table output catches silent serializer regressions.

anderdc (approved):

PR #917, opened 2026-04-30
refactor: bundle of small refactors

ebios-star:
## Summary

Per @anderdc's review, consolidating my open small refactor PRs into this one. #864 and #873 are mid-review and remain separate as previously requested. (Note: #873 has since merged — its work is built on by the last commit here.)

Each commit is preserved as a separate revision so the bundle can be reviewed commit-by-commit or squash-merged.

### Commits in this bundle (oldest → newest)

1. **#917** — `refactor: use BASE_GITHUB_API_URL constant in cli/issue_commands/helpers.py`
2. **#942** — `refactor: use GITHUB_HTTP_TIMEOUT_SECONDS at the three remaining timeout=15 sites`
3. **#943** — `refactor: remove unused REPO_SCAN_* constants`
4. **#948** — `refactor: inline single-use get_github_pat and is_comment_node helpers`
5. **#949** — `refactor: inline single-use _calculate_base_score adapter into score_mirror_pr`
6. **#952** — `refactor: centralize ink! issues-mapping selector as ISSUES_MAPPING_ROOT_KEY`
7. **#953** — `refactor: remove unused U32_MAX constant`
8. **#954** — `refactor: remove unused GITHUB_DOMAIN constant`
9. **#955** — `refactor: normalize __version__ import in neurons/validator.py`
10. **#1015** — `refactor: move MAX_ISSUE_ID to gittensor/constants.py and use it on the validator side` (follow-up to merged #873)

### Aggregate change

- 10 files changed, **+28 / -52 lines** (net **-24 lines**).
- No behaviour change in any commit.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- `pytest tests/` on the merged bundle — all **754 tests pass**.

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #949, opened 2026-05-03
refactor: inline single-use _calculate_base_score adapter

ebios-star:
## Summary

`_calculate_base_score` in [gittensor/validator/oss_contributions/mirror/scoring.py](gittensor/validator/oss_contributions/mirror/scoring.py) is described in its own docstring as a *"Thin wrapper: run the shared helper and copy fields onto ScoredMirrorPR"* and has exactly one caller (`score_mirror_pr`).

The wrapper mutates **seven** fields on `scored` while returning only `base_score`, hiding most of its side effects behind a misleading return signature — the only visible mutation at the call site is `scored.base_score = ...`. Inlining moves the field-copying out into the caller alongside that existing assignment, so all eight mutations land on one block in `score_mirror_pr`.

The shared helper `calculate_base_score_for_pr_files` (which actually performs the computation) is unaffected and still callable from any future non-mutating consumer.

`grep -rn _calculate_base_score` confirms no other call sites or test references in `gittensor/`, `neurons/`, or `tests/`.

Net: **-20 / +9 lines**, one fewer module-level helper.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- `pytest tests/` — all 726 tests pass.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/oss_contributions/mirror/scoring.py
+++ b/gittensor/validator/oss_contributions/mirror/scoring.py
@@ -144,7 +144,15 @@ def score_mirror_pr(
 
     file_changes, file_contents = mirror_files_to_legacy(pr.repo_full_name, pr.pr_number, files)
 
-    scored.base_score = _calculate_base_score(scored, file_changes, file_contents, programming_languages, token_config)
+    result = calculate_base_score_for_pr_files(file_changes, file_contents, programming_languages, token_config)
+    scored.token_score = result.token_score
+    scored.structural_count = result.structural_count
+    scored.structural_score = result.structural_score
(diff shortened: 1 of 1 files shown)


ebios-star:
@anderdc whenever you have a moment, this is a small inline of a single-use adapter that the wrapper itself describes as a thin wrapper. Happy to revise or scope differently if you'd prefer. Thanks!

anderdc:
Closing — please consolidate into #917.

ebios-star:
Consolidated into #955 per @anderdc's request.

PR #963, opened 2026-05-04
fix(mirror): fall back to main and gate head_ref on acceptable set

### LINKED ISSUE #958: [Bug] Mirror merged-PR gate skips base-branch validation when default_branch is absent and repo has no additional_acceptable_branches
## Description

For the mirror OSS path, `_should_skip_merged_mirror_pr` builds the acceptable base-branch set as:

`acceptable = ([pr.default_branch] if pr.default_branch else []) + additional`

The base-ref check only runs when `acceptable` is non-empty. If the mirror payload has **no** `default_branch` (older/partial mirror rows are explicitly called out in comments elsewhere) **and** `RepositoryConfig.additional_acceptable_branches` is empty, then `acceptable` is empty and **the base_ref gate is skipped entirely**.

Any merged PR then passes the base-branch rule regardless of `base_ref`, which breaks parity with the legacy path: legacy `should_skip_merged_pr` always has a usable default branch name because it falls back to `'main'` when GraphQL omits `defaultBranchRef`.

So miners can be scored for merges that would be rejected on the legacy path, or that should be rejected when mirror metadata is incomplete.

## Steps to Reproduce

1. Configure a `RepositoryConfig` entry with **no** `additional_acceptable_branches` (or empty list).
2. Use a mirror `MirrorPullRequest` where `default_branch` is `None` or missing in JSON (so `from_dict` leaves it falsy).
3. Set `base_ref` to a branch that would **not** be allowed under normal rules (e.g. not `main`).
4. Run `_should_skip_merged_mirror_pr` / load path so the PR is considered for `mirror_merged_prs`.
5. Observe `should_skip` is `False` (PR accepted) even though the same PR would fail legacy’s default-branch check.

## Expected Behavior

Mirror path should match legacy semantics: there must always be a defined acceptable base set (e.g. fall back to `repo_config` default, mirror default, or at minimum `'main'` like legacy), and PRs merged to non-acceptable bases must be skipped.

## Actual Behavior

When `acceptable` is empty, the `if acceptable and not branch_matches_pattern(...)` branch does not run, so **no** base-branch validation is applied for that PR.

## Environment

- OS: Linux
- Python version: 3.12+
- Commit/Version: current `test` (mirror scoring in `gittensor/validator/oss_contributions/mirror/scoring.py`)

## Additional Context

Relevant code (`gittensor/validator/oss_contributions/mirror/scoring.py`):

- `acceptable = ([pr.default_branch] if pr.default_branch else []) + additional`
- Base check: `if acceptable and not branch_matches_pattern(pr.base_ref or '', acceptable):`

Legacy reference (`gittensor/utils/github_api_tools.py`, `should_skip_merged_pr`): default branch resolves to `'main'` when `defaultBranchRef` is missing, so the acceptable set is rarely empty.

### LINKED ISSUE #959: [Bug] Mirror path skips “acceptable head branch” self-merge gate when repo has no additional_acceptable_branches
## Description

In `_should_skip_merged_mirror_pr`, the check that blocks internal merges where the **source branch is itself an acceptable branch** (legacy: “merging between acceptable branches not allowed”) is guarded by:

`if additional and pr.head_ref and is_same_repo and branch_matches_pattern(pr.head_ref, acceptable):`

`additional` is only `repo_config.additional_acceptable_branches`. When that list is **empty** (the common case: repo only uses the default branch), the condition is false and **the entire head-ref gate is skipped**, even though `acceptable` still includes `pr.default_branch`.

Legacy `should_skip_merged_pr` in `github_api_tools.py` does **not** require extra branches: it always builds `acceptable_branches = [default_branch] + additional_branches` and applies the head-ref pattern check whenever it is a same-repo PR.

So a PR that legacy would reject as “source branch is an acceptable branch” can be **accepted and scored on the mirror path** for repos with no `additional_acceptable_branches`. That is a direct scoring / incentive inconsistency between the two code paths.

## Steps to Reproduce

1. Pick a `RepositoryConfig` with **empty** `additional_acceptable_branches` (or unset).
2. Construct a same-repo merged PR where `head_ref` matches the default acceptable branch pattern (e.g. internal `main` → `main` style case legacy blocks per `should_skip_merged_pr`).
3. Run legacy load gate: `should_skip_merged_pr` → expect skip when head is in acceptable set.
4. Run mirror gate: `_should_skip_merged_mirror_pr` with matching `default_branch` / `head_ref` / same-repo → observe the head-ref block is **not** entered because `additional` is falsy.
5. Observe mirror can admit the PR into scoring while legacy would drop it.

## Expected Behavior

Mirror merged-PR eligibility should match legacy semantics for the head-ref / acceptable-branch anti-gaming rule: the gate should apply whenever there is a non-empty acceptable branch set (at least default branch), not only when `additional_acceptable_branches` is non-empty.

## Actual Behavior

If `additional_acceptable_branches` is empty, mirror never evaluates the head-ref branch gate, so PRs that legacy would skip can still be scored on the mirror path.

## Environment

- OS: Linux
- Python version: 3.12+
- Commit/Version: current `test`

## Additional Context

- Mirror gate: `gittensor/validator/oss_contributions/mirror/scoring.py` (`_should_skip_merged_mirror_pr`, head-ref block prefixed with `if additional and ...`).
- Legacy reference: `gittensor/utils/github_api_tools.py` (`should_skip_merged_pr`, head vs `acceptable_branches` without requiring `additional_branches` to be non-empty).

597226617:
## Summary

Two fixes to restore parity between mirror and legacy branch-gate logic in `_should_skip_merged_mirror_pr`:

### Fix 1 — base_ref fallback (#958)

When both `default_branch` and `additional_acceptable_branches` are empty, the `acceptable` set was empty and the `base_ref` gate was skipped entirely. Now falls back to `["main"]`, matching legacy `should_skip_merged_pr` which always has a usable default branch name (falls back to `"main"` when GraphQL omits `defaultBranchRef`).

### Fix 2 — head_ref guard condition (#959)

The head-ref (self-merge between acceptable branches) gate was guarded by `additional` (only `additional_acceptable_branches`), so repos with no extra branches never evaluated it. Changed the guard to `acceptable`, which always includes `default_branch`, matching legacy behaviour that applies the check whenever there is a non-empty acceptable branch set.

Fixes #958
Fixes #959

### DIFF
--- a/gittensor/validator/oss_contributions/mirror/scoring.py
+++ b/gittensor/validator/oss_contributions/mirror/scoring.py
@@ -203,6 +203,12 @@ def _should_skip_merged_mirror_pr(scored: ScoredMirrorPR, repo_config: Repositor
     additional = repo_config.additional_acceptable_branches or []
     acceptable = ([pr.default_branch] if pr.default_branch else []) + additional
 
+    # Fall back to 'main' when both default_branch and additional are missing
+    # to match legacy behaviour (should_skip_merged_pr falls back to 'main'
+    # when GraphQL omits defaultBranchRef).
+    if not acceptable:
+        acceptable = ['main']
+
     # base_ref check — only enforce when we have an acceptable set to compare against.
     if acceptable and not branch_matches_pattern(pr.base_ref or '', acceptable):
         return True, (f'PR #{pr.pr_number} merged to {pr.base_ref!r} not in acceptable branches={acceptable}')
@@ -211,7 +217,7 @@ def _should_skip_merged_mirror_pr(scored: ScoredMirrorPR, repo_config: Repositor
     # branch. Only applies to same-repo PRs: fork branch names are arbitrary.
     # Falls through when head_ref or head_repo_full_name is missing (older data).
     is_same_repo = pr.head_repo_full_name is not None and pr.head_repo_full_name == pr.repo_full_name
-    if additional and pr.head_ref and is_same_repo and branch_matches_pattern(pr.head_ref, acceptable):
+    if acceptable and pr.head_ref and is_same_repo and branch_matches_pattern(pr.head_ref, acceptable):
         return True, (
             f'PR #{pr.pr_number} source branch {pr.head_ref!r} is itself in '
             f'acceptable branches — merging between acceptable branches not allowed'


anderdc:

PR #967, opened 2026-05-04
perf: optimizing collect_node_signatures function

### LINKED ISSUE #966: perf: unoptimized collect_node_signatures function
## Problem

`gittensor.validator.utils.tree_sitter_scoring.collect_node_signatures` is on the per-file hot path of `score_tree_diff`, which the validator calls for every code-bearing file in every PR it scores. The current implementation has at least two avoidable costs:

- `walk_node(node)`, the whole recursive approach is expensive for such a tight loop. `treesitter` has `TreeCursor`, which is intended specifically for such purposes.
- `node.text.decode('utf-8')`, which is actually unused, so at least we can skip `decode`ing it.

## Motivation

As this function visits every node in every file, even small changes may affect the overall performance. Exactly these changes may affect it significantly, but this has to be checked.

seroperson:
Closes #966

## Summary

This PR introduces set of optimizations to the `collect_node_signatures`:

- Recursive `walk_node` approach is replaced with `TreeCursor` (`tree.walk()`) and a plain `while` loop.
- Leaf `("leaf", node_type, text)` replaced with `("leaf", node_type, bytes)` to strip unnecessary utf-8 decoding.
- Some micro-optimizations, like double `node.text` call, list allocations, and sticking to C calls.

## Benchmark

Benchmarking e2e scoring:

```
  | fixture     | before (HEAD~1)   | after (this PR)   | speedup |
  |-------------|-------------------|-------------------|---------|
  | py-tiny     |     9.61us        |     8.39us        | 1.15x   |
  | py-small    |    87.65us        |    67.23us        | 1.30x   |
  | py-medium   |     2626us        |     2030us        | 1.29x   |
  | py-large    |    10827us        |     8192us        | 1.32x   |
  | rs-tiny     |    22.35us        |    18.24us        | 1.23x   |
  | rs-small    |   122.10us        |    97.90us        | 1.25x   |
  | rs-medium   |     2193us        |     1700us        | 1.29x   |
  | rs-large    |     9024us        |     6527us        | 1.38x   |
  | sh-tiny     |     7.68us        |     6.32us        | 1.21x   |
  | sh-small    |    76.20us        |    63.66us        | 1.20x   |
  | sh-medium   |     1633us        |     1198us        | 1.36x   |
  | sh-large    |     6595us        |     4775us        | 1.38x   |
```

`python -m benchmarks.bench_collect_node_signatures` (run in the branch with actual changes)

[benchmarks/bench_collect_node_signatures.py](https://github.com/user-attachments/files/27351022/bench_collect_node_signatures.py)

## Test plan

Added `tests/validator/test_collect_node_signatures.py` which pins the exact function behavior, and also added more e2e tests to the `tests/validator/test_token_scoring_integration.py` to pin even more the existing scoring behavior (to not break anything in future).

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -82,8 +82,8 @@ def parse_code(content: str, language: str) -> Optional[Tree]:
 
 # Type alias for node signatures
 # Structural: ("structural", node_type)
-# Leaf: ("leaf", node_type, text)
-NodeSignature = Union[Tuple[str, str], Tuple[str, str, str]]
+# Leaf: ("leaf", node_type, text_bytes)
+NodeSignature = Union[Tuple[str, str], Tuple[str, str, bytes]]
 
 
 def collect_node_signatures(
@@ -95,7 +95,7 @@ def collect_node_signatures(
 
     Walks the tree and collects signatures for:
     - Structural nodes: ("structural", node_type) - captures structure without content
-    - Leaf nodes: ("leaf", node_type, text) - captures content for meaningful tokens
+    - Leaf nodes: ("leaf", node_type, text_bytes) - captures content for meaningful tokens
 
     Comments are skipped entirely (score 0).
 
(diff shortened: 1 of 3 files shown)


anderdc (approved):
good stuff

PR #975, opened 2026-05-04
fix: recycle treasury emission share when treasury is disabled or unregistered

### LINKED ISSUE #974: Disabled/unregistered treasury silently drops 15% emissions instead of recycling
**Description:**

In `blend_emission_pools()` ([gittensor/validator/forward.py:179-185](gittensor/validator/forward.py#L179-L185)), the treasury pool only allocates if `ISSUES_TREASURY_UID > 0 and ISSUES_TREASURY_UID in miner_uids`. When that condition is false, the 15% share is **silently dropped** — it isn't added to `recycle_extra` like the other unclaimed pools are.

### When this triggers
- Treasury neuron (UID 111) is deregistered or temporarily missing from the metagraph.
- Treasury is intentionally disabled by setting `ISSUES_TREASURY_UID = RECYCLE_UID` (per the comment at [gittensor/constants.py:215](gittensor/constants.py#L215)).

### Why it matters
- Pools 1 (OSS) and 2 (issue discovery) correctly recycle their share when empty, so the validator always emits ~100% of weight.
- Pool 3 (treasury) does not, so in the disabled/missing case the validator emits only ~85% — 15% is unallocated.
- The docstring at [forward.py:51-56](gittensor/validator/forward.py#L51-L56) implies all four pools sum to 100%, so this is a behavior/documentation mismatch.

### Secondary issue
The guard `ISSUES_TREASURY_UID > 0` only handles the specific case where treasury is set to UID 0. The constants comment says treasury is disabled when `ISSUES_TREASURY_UID == RECYCLE_UID` — these happen to coincide today (`RECYCLE_UID = 0`), but the intent should be expressed directly.

aliangm:
**Description:**

## Summary
- When `ISSUES_TREASURY_UID` is set to `RECYCLE_UID` (treasury disabled per [constants.py:215](gittensor/constants.py#L215)) or the treasury neuron is not registered on the metagraph, the 15% treasury share was silently dropped instead of being recycled.
- Pools 1 (OSS) and 2 (issue discovery) already fall through to `recycle_extra` when empty; pool 3 (treasury) now does the same so total emissions consistently sum to 100%.
- Also replaces the `ISSUES_TREASURY_UID > 0` guard with the more direct `ISSUES_TREASURY_UID != RECYCLE_UID` to match the intent stated in the constants comment.

## Test plan
- [ ] Validator with treasury UID registered: 15% still flows to treasury, recycle pool unchanged.
- [ ] Validator with `ISSUES_TREASURY_UID = RECYCLE_UID`: recycle UID receives 25% + 15% = 40% (plus any unclaimed OSS/issue-discovery share).
- [ ] Validator where treasury UID is not in metagraph: same as above — 15% flows to recycle.
- [ ] Sum of emitted weights equals 1.0 in all three cases.

Fixes #974

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -175,14 +175,16 @@ def blend_emission_pools(
     else:
         recycle_extra += ISSUE_DISCOVERY_EMISSION_SHARE
 
-    # Pool 3: Issue treasury (15% flat to UID 111)
-    if ISSUES_TREASURY_UID > 0 and ISSUES_TREASURY_UID in miner_uids:
+    # Pool 3: Issue treasury (15% flat to ISSUES_TREASURY_UID, recycled if disabled/unregistered)
+    if ISSUES_TREASURY_UID != RECYCLE_UID and ISSUES_TREASURY_UID in miner_uids:
         treasury_idx = sorted_uids.index(ISSUES_TREASURY_UID)
         rewards[treasury_idx] += ISSUES_TREASURY_EMISSION_SHARE
         bt.logging.info(
             f'Treasury allocation: UID {ISSUES_TREASURY_UID} receives '
             f'{ISSUES_TREASURY_EMISSION_SHARE * 100:.0f}% of emissions'
         )
+    else:
+        recycle_extra += ISSUES_TREASURY_EMISSION_SHARE
 
     # Pool 4: Recycle (25% + unclaimed from empty pools)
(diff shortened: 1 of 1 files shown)


anderdc:
These paths don't fire in production — `ISSUES_TREASURY_UID` is hardcoded with no runtime override, and on mainnet UIDs are reused so UID 111 stays in the metagraph. Closing.

PR #977, opened 2026-05-04
fix: bound old_content size in tree-sitter scoring

### LINKED ISSUE #976: `MAX_FILE_SIZE_BYTES` guard skipped for `old_content` in tree-sitter scoring
**Description:**

In `calculate_token_score_from_file_changes()` ([gittensor/validator/utils/tree_sitter_scoring.py:312](gittensor/validator/utils/tree_sitter_scoring.py#L312)), the `MAX_FILE_SIZE_BYTES` size check is applied only to `content_pair.new_content`. When the file passes the check, both old and new versions are parsed inside `score_tree_diff` ([line 190-198](gittensor/validator/utils/tree_sitter_scoring.py#L190-L198)) — but the old version is never bounded.

### Why old_content matters
`score_tree_diff` intentionally scores both additions and deletions ([line 169](gittensor/validator/utils/tree_sitter_scoring.py#L169) docstring), so the old AST must be parsed to identify deleted nodes. The parsing itself is necessary — what's missing is a size guard.

### When this triggers
- A PR that **shrinks or rewrites** a previously large file: `old_content` is large, `new_content` is small.
- Examples: refactors that split a big file, removing a vendored blob, truncating an embedded dataset.

### Why it matters
- The guard's stated intent (avoid parsing pathologically large blobs) is silently bypassed in this case.
- A miner could submit PRs that delete huge files specifically to pay parsing cost on validators while still passing the new-file check.

### Proposed fix
Apply the same size cap to `old_content`. Example:

​```python
old_too_big = (
    content_pair.old_content is not None
    and len(content_pair.old_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES
)
new_too_big = len(content_pair.new_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES
if old_too_big or new_too_big:
    # produce the existing 'skipped-large' FileScoreResult
    ...
​```

aliangm:
**Description:**

## Summary
- The `MAX_FILE_SIZE_BYTES` guard in `calculate_token_score_from_file_changes()` ([gittensor/validator/utils/tree_sitter_scoring.py:312](gittensor/validator/utils/tree_sitter_scoring.py#L312)) only checked `new_content`. A PR that shrinks or rewrites a previously large file would pass the new-side check, then `score_tree_diff` would still parse an arbitrarily large old AST to score deletions ([line 190-193](gittensor/validator/utils/tree_sitter_scoring.py#L190-L193)).
- Now skipped (`scoring_method='skipped-large'`) when either side exceeds the limit, restoring the guard's stated intent.

## Test plan
- [ ] PR adding a large file (>`MAX_FILE_SIZE_BYTES`): still skipped as before.
- [ ] PR shrinking a large file (old large, new small): now skipped instead of silently parsing the old AST.
- [ ] PR replacing a small file with a small file: scored normally.
- [ ] PR deleting a large file (`status == 'removed'`): still hits the existing `'removed'` branch and isn't affected.

Fixes #976

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -309,12 +309,9 @@ def calculate_token_score_from_file_changes(
                     is_test_file=is_test_file,
                     scoring_method='skipped-binary',
                 )
-            elif (
-                len(content_pair.new_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES
-                or (
-                    content_pair.old_content is not None
-                    and len(content_pair.old_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES
-                )
+            elif len(content_pair.new_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES or (
+                content_pair.old_content is not None
+                and len(content_pair.old_content.encode('utf-8')) > MAX_FILE_SIZE_BYTES
             ):
                 bt.logging.debug(f'  │   {file.short_name}: skipped (file too large, >{MAX_FILE_SIZE_BYTES} bytes)')
                 file_result = FileScoreResult(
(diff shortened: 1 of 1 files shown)


anderdc (approved):
Old-side size guard symmetric to new-side. Closes the shrink-large-file path where score_tree_diff parsed an unbounded old AST.

PR #982, opened 2026-05-04
fix: avoid recursion limit in tree-sitter signature collection

### LINKED ISSUE #978: `walk_node` in tree-sitter scoring can hit Python's recursion limit on deeply nested ASTs
**Description:**

`collect_node_signatures()` in [gittensor/validator/utils/tree_sitter_scoring.py:116-138](gittensor/validator/utils/tree_sitter_scoring.py#L116-L138) walks the tree-sitter AST via a recursive inner function `walk_node`. Python's default recursion limit is ~1000 frames, and tree-sitter ASTs are not depth-bounded — so any file with a sufficiently deep tree raises `RecursionError` mid-scoring and the file's contribution is lost.

### When this triggers
- Long left-associative method/attribute chains: `a.b.c.d.e.f...` (each access nests).
- Deeply nested data literals: large nested JSON-as-code, `dict`/`list`/`tuple` pyramids, deeply nested JSX.
- Long binary-operator chains (`a + b + c + ...`) in some grammars.
- Generated or minified code.

A miner submitting an otherwise legitimate PR that happens to include such a file would get a partial or zero score, with no clear signal as to why beyond a debug log.

### Why it matters
- Silent scoring failures: the recursion error is caught implicitly only if it bubbles up through `parse_code` paths; otherwise it propagates and skips the file or aborts the round.
- Adversarial angle: a miner aware of the limit could craft pathological PRs to disrupt other miners' scoring rounds (depending on how the error propagates through `get_rewards`).

aliangm:
**Description:**

## Summary
- `collect_node_signatures()` in [gittensor/validator/utils/tree_sitter_scoring.py:116-138](gittensor/validator/utils/tree_sitter_scoring.py#L116-L138) walked the AST via a recursive inner function. Deeply nested ASTs (long chained method calls, deeply nested data literals, generated/minified code) could exceed Python's default recursion limit (~1000) and raise `RecursionError` mid-scoring.
- Replaced with an iterative DFS using an explicit stack — same semantics, depth bounded only by available memory.
- Drops the unreachable inner `COMMENT_NODE_TYPES` check at [line 131](gittensor/validator/utils/tree_sitter_scoring.py#L131); the top-of-loop comment guard already covers it.

## Test plan
- [ ] Existing tree-sitter scoring tests pass unchanged.
- [ ] Spot-check a real PR's scoring output before/after — totals identical.
- [ ] Synthetic input with >1000-deep nesting (e.g. `a.b.c...` chain) no longer raises `RecursionError` and produces a non-zero breakdown.
- [ ] Comment-only files still score 0.

Fixes #978

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -113,10 +113,15 @@ def collect_node_signatures(
     """
     signatures: Counter[NodeSignature] = Counter()
 
-    def walk_node(node: Node) -> None:
-        # Skip comments entirely
+    # Iterative DFS to avoid Python's recursion limit on deeply nested ASTs
+    # (e.g. long chained method calls, deeply nested data literals).
+    stack: List[Node] = [tree.root_node]
+    while stack:
+        node = stack.pop()
+
+        # Skip comments entirely (and their subtrees)
         if is_comment_node(node):
-            return
+            continue
 
         node_type = node.type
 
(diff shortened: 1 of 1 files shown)


anderdc:
Real-world tree-sitter ASTs don't cross Python's recursion limit in legitimate code paths. Closing.

PR #984, opened 2026-05-04
perf: encode file contents once in tree-sitter scoring

### LINKED ISSUE #983: Redundant UTF-8 encoding of file contents in tree-sitter scoring
**Description:**

In `calculate_token_score_from_file_changes()` at [gittensor/validator/utils/tree_sitter_scoring.py:312-315](gittensor/validator/utils/tree_sitter_scoring.py#L312-L315), `new_content.encode('utf-8')` (and now `old_content.encode('utf-8')` after the size-guard fix) is called for the `MAX_FILE_SIZE_BYTES` length check. The same strings are then re-encoded inside `parse_code()` at [line 77](gittensor/validator/utils/tree_sitter_scoring.py#L77) before being handed to tree-sitter.

### Why it matters
- `str.encode('utf-8')` allocates a fresh `bytes` object each call. For multi-MB source files this doubles allocation and CPU on the encode path before any AST work begins.
- Validators score thousands of files per round, so the redundancy compounds.
- Purely a perf/cleanup concern — no behavioral change.

### Scope
Confirmed via codebase scan that this is the **only** site with this pattern; `gittensor/` and `neurons/` have no other duplicate-encode hot paths. Fix is contained to a single file.

aliangm:
**Description:**

## Summary
- `new_content` was encoded to UTF-8 twice — once for the `MAX_FILE_SIZE_BYTES` size check at [tree_sitter_scoring.py:312](gittensor/validator/utils/tree_sitter_scoring.py#L312) and again inside `parse_code()` at [line 77](gittensor/validator/utils/tree_sitter_scoring.py#L77). `old_content` was likewise re-encoded inside `parse_code()`.
- Encode each side once at the call site in `calculate_token_score_from_file_changes()` and pass `bytes` through. `parse_code()` and `score_tree_diff()` now accept `Union[str, bytes]` so existing `str` callers (tests) keep working unchanged.
- Reorders the skip checks so `supports_tree_sitter(ext)` short-circuits before any encoding — unsupported extensions no longer pay encode cost.

## Behavior change
A file that is both unsupported-extension **and** larger than `MAX_FILE_SIZE_BYTES` will now log as `skipped-unsupported` instead of `skipped-large`. Same final outcome (skipped, score 0), only the metric label differs.

## Test plan
- [x] Existing token-scoring integration tests pass (`pytest tests/validator/test_token_scoring_integration.py`).
- [ ] Spot-check scoring totals on a real PR before/after — should be identical.
- [ ] Confirm large `.py` file is skipped as `skipped-large` (size check still effective).
- [ ] Confirm large unsupported file is skipped as `skipped-unsupported` (new behavior).

Fixes #983

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -58,12 +58,13 @@ def get_parser(language: str) -> Optional[Parser]:
         return None
 
 
-def parse_code(content: str, language: str) -> Optional[Tree]:
+def parse_code(content: Union[str, bytes], language: str) -> Optional[Tree]:
     """
     Parse source code into a tree-sitter AST.
 
     Args:
-        content: Source code as string
+        content: Source code as ``str`` or pre-encoded UTF-8 ``bytes``. Bytes are
+            preferred on hot paths to avoid re-encoding the same string twice.
         language: Tree-sitter language name
 
     Returns:
@@ -73,8 +74,11 @@ def parse_code(content: str, language: str) -> Optional[Tree]:
     if not parser:
         return None
 
+    if isinstance(content, str):
+        content = content.encode('utf-8')
+
     try:
-        return parser.parse(content.encode('utf-8'))
+        return parser.parse(content)
     except Exception as e:
         bt.logging.debug(f'Failed to parse code: {e}')
         return None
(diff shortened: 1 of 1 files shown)


anderdc:
fix CI

aliangm:
@anderdc Fixed CI, ty

anderdc:
fix conflicts

PR #988, opened 2026-05-05
fix: remove dead branches in _encode_args and compact length encoder

### LINKED ISSUE #987: fix: remove dead branches in _encode_args and compact length encoder
## Summary

Three dead branches exist in `contract_client.py` that are unreachable given the current callers but contain incorrect code that would crash or silently corrupt if ever reached.

---

### Dead branch 1 — SCALE compact encode overflow (`_call_contract_method_raw`, line 272)

```python
else:
    compact_len = bytes([(data_len << 2) | 1, data_len >> 6])  # wrong + unreachable
```

`input_data` is always a 4-byte selector (`CONTRACT_SELECTORS.get(method_name)`), so `data_len` is always 4 and the `else` branch (data_len ≥ 64) is never reached. The formula is also wrong: `(data_len << 2) | 1` overflows a single byte for any `data_len ≥ 64` (e.g. `data_len=64` → byte value 257 → `ValueError`).

---

### Dead branch 2 — `u128` encoder in `_encode_args` (line 533)

```python
elif type_def == 'u128':
    encoded += struct.pack('<QQ', value & 0xFFFFFFFFFFFFFFFF, value >> 64)
```

`u128` only appears in `register_issue` in `metadata.json`. But `register_issue` is never called through `_exec_contract_raw` → `_encode_args` — it is called from the CLI (`mutations.py`) via a different execution path. Every actual caller of `_encode_args` uses only `u64`, `u32`, `AccountId`, or `array32`.

---

### Dead branch 3 — `str` / `unknown` fallback `raise` in `_encode_args` (line 554)

```python
else:
```

`str` appears only in `register_issue` (never routed here — same reason as above). `unknown` appears only in `get_issues_by_status`, which is a read call via `_call_contract_method_raw`, not `_encode_args`. No reachable method produces an unhandled type.

---

aliangm:
Fixes #987 
## Summary

Removes three dead/unreachable branches in `contract_client.py` that contained incorrect code and replaced them with assertions that make the invariants explicit.

- **Compact encode `else` branch** (`_call_contract_method_raw`): `input_data` is always a 4-byte selector so `data_len` is always 4 — the `else` (data_len ≥ 64) was never reached. The formula was also wrong: `(data_len << 2) | 1` overflows a single byte for data_len ≥ 64, producing a `ValueError` if ever triggered.
- **`u128` branch in `_encode_args`**: only `register_issue` uses `u128`, but that method is called from the CLI via a different execution path (`ContractInstance.exec`), never through `_exec_contract_raw` → `_encode_args`.
- **`else: raise ValueError` fallback in `_encode_args`**: only `str` (from `register_issue`) and `unknown` (from `get_issues_by_status`) would reach this branch — neither method is routed through `_encode_args`.

## Changes

- Replace compact encode `if/else` with `assert data_len < 64`
- Remove unreachable `u128` encoder branch
- Replace `raise ValueError` fallback with `assert False` and an explanatory message

## Testing

No behavior change for any current caller. The assertions will surface loudly if a future refactor accidentally routes a new method through these paths.

### DIFF
--- a/gittensor/validator/issue_competitions/contract_client.py
+++ b/gittensor/validator/issue_competitions/contract_client.py
@@ -268,8 +268,10 @@ class IssueCompetitionContractClient:
             data_len = len(input_data)
             if data_len < 64:
                 compact_len = bytes([data_len << 2])
+            elif data_len < 16384:
+                compact_len = struct.pack('<H', (data_len << 2) | 1)
             else:
-                compact_len = bytes([(data_len << 2) | 1, data_len >> 6])
+                raise ValueError(f'input_data too large for 2-byte compact: {data_len}')
 
             call_params = origin + dest + value + gas_limit + storage_limit + compact_len + input_data
 
@@ -530,8 +532,6 @@ class IssueCompetitionContractClient:
                 encoded += struct.pack('<I', value)
             elif type_def == 'u64':
                 encoded += struct.pack('<Q', value)
-            elif type_def == 'u128':
-                encoded += struct.pack('<QQ', value & 0xFFFFFFFFFFFFFFFF, value >> 64)
             elif type_def == 'AccountId':
                 if isinstance(value, str):
                     encoded += bytes.fromhex(self.subtensor.substrate.ss58_decode(value))


anderdc:
Branches are unreachable as the issue notes; replacing them with `assert` is no functional change (asserts strip under `-O`). Closing.

PR #769, opened 2026-04-24
fix: sync pr_state for stale-closed PRs excluded from scoring

RenzoMXD:
## Summary

Fixes phantom OPEN rows on miner dashboards caused by the 35-day creation cliff introduced in #406. Stale-closed PRs now have their `pr_state` persisted while remaining excluded from credibility scoring — #406's intent (no fresh credibility penalty when miners close old backlog PRs) is preserved.

- Add `stale_closed_pull_requests` bucket on `MinerEvaluation` for CLOSED PRs that are observed by the scanner but excluded from scoring.
- Route the `created_dt < lookback_date_filter` skip in `try_add_open_or_closed_pr` into the new bucket instead of dropping the PR entirely.
- Persist the new bucket through the existing `store_pull_requests_bulk`  / `BULK_UPSERT_PULL_REQUESTS` path — no schema change, no migration.
  Previously frozen `pr_state='OPEN'` rows now UPSERT to their real GitHub state.
- `total_{merged,open,closed}_prs` properties are intentionally unchanged — the new bucket is storage-only and does not inflate counts. This keeps `calculate_credibility` inputs and all scoring buckets identical to `test`. The dashboard mismatch (e.g. 7 Open in the table vs 3 in the header tile) resolves because the table now reads the refreshed `pr_state='CLOSED'`, not because totals changed.

## UPSERT semantics

`BULK_UPSERT_PULL_REQUESTS` zeroes scoring columns (`earned_score`, `token_score`, `collateral_score`, multipliers) on conflict. For a PR that was previously OPEN with collateral and is now CLOSED-not-merged and out-of-window, zero is the correct value — the PR no longer provides collateral. No partial-update query needed.

## Related Issues

Fixes #768

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- [x] `tests/validator/test_unscored_stale_prs.py` — 5 tests covering: stale routing to the storage-only bucket, totals NOT inflated, fresh-PR regression guard, mixed fresh/stale, multi-stale totals invariant.
- [x] `tests/validator/utils/test_storage_mirror.py` — 1 test confirming `store_evaluation` calls `store_pull_requests_bulk` with the new bucket as a separate (4th) call.
- [x] Full validator test suite passes locally (708/708).
- [x] Ruff lint + format clean.
- [x] Pyright: 0 errors.

## Checklist

- [x] Code follows project style guidelines (ruff + pyright clean).
[...]

PR #992, opened 2026-05-05
fix: prevent scoring data loss and fix fragile test in stale PR refresh 

### LINKED ISSUE #991: [Bug] Three gaps in #769 stale-closed PR fix: scoring data loss, silent drop path, and fragile test
## Description

PR #769 introduced the `stale_closed_pull_requests` bucket to refresh `pr_state` for out-of-window PRs but left two issues unaddressed.

**Gap 1 — Stale UPSERT silently zeroes previously-scored rows**
`store_pull_requests_bulk` in `gittensor/validator/utils/storage.py:74-76` uses `BULK_UPSERT_PULL_REQUESTS`, an `ON CONFLICT DO UPDATE SET` that overwrites all columns via `EXCLUDED.*`. `PullRequest.from_graphql_response()` leaves every scoring field at its dataclass default (`base_score = 0.0`, `earned_score = 0.0`, `credibility_multiplier = 1.0`, etc.), so any row that was previously scored has those computed values silently replaced with zeros on every subsequent scan.

**Gap 2 — Fragile positional assertion in `test_stale_closed_prs_are_stored_separately`**
`tests/validator/utils/test_storage_mirror.py:176` asserts stale storage via a hardcoded call index `call_args_list[3]`. Any reordering of bulk calls in `store_evaluation` silently breaks the assertion's target without failing the test.

## Steps to Reproduce

1. A miner has a PR that was previously scored (e.g. was OPEN with collateral) and later closed outside the 35-day lookback window.
2. On the next validator scan, `try_add_open_or_closed_pr` routes it to `stale_closed_pull_requests`.
3. `store_pull_requests_bulk` fires the full `BULK_UPSERT_PULL_REQUESTS` with dataclass-default scoring fields.
4. The previously-computed `earned_score`, `base_score`, `credibility_multiplier`, etc. are overwritten with `0.0` / `1.0` in the database.

## Expected Behavior

- Only `pr_state` and `updated_at` should be modified for stale PRs. All scoring columns on existing rows must be left untouched.
- The stale storage test should assert against the dedicated stale method by name, not by positional call index.

## Actual Behavior

- Previously-scored rows have their scoring columns zeroed out by the stale UPSERT on every subsequent scan.
- `test_stale_closed_prs_are_stored_separately` asserts via `call_args_list[3]` and would pass incorrectly if call order in `store_evaluation` changed.

## Environment

- OS: — Linux
- Python version: — 3.13
- Commit/Version: `86390ad` (post-#769 merge)

## Additional Context

Follow-up to #769, #768. No schema changes are required to fix either gap. Gap 1 can be resolved with a dedicated method backed by a targeted `UPDATE` touching only `pr_state` and `updated_at` instead of routing through the full-column UPSERT path.

wdeveloper16:
## Description

Two gaps left unaddressed by #769.

**Gap 1 — Stale UPSERT silently zeroes previously-scored rows**

`store_pull_requests_bulk` in `gittensor/validator/utils/storage.py:74-76` routes stale PRs through `BULK_UPSERT_PULL_REQUESTS`, which uses `ON CONFLICT DO UPDATE SET` with `EXCLUDED.*` — overwriting every column on conflict. `PullRequest.from_graphql_response()` leaves all scoring fields at dataclass defaults (`base_score = 0.0`, `earned_score = 0.0`, `credibility_multiplier = 1.0`, etc.). Any PR that was previously scored — for example, one that was OPEN with collateral before the window shifted — has those computed values silently replaced with zeros on every subsequent scan.

**Gap 2 — Fragile positional assertion in `test_stale_closed_prs_are_stored_separately`**

`tests/validator/utils/test_storage_mirror.py:176` asserts stale storage correctness via `call_args_list[3]`. If the call order inside `store_evaluation` ever changes, the assertion silently targets the wrong call and the test continues to pass.

## Changes

- Add `REFRESH_STALE_PR_STATES` to `queries.py` — a targeted `UPDATE` that touches only `pr_state` and `updated_at`, leaving all scoring columns on existing rows untouched
- Add `Repository.refresh_stale_pr_states()` backed by the new query
- Wire `store_evaluation` to call `refresh_stale_pr_states` instead of `store_pull_requests_bulk` for the stale bucket
- Replace `call_args_list[3]` with `refresh_stale_pr_states.assert_called_once()` and add a leak guard asserting stale PRs do not reach `store_pull_requests_bulk`

No schema changes required.

Closes #991

### DIFF
--- a/gittensor/validator/storage/queries.py
+++ b/gittensor/validator/storage/queries.py
@@ -100,6 +100,16 @@ DO UPDATE SET
     updated_at = NOW()
 """
 
+# Targeted state-only refresh for stale-closed PRs (avoids overwriting scored columns)
+REFRESH_STALE_PR_STATES = """
+UPDATE pull_requests
+   SET pr_state   = 'CLOSED',
+       updated_at = NOW()
+ WHERE number               = %s
+   AND repository_full_name = %s
+   AND pr_state            != 'CLOSED'
+"""
+
 # Issue Queries
 BULK_UPSERT_ISSUES = """
 INSERT INTO issues (
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -23,6 +23,7 @@ from .queries import (
     CLEANUP_STALE_MINER_EVALUATIONS_BY_HOTKEY,
     CLEANUP_STALE_MINERS,
     CLEANUP_STALE_MINERS_BY_HOTKEY,
+    REFRESH_STALE_PR_STATES,
     SET_MINER,
 )
 
@@ -215,6 +216,28 @@ class Repository(BaseRepository):
             if commit:
                 self.db.rollback()
             self.logger.error(f'Error in bulk pull request storage: {e}')
+            return 0
+
+    def refresh_stale_pr_states(self, pull_requests: List[PullRequest], commit: bool = True) -> int:
+        """Update pr_state to CLOSED for stale PRs without touching scoring columns.
+
+        Uses a targeted UPDATE so previously-computed scores (earned_score, base_score,
+        credibility_multiplier, etc.) are preserved on rows that were already scored.
+        Only rows currently stored as non-CLOSED are affected.
+        """
+        if not pull_requests:
+            return 0
+        values = [(pr.number, pr.repository_full_name) for pr in pull_requests]
+        try:
+            with self.get_cursor() as cursor:
+                cursor.executemany(REFRESH_STALE_PR_STATES, values)
+                if commit:
+                    self.db.commit()
+                return len(values)
+        except Exception as e:
+            if commit:
+                self.db.rollback()
+            self.logger.error(f'Error refreshing stale PR states: {e}')
             return 0
 
     def store_issues_bulk(self, issues: List[Issue], commit: bool = True) -> int:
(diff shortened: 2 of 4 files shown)


wdeveloper16:
Hi, @anderdc 
Can you review the PR, please?

wdeveloper16:
- `store_evaluation` routes `stale_closed_pull_requests` through `store_pull_requests_bulk`, which runs `BULK_UPSERT_PULL_REQUESTS` — an `ON CONFLICT DO UPDATE SET` that overwrites every column including `earned_score`, `base_score`, `credibility_multiplier`, `pioneer_rank`, and `pioneer_dividend`. 
- `PullRequest.from_graphql_response()` never computes scoring fields for stale PRs, so those columns arrive as dataclass defaults (`0.0` / `1.0`). 
- The overwrite is deterministic: any PR scored in epoch N becomes stale in epoch N+35 as the lookback window shifts, and the next scan silently zeroes its stored scores — destroying audit-accurate scoring history every cycle for any miner with aged-out PRs.

This PR replaces the full UPSERT for the stale bucket with a dedicated `REFRESH_STALE_PR_STATES` query that touches only `pr_state` and `updated_at`, leaving all scoring columns untouched. No schema changes, no new tables, no behavioural change to the scoring pipeline.

anderdc (approved):

PR #993, opened 2026-05-05
fix: remove dead branches in contract_client.py

### LINKED ISSUE #987: fix: remove dead branches in _encode_args and compact length encoder

597226617:
Fixes #987

## Summary
Removed two dead branches in `contract_client.py` that are unreachable given current callers but contained incorrect code.

### 1. SCALE compact encode overflow (line 272)
**Before:** Unreachable `else` branch for `data_len >= 64` with incorrect formula that would `ValueError` on overflow.
**After:** Removed the dead `else` branch. `input_data` is always a 4-byte contract selector, so `data_len` is always < 64.

### 2. u128 encoder in `_encode_args` (line 533)
**Before:** Dead `elif type_def == 'u128'` branch — only `register_issue` uses u128, but it bypasses `_encode_args` via CLI mutations path.
**After:** Retained the encoder with a comment explaining it is dead code. Kept (rather than removed) for forward-compatibility if future methods add u128 args.

### 3. `parse_code` accepts `Union[str, bytes]` (bonus fix for #983)
Modified `parse_code` to accept pre-encoded bytes, enabling callers to avoid redundant UTF-8 encoding without changing the public API.

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -370,6 +370,10 @@ class MinerEvaluation:
     total_leaf_score: float = 0.0
     failed_reason: Optional[str] = None
     github_pr_fetch_failed: bool = False
+    # Mirror-source-specific fetch flag set by mirror.combine.combine alongside
+    # the OR into github_pr_fetch_failed. Lets the validator tell a complete
+    # mirror outage apart from a legacy partial-pagination failure.
+    mirror_pr_fetch_failed: bool = False
     evaluation_timestamp: Optional[datetime] = None
     merged_pull_requests: List[PullRequest] = field(default_factory=list)
     open_pull_requests: List[PullRequest] = field(default_factory=list)
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -305,12 +305,8 @@ class PullRequest:
         last_edited_at = parse_github_timestamp_to_cst(raw_edited_at) if isinstance(raw_edited_at, str) else None
         merged_at = parse_github_timestamp_to_cst(pr_data['mergedAt']) if is_merged else None
 
-        changes_requested_count = 0
-        if is_merged:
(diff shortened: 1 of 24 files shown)


anderdc:

PR #998, opened 2026-05-05
fix(cli): surface excluded validator list in error path

### LINKED ISSUE #989: [Bug] gitt miner post/check error path discards excluded validator list when --min-stake / --min-vtrust filters eliminate all candidates
# Description:
## Bug
`gitt miner post` and `gitt miner check` both call `_require_validator_axons(...)`, which already returns an `excluded` list explaining *why* each high-vtrust validator was filtered (e.g. `stake 5,000 α below 15,000 α threshold`, `not serving an axon`). The **success path** uses this — it renders a `Skipped Validators` table in TTY mode and includes a `skipped` field in the JSON envelope. The **error path** discards `excluded` entirely and emits the generic message `"No reachable validator axons found on the network."`, even when the real cause is the operator's `--min-stake` / `--min-vtrust` filter eliminating every candidate.

## Source

[`gittensor/cli/miner_commands/helpers.py:127-141`](https://github.com/entrius/gittensor/blob/test/gittensor/cli/miner_commands/helpers.py#L127-L141):

```python
def _require_validator_axons(metagraph, json_mode, *, min_vtrust=..., min_stake=...):
    validator_axons, validator_uids, excluded = _get_validator_axons(
        metagraph, min_vtrust=min_vtrust, min_stake=min_stake
    )
    if not validator_axons:
        _error('No reachable validator axons found on the network.', json_mode)  # ← excluded discarded
        sys.exit(1)
    return validator_axons, validator_uids, excluded
```

## Reproduction

Network with three high-vtrust validators (vtrust 0.99 / 0.85 / 0.72), all carrying stakes below the default `--min-stake=15000`:

```
$ gitt miner post --pat ghp_xxx --wallet alice --hotkey default
Error: No reachable validator axons found on the network.
```

Same in JSON mode — three known-but-skipped validators, each with explanation, are not surfaced:

```json
{"success": false, "error": "No reachable validator axons found on the network."}
```

The success path of the same command always emits a `skipped` array and prints the `Skipped Validators` table; the error path silently drops it.

## Expected behavior

When all candidates are excluded by `--min-stake` / `--min-vtrust`:

1. **TTY**: print the existing `Skipped Validators` table (already implemented in `_render_skipped_validators`) before the error, and use an actionable message like `No validators passed --min-vtrust=0.25 / --min-stake=15,000 α; all 3 candidate(s) excluded.`
2. **JSON**: include the `skipped` array on the error envelope so JSON consumers see why the broadcast was empty.

When the underlying metagraph is genuinely empty (no excluded UIDs either), keep the current generic message.

## Affected commands

Both `gitt miner post` and `gitt miner check` go through `_require_validator_axons`.

597226617:
Fixes #989

## Summary
`_require_validator_axons` already computes an excluded list with per-UID reasons (stake too low, no axon, etc.) but the error path discarded it and showed a generic message.

## Fix
When all candidates are excluded, the error message now surfaces each UID and its exclusion reason so operators can fix their `--min-stake`/`--min-vtrust` filters.

### DIFF
--- a/gittensor/cli/miner_commands/helpers.py
+++ b/gittensor/cli/miner_commands/helpers.py
@@ -136,7 +136,18 @@ def _require_validator_axons(
         metagraph, min_vtrust=min_vtrust, min_stake=min_stake
     )
     if not validator_axons:
-        _error('No reachable validator axons found on the network.', json_mode)
+        # Surface why validators were excluded instead of a generic message
+        if excluded:
+            reasons = '; '.join(
+                f'UID {e.get("uid","?")} — {", ".join(e.get("reasons",[]))}'
+                for e in excluded
+            )
+            _error(
+                f'No reachable validator axons found. All candidates were excluded: {reasons}',
+                json_mode,
+            )
+        else:
+            _error('No reachable validator axons found on the network.', json_mode)
         sys.exit(1)
     return validator_axons, validator_uids, excluded
 


anderdc:
Duplicate of #990 which is better scoped/implemented. Closing.

PR #1009, opened 2026-05-05
Fix mirror open issue count to ignore transferred issues

### LINKED ISSUE #1007: [Bug] Mirror issue-discovery open_issue_count inflates spam multiplier by including transferred-OPEN issues that are otherwise classified as 'ignore'
## Bug

[`gittensor/validator/issue_discovery/mirror_scan.py:174`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/issue_discovery/mirror_scan.py#L174) computes the spam-multiplier signal:

```python
open_issue_count = sum(1 for i in filtered if i.state == 'OPEN')
```

This count includes issues whose `is_transferred == True`. Everywhere else in the same module, transferred issues are treated as having **no scorable meaning at all**:

- Module docstring lines 12-19 list "not `issue.is_transferred`" among the anti-gaming gates.
- `_classify_issue` ([line 468-470](https://github.com/entrius/gittensor/blob/test/gittensor/validator/issue_discovery/mirror_scan.py#L468-L470)) returns `'ignore'` for transferred issues regardless of state.

The spam-multiplier path is the only place in the module that treats transferred-OPEN issues as if they were real OPEN issues this miner is responsible for.

## Source

`mirror_scan.py:169-175`:

```python
# Count this miner's currently-open issues across mirror-enabled repos
# (within the lookback window). Used as the spam-multiplier signal and
# also written to evaluation.total_open_issues so the DB row reflects
# mirror-scoped state ...
open_issue_count = sum(1 for i in filtered if i.state == 'OPEN')   # ← does not filter is_transferred
pending.append((evaluation, filtered, open_issue_count))
```

`mirror_scan.py:458-470` (the canonical "no scorable meaning" check):

```python
def _classify_issue(issue: MirrorIssue) -> str:
    """Return 'solved', 'not-solved-closed', or 'ignore' per anti-gaming gates.

    'ignore' = issue is open / transferred / has no scorable meaning at all.
    ...
    """
    if issue.is_transferred:
        bt.logging.debug(f'  issue #{issue.issue_number} ({issue.repo_full_name}): ignore (transferred)')
        return 'ignore'
```

## Impact

1. **Inflated spam multiplier** — `calculate_open_issue_spam_multiplier(open_issue_count, …)` reads the inflated count and reduces the per-solved-issue discovery score. A miner with 5 legitimate OPEN issues plus 3 transferred-OPEN issues gets penalized as though they had 8 currently-open issues.

2. **Corrupted DB audit row** — [`mirror_scan.py:376`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/issue_discovery/mirror_scan.py#L376) writes `evaluation.total_open_issues = open_issue_count`. The `miner_evaluations` row carries the inflated number, so analytics queries (`SELECT total_open_issues …`) show miners as having more open issues than they actually own.

3. **Asymmetric anti-gaming** — the docstring promises that transferred issues "have no scorable meaning at all," but they DO have negative scoring meaning via the spam multiplier path. An attacker controlling repo administration could file an issue, transfer it to a target miner's tracking surface, and inflate that miner's spam count even though the issue itself can never score.

## Reproduction

Construct a `MirrorIssue` list containing two entries:
- `state='OPEN', is_transferred=False`
- `state='OPEN', is_transferred=True`

Drive them through `_score_miner_mirror_issues`. Observe `open_issue_count == 2`. Observe the resulting spam multiplier as if both issues were real. Then mark `is_transferred=False` on the second issue and run again — multiplier changes, demonstrating the transferred issue affected the score.

## Suggested fix

One-line change at line 174:

```python
open_issue_count = sum(1 for i in filtered if i.state == 'OPEN' and not i.is_transferred)
```

Optionally, add a regression test that asserts transferred-OPEN issues do not contribute to `open_issue_count` and `total_open_issues`.

## Related

- [#929](https://github.com/entrius/gittensor/issues/929) / [#930](https://github.com/entrius/gittensor/pull/930) address a **different** gap on the same line — old still-open issues outside the lookback window being undercounted. Fixing that does not address transferred-issue inclusion; the two filters are independent.
- The merged [#796](https://github.com/entrius/gittensor/pull/796) introduced this scoring path with `is_transferred` already being a recognized "ignore" signal in `_classify_issue`; the omission in the spam-count branch appears to be an oversight rather than a deliberate semantic choice.

zeus1959:
## Summary
Closes  #1007

Exclude transferred OPEN mirror issues from `open_issue_count`. Transferred issues are already classified as `ignore`, so they should not affect the spam multiplier or `total_open_issues`.
- Exclude transferred OPEN issues from mirror issue-discovery `open_issue_count`
- Prevent transferred issues from inflating the open-issue spam multiplier
- Keep `evaluation.total_open_issues` aligned with mirror anti-gaming semantics
- Add regression coverage for transferred OPEN issues not affecting spam scoring or audit counts

## Changes
- Update `mirror_scan.py` to count only non-transferred OPEN issues.
- Add regression coverage for transferred OPEN issues.
- Verify transferred OPEN issues do not inflate `evaluation.total_open_issues` or trip spam scoring.

## Test plan
- `python3 -m pytest tests/validator/issue_discovery/test_mirror_scan.py`

### DIFF
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -171,7 +171,7 @@ async def run_mirror_issue_discovery(
         # also written to evaluation.total_open_issues so the DB row reflects
         # mirror-scoped state (the legacy GraphQL global open-issue count was
         # never the right signal for the gate).
-        open_issue_count = sum(1 for i in filtered if i.state == 'OPEN')
+        open_issue_count = sum(1 for i in filtered if i.state == 'OPEN' and not i.is_transferred)
         pending.append((evaluation, filtered, open_issue_count))
 
     canonical_pr_owners = _build_canonical_pr_owners(pending)
--- a/tests/validator/issue_discovery/test_mirror_scan.py
+++ b/tests/validator/issue_discovery/test_mirror_scan.py
@@ -702,6 +702,39 @@ class TestOpenIssueSpamSourceIsMirror:
         assert eval_.issue_discovery_score > 0
         assert eval_.total_open_issues == 2
 
+    def test_transferred_open_issues_do_not_count_for_spam_or_audit(self):
+        """Transferred issues are ignored by scoring and by the open-issue count."""
+        open_issues = [
+            _issue_dict(
+                issue_number=200 + i,
+                state='OPEN',
+                state_reason=None,
+                solved_by_pr=None,
+                is_transferred=(i >= 5),
+            )
+            for i in range(8)
+        ]
+        solved_issues = [_issue_dict(issue_number=300 + i, author_github_id=f'discoverer{i}') for i in range(8)]
+        client = Mock()
+        client.get_miner_issues.return_value = _response(open_issues + solved_issues)
+
+        eval_ = _eval()
+        eval_.mirror_merged_prs = [_scored_mirror_pr('entrius/gittensor-ui', 100, token_score=100.0)]
+
+        _run(
+            run_mirror_issue_discovery(
+                {1: eval_},
+                _mirror_repos('entrius/gittensor-ui'),
+                _EMPTY_LANGS,
+                _EMPTY_TOKEN_CONFIG,
+                client=client,
+            )
+        )
+
+        # Only the 5 non-transferred open issues count, so spam stays below/at threshold.
+        assert eval_.issue_discovery_score > 0
+        assert eval_.total_open_issues == 5
+
 
 class TestCrossMinerOneIssuePerPr:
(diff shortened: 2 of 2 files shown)


anderdc:
`is_transferred` is never written to `true` anywhere in das-github-mirror — `webhook/handlers/issue.handler.ts` doesn't set it, no backfill or ingest path does, default is FALSE in both the entity and SQL schema. The existing filter in `_classify_issue` is already dead code; this PR adds a second dead-code guard against a value that never appears. Closing.

PR #1025, opened 2026-05-05
fix: quantize PR lookback cutoff to UTC day so validators agree on boundary

### LINKED ISSUE #1003: [Bug] PR/issue lookback cutoff uses per-validator wall clock, causing boundary PRs to diverge across validators
## Description

Three PR/issue fetch sites compute the lookback cutoff as `datetime.now(timezone.utc) - timedelta(days=PR_LOOKBACK_DAYS)` at the moment of the call. Because each validator reads its own wall clock at its own forward-pass start, validators disagree on the cutoff by up to several seconds. A PR whose `merged_at` (or `created_at`) falls inside that window is included by some validators and excluded by others — a binary include/exclude divergence on the boundary, not a sub-quantization perturbation.

Affected call sites:

- [gittensor/utils/github_api_tools.py:875](gittensor/utils/github_api_tools.py#L875) — `lookback_date_filter` used at lines 763, 767, 796 for legacy PR fetch
- [gittensor/validator/oss_contributions/mirror/load.py:57](gittensor/validator/oss_contributions/mirror/load.py#L57) — `lookback_date` passed as `since=` to mirror PR fetch and used at line 113
- [gittensor/validator/issue_discovery/mirror_scan.py:131](gittensor/validator/issue_discovery/mirror_scan.py#L131) — `lookback_date` passed as `since=` to mirror issue fetch

All three resolve `PR_LOOKBACK_DAYS = 35` ([gittensor/constants.py:77](gittensor/constants.py#L77)).

## Steps to Reproduce

1. Pick a PR whose `merged_at` is at `T0`.
2. Validator V1 starts its forward pass at `T0 + 35d - 2s` and computes its lookback cutoff as `T0 - 2s`. The PR's `merged_at = T0` is `>= cutoff` → included.
3. Validator V2 starts its forward pass at `T0 + 35d + 2s` and computes its lookback cutoff as `T0 + 2s`. The PR's `merged_at = T0` is `< cutoff` → excluded.

V1 and V2 disagree on whether the PR contributes to the miner's score that round, and the disagreement is `pr.earned_score`-sized, not a sigmoid micro-shift. The same divergence applies to `created_at` for open/closed PR filtering at [github_api_tools.py:763-767](gittensor/utils/github_api_tools.py#L763-L767) and [mirror/load.py:113](gittensor/validator/oss_contributions/mirror/load.py#L113).

## Expected Behavior

The 35-day lookback cutoff should resolve to the same value across all validators scoring the same chain state, so a PR is either uniformly inside or uniformly outside the window. The cutoff should not depend on each validator's wall-clock read.

## Actual Behavior

Each validator anchors the lookback on its own `datetime.now()`. Boundary PRs (merged within seconds of `now − 35d`) are included by some validators and excluded by others, producing a step-function score difference for affected miners.

## Environment

- OS: any
- Python version: any
- Commit/Version: HEAD of `test` branch (`86390ad` at filing); applies to all three call sites unchanged on `main`

## Additional Context

**Why this is worth filing as a discrete issue, not a sweeping refactor:** an earlier draft of this report tried to extend the same critique to `calculate_time_decay` ([datetime_utils.py:50-60](gittensor/validator/utils/datetime_utils.py#L50-L60)) and frame it as "vtrust drift." That framing does not survive the magnitude check — at the sigmoid midpoint the derivative is ≈0.42%/hour, NTP-typical inter-validator skew is in the 10–100 ms range, and the resulting per-PR shift (~10⁻⁶ to 10⁻⁵) sits at or below uint16 weight quantization (1/65535 ≈ 1.5×10⁻⁵). The lookback boundary is the one location where wall-clock skew produces a discrete, observable score change rather than a sub-quantization perturbation.

**Why threading a single `evaluation_timestamp` through the forward pass would not fix this:** that approach makes one validator's forward pass internally consistent (V1 uses one `T1` everywhere, V2 uses one `T2` everywhere), which it already approximately is — a forward pass takes seconds, the cutoff has 1-day-ish resolution sensitivity. The cross-validator divergence between `T1` and `T2` is unchanged. The fix has to anchor on something validators share, not on each validator's own snapshot.

**Suggested fix shape:** quantize the cutoff to a coarser shared unit so all validators in a window resolve to the same boundary. Two candidates, in order of preference:

1. Round to the nearest UTC day: `cutoff = (datetime.now(timezone.utc) - timedelta(days=PR_LOOKBACK_DAYS)).replace(hour=0, minute=0, second=0, microsecond=0)`. Two validators within the same UTC day produce identical cutoffs. Trivial change at three call sites; no signature ripple.
2. Anchor on a Bittensor block height / epoch boundary if the validator already has a handle on that in `forward()`. Stronger guarantee, larger change.

Option 1 is probably enough — it eliminates the boundary divergence for ~24-hour windows and is small enough to land without an architecture discussion. Option 2 can follow if there's appetite.

**Not in scope for this issue:** the wall-clock read in `calculate_time_decay` and the four non-scoring `datetime.now()` reads (`pat_storage.py:46`, `classes.py:676`, `validator/utils/storage.py:49`). Those either don't affect scores or affect them below quantization grain.

RenzoMXD:
## Summary

The 35-day PR/issue lookback cutoff is computed at three sites as `datetime.now(timezone.utc) - timedelta(days=PR_LOOKBACK_DAYS)`. Each validator reads its own wall clock, so two validators starting their forward pass a few seconds apart end up with different cutoffs and disagree on whether a PR near the 35-day boundary is in or out of the window. The disagreement is a full include/exclude flip — the PR's entire `earned_score` shifts between validators, not a small numeric drift.

## Fix

Add a small `lookback_cutoff(lookback_days)` helper in `gittensor/validator/utils/datetime_utils.py` that returns `now - lookback_days` floored to `00:00:00 UTC`. All three call sites switch to it:

- `gittensor/utils/github_api_tools.py:875`
- `gittensor/validator/oss_contributions/mirror/load.py:57`
- `gittensor/validator/issue_discovery/mirror_scan.py:131`

Validators reading their clock at any point during the same UTC day now compute the same cutoff. The residual disagreement is one full day at UTC midnight rollovers, which is the tradeoff Option 1 in #1003 suggests.

Function signatures don't change — `should_skip_merged_pr`, `try_add_open_or_closed_pr`, and `_maybe_add_pr` still take a `datetime`; only the value passed in is now quantized.

## Out of scope

`calculate_time_decay` and the four non-scoring `datetime.now()` reads (`pat_storage.py`, `classes.py`, `validator/utils/storage.py`) are not touched. Per the analysis in #1003, their per-validator skew is below the uint16 weight quantization grain, so they don't produce the discrete divergence this PR fixes.

## Testing

New: `tests/validator/utils/test_datetime_utils.py` (4 cases)

- `test_returns_midnight_utc` — cutoff floors to 00:00:00 UTC.
- `test_two_validators_within_same_utc_day_get_identical_cutoff` — regression test for the bug: two `datetime.now()` reads 5 seconds apart produce the same cutoff.
- `test_two_validators_straddling_midnight_get_one_day_offset` — pins the residual divergence at exactly 1 day.
- `test_subtracts_correct_number_of_days` — arithmetic sanity for 0, 1, 35, 365.

Full suite: 766 passing. Ruff and Pyright clean.

## Related

Closes #1003

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -5,7 +5,7 @@ import os
 import time
 from contextlib import contextmanager
 from dataclasses import dataclass
-from datetime import datetime, timedelta, timezone
+from datetime import datetime
 from math import ceil
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
 
@@ -32,7 +32,7 @@ from gittensor.constants import (
     REVIEW_PENALTY_RATE,
 )
 from gittensor.utils.models import PRInfo
-from gittensor.validator.utils.datetime_utils import parse_github_iso_to_utc
+from gittensor.validator.utils.datetime_utils import lookback_cutoff, parse_github_iso_to_utc
 from gittensor.validator.utils.load_weights import RepositoryConfig
 
 # Beyond this many CHANGES_REQUESTED reviews the quality multiplier is already 0
@@ -872,7 +872,7 @@ def load_miners_prs(
         bt.logging.warning(f'UID {miner_eval.uid} has no github_pat, skipping PR fetch')
         return
 
-    lookback_date_filter = datetime.now(timezone.utc) - timedelta(days=PR_LOOKBACK_DAYS)
+    lookback_date_filter = lookback_cutoff(PR_LOOKBACK_DAYS)
     global_user_id = base64.b64encode(f'04:User{miner_eval.github_id}'.encode()).decode()
 
     cursor = None
--- a/gittensor/validator/issue_discovery/mirror_scan.py
+++ b/gittensor/validator/issue_discovery/mirror_scan.py
@@ -33,7 +33,7 @@ the cache so sibling discoveries benefit.
 """
 
 from dataclasses import dataclass
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timezone
 from typing import Dict, List, Optional, Set, Tuple
 
 import bittensor as bt
@@ -54,7 +54,7 @@ from gittensor.validator.oss_contributions.mirror.adapters import mirror_files_t
 from gittensor.validator.oss_contributions.mirror.scoring import (
     calculate_base_score_for_pr_files,
 )
-from gittensor.validator.utils.datetime_utils import calculate_time_decay
+from gittensor.validator.utils.datetime_utils import calculate_time_decay, lookback_cutoff
 from gittensor.validator.utils.load_weights import (
     LanguageConfig,
     RepositoryConfig,
@@ -128,7 +128,7 @@ async def run_mirror_issue_discovery(
         return
 
     client = client or MirrorClient()
-    lookback_date = datetime.now(timezone.utc) - timedelta(days=PR_LOOKBACK_DAYS)
+    lookback_date = lookback_cutoff(PR_LOOKBACK_DAYS)
     enabled_names: Set[str] = set(mirror_repos.keys())
 
     solving_pr_cache: Dict[Tuple[str, int], CachedSolvingPR] = _build_solving_pr_cache(miner_evaluations)
--- a/gittensor/validator/oss_contributions/mirror/load.py
+++ b/gittensor/validator/oss_contributions/mirror/load.py
@@ -15,7 +15,7 @@ Filtering applied at load time (legacy parity):
 """
 
 import os
-from datetime import datetime, timedelta, timezone
+from datetime import datetime
 from typing import Dict, Optional
 
 import bittensor as bt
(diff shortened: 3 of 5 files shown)


anderdc:
The boundary disagreement is rare (~5s window in a 35d range, ~1.65e-6 per PR per round) and self-healing — once `now` advances by a round, the contested PR is no longer at the boundary, and `update_scores` EMAs out single-round epsilon shifts before they affect vtrust. The proposed fix also trades a continuous narrow window for a once-per-day 24h-wide window at UTC midnight, which is round-correlated across validators straddling midnight — arguably worse for consensus than the status quo. Closing.

PR #1040, opened 2026-05-06
fix: evict duplicate-penalty cache snapshots

### LINKED ISSUE #1039: [Bug] MinerEvaluationCache.store runs before duplicate-GitHub penalty, persisting pre-penalty scores; future API failure restores them as-if-never-penalized
## Description:
In `neurons/validator.py:store_or_use_cached_evaluation`, the cache **store** path runs in the same loop as the cache **restore** path — and that combined call happens **before** `detect_and_penalize_miners_sharing_github` in `gittensor/validator/oss_contributions/reward.py:get_rewards`. As a result, a miner whose eval is about to be zeroed for sharing a GitHub account is **first cached with the pre-penalty data**, then penalized.

[#600](https://github.com/entrius/gittensor/pull/600) made sure the **current round's** zeroed score reaches the DB by removing penalized UIDs from `cached_uids` so the storage-skip-set excludes them. But the cache content itself was not touched — the cached entry still holds pre-penalty PR buckets with full `earned_score`, `pioneer_dividend`, etc.

When a future round hits a transient GitHub fetch failure for that UID **and the duplicate partner has since deregistered or rotated GitHub**, the cache restore returns the round-T pre-penalty snapshot, the post-restore `detect_and_penalize` no longer trips (no duplicate now), and the validator silently uses the resurrected pre-penalty scores — undoing the lingering effect of the round-T penalty under conditions the operator can't see.

## Source

[`gittensor/validator/oss_contributions/reward.py:148-160`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/oss_contributions/reward.py#L148-L160):

```python
# If evaluation of miner was successful, store to cache, if api failure, fallback to previous successful evaluation if any
cached_uids = self.store_or_use_cached_evaluation(miner_evaluations)   # ← cache.store runs HERE

# Adjust scores for duplicate accounts; returns penalized UIDs so they are
# removed from cached_uids and their zeroed scores are written to DB.
penalized_uids = detect_and_penalize_miners_sharing_github(miner_evaluations)   # ← penalty runs AFTER
cached_uids -= penalized_uids
```

[`neurons/validator.py:160-165`](https://github.com/entrius/gittensor/blob/test/neurons/validator.py#L160-L165) — the inner store call:

```python
if not miner_eval.github_pr_fetch_failed:
    if miner_eval.total_prs > 0:
        self.evaluation_cache.store(miner_eval)   # ← `failed_reason` is None here even for soon-to-be-penalized duplicates
    continue
```

[`gittensor/classes.py:store`](https://github.com/entrius/gittensor/blob/test/gittensor/classes.py) — gate is on `failed_reason`:

```python
def store(self, evaluation: 'MinerEvaluation') -> None:
    if evaluation.failed_reason is not None:
        return
    ...
    cached_eval = self._build_cache_entry(evaluation)   # shallow copy: failed_reason is None at this snapshot
    self._cache[evaluation.uid] = CachedEvaluation(...)
```

`_build_cache_entry` does `copy.copy(evaluation)`, which captures `failed_reason=None` at copy time. The subsequent penalty mutates the **live** eval but the **cached** copy retains `failed_reason=None` and the full pre-penalty PR buckets.

## Concrete scenario

1. **Round T.** UID-100 (legitimate) and UID-250 (attacker) both broadcast PATs for the same GitHub account. Both produce normal-looking PR buckets (same GitHub → same scoring data).
2. `store_or_use_cached_evaluation(...)` runs — both UIDs are cached with pre-penalty PR buckets and `failed_reason=None`.
3. `detect_and_penalize_miners_sharing_github(...)` zeroes both. DB rows for round T correctly show zero.
4. **Round T+1.** UID-250 deregisters or rotates to a different GitHub. UID-100 is now alone with that GitHub. UID-100's GitHub fetch **fails** (transient `/user` outage, mirror outage, network blip).
5. Cache fallback: `cached_eval = self.evaluation_cache.get(100, ...)` returns the round-T pre-penalty snapshot — `failed_reason=None`, full `merged_pull_requests`, `mirror_merged_prs`, full `earned_score`, `pioneer_dividend`.
6. `miner_evaluations[100] = cached_eval`. UID-100 is now in `cached_uids`.
7. `detect_and_penalize_miners_sharing_github(...)` runs on the restored eval. UID-100's `failed_reason` is `None` (cached), `github_id` is the shared GH from round T. **No other UID has this GH now** (UID-250 is gone), so no duplicate detected, no penalty.
8. `finalize_miner_scores` runs on UID-100's restored eval — full pre-penalty scores survive. Storage path persists them (UID-100 is in `cached_uids` so DB write is skipped, but the eval's normalized reward in `get_rewards`'s return value is high).

The lingering effect of the round-T penalty is silently erased the first time UID-100 hits an API failure with the partner gone. Operators have no signal — the cache doesn't expose "this was penalized when stored," so the resurrection passes every defensive gate.

## Why this matters

- **Penalty escape vector.** A miner who shares a GitHub once and gets penalized can wait for any of `(transient GitHub /user outage | mirror outage | partial-page failure)` to combine with the partner being gone, and the cached pre-penalty scores quietly resurface.
- **Asymmetry with the merged [#600](https://github.com/entrius/gittensor/pull/600) intent.** That fix made the **current round** correct (zero score reaches DB). The **future round** is not addressed; the bad data is still in cache.
- **Hard to detect.** The resurrection looks identical to a normal cache fallback in logs.

## Suggested fix

Either of:

1. **Re-store after penalty.** After `detect_and_penalize_miners_sharing_github` runs, walk `penalized_uids` and call `self.evaluation_cache._cache.pop(uid, None)` so future cache lookups for those UIDs don't return the pre-penalty snapshot. Re-storing the post-penalty (zeroed) eval is a no-op because `evaluation_cache.store` skips entries with `failed_reason is not None`.

2. **Reorder.** Move the cache **store** (but not the cache **restore**) to after the penalty step. The restore must run first because penalty needs the restored eval to detect persistent duplicates; the store can move after.

Option 1 is the smaller change — adds one helper (`evaluation_cache.evict(uid)`) and one loop in `get_rewards`.

## Distinct from prior filings

- [#596](https://github.com/entrius/gittensor/issues/596) (closed) and [#600](https://github.com/entrius/gittensor/pull/600) (merged) addressed the **current-round DB-write skip**. They did not touch the cache content. This issue is about **future-round cache resurrection** of pre-penalty data.
- [#869](https://github.com/entrius/gittensor/issues/869) / [#1006](https://github.com/entrius/gittensor/issues/1006) cover `_zero_for_duplicate_penalty` field-list completeness — a separate concern (penalty content vs cache timing).

statxc:
## Summary

Fixes #1039.

Duplicate-GitHub penalties currently run after `store_or_use_cached_evaluation()`. That means a miner can be cached with a successful, pre-penalty evaluation before `detect_and_penalize_miners_sharing_github()` zeroes the live round state.

This PR evicts cache entries for duplicate-penalized UIDs so future GitHub fetch-failure fallback cannot restore the old pre-penalty snapshot.

## Problem

`get_rewards()` currently performs cache store/restore before duplicate-account detection:

```python
cached_uids = self.store_or_use_cached_evaluation(miner_evaluations)
penalized_uids = detect_and_penalize_miners_sharing_github(miner_evaluations)
cached_uids -= penalized_uids
```

PR #600 fixed the current-round DB persistence path by removing penalized UIDs from `cached_uids`, so their zeroed evaluations are written to DB.

However, the in-memory cache entry itself was still left untouched. If the cache store path ran before the penalty, the cache could retain a pre-penalty evaluation with full PR buckets and `failed_reason=None`.

In a later round, if GitHub fetch fails for that UID and the duplicate partner is no longer present, cache fallback can restore that stale pre-penalty snapshot. Duplicate detection then no longer trips, so the old scores can re-enter the reward path.

## Fix

This PR adds explicit cache eviction for penalized UIDs:

- `MinerEvaluationCache.evict(uid)` removes one cached evaluation.
- `MinerEvaluationCache.evict_many(uids)` removes a set of cached evaluations.
- `get_rewards()` calls `self.evaluation_cache.evict_many(penalized_uids)` immediately after duplicate detection.

This keeps the existing ordering intact:

1. cache restore still happens before duplicate detection, so restored evaluations can still be inspected for persistent duplicates
2. duplicate penalty still runs before final scoring
3. penalized UIDs are still removed from `cached_uids` so current-round zeroed evaluations reach DB
4. pre-penalty cache snapshots are now removed so future fallback cannot resurrect them

## Changes

- Add `MinerEvaluationCache.evict()`.
- Add `MinerEvaluationCache.evict_many()`.
- Evict duplicate-penalized UIDs from the evaluation cache in `get_rewards()`.
  - duplicate-penalized UIDs being removed from `cached_uids`
  - pre-penalty cache entries being evicted
  - a later fetch-failure round not restoring the old pre-penalty cached evaluation

## Testing

```bash
UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/validator/oss_contributions tests/validator/test_validator_cache_fallback.py tests/validator/test_duplicate_penalty_propagation.py -q
```

Result:

```text
134 passed
```

```bash
UV_CACHE_DIR=/tmp/uv-cache uv run ruff check gittensor/classes.py gittensor/validator/oss_contributions/reward.py tests/validator/test_validator_cache_fallback.py
UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check gittensor/classes.py gittensor/validator/oss_contributions/reward.py tests/validator/test_validator_cache_fallback.py
UV_CACHE_DIR=/tmp/uv-cache uv run pre-commit run --files gittensor/classes.py gittensor/validator/oss_contributions/reward.py tests/validator/test_validator_cache_fallback.py
git diff --check
```

All passed.

Also verified touched files with pyright using a temp config pointing at the local venv:

```bash
UV_CACHE_DIR=/tmp/uv-cache uv run pyright -p /tmp/pyright-gittensor-cache-fix.json
```

Result:

```text
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -721,6 +721,12 @@ class MinerEvaluationCache:
 
         return self._isolate_for_downstream(cached.evaluation)
 
+    def evict_many(self, uids: Set[int]) -> None:
+        """Remove cached evaluations for all provided UIDs."""
+        for uid in uids:
+            if self._cache.pop(uid, None) is not None:
+                bt.logging.debug(f'Evicted cached evaluation for UID {uid}')
+
     @staticmethod
     def _build_cache_entry(evaluation: 'MinerEvaluation') -> 'MinerEvaluation':
         # Cached evaluations feed only the GitHub-fetch-failure fallback path
--- a/gittensor/validator/oss_contributions/reward.py
+++ b/gittensor/validator/oss_contributions/reward.py
@@ -171,6 +171,9 @@ async def get_rewards(
     # removed from cached_uids and their zeroed scores are written to DB.
     penalized_uids = detect_and_penalize_miners_sharing_github(miner_evaluations)
     cached_uids -= penalized_uids
+    # The cache store path ran before the penalty. Drop those snapshots so a
+    # future fetch failure cannot restore pre-penalty PR scores.
+    self.evaluation_cache.evict_many(penalized_uids)
 
     # Finalize scores: apply eligibility gate, credibility, pioneer dividends, collateral
     finalize_miner_scores(miner_evaluations, master_repositories)
--- a/tests/validator/test_validator_cache_fallback.py
+++ b/tests/validator/test_validator_cache_fallback.py
@@ -1,29 +1,36 @@
 # Entrius 2025
 
+import asyncio
 from datetime import datetime, timezone
+from types import SimpleNamespace
 from typing import cast
-from unittest.mock import Mock, patch
+from unittest.mock import AsyncMock, Mock, patch
 
 from gittensor.classes import FileChange, Issue, MinerEvaluation, MinerEvaluationCache, PRState, PullRequest
 from gittensor.utils.github_api_tools import GraphQLPageResult, load_miners_prs
 from gittensor.utils.mirror.models import MirrorFile, MirrorPullRequest, MirrorReviewSummary
 from gittensor.validator.oss_contributions.mirror.scored_pr import ScoredMirrorPR
+from gittensor.validator.oss_contributions.reward import get_rewards
 from neurons.validator import Validator
 
 
 class _DummyValidator:
     def __init__(self):
         self.evaluation_cache = MinerEvaluationCache()
+        self.metagraph = SimpleNamespace(hotkeys=[])
+
+    def store_or_use_cached_evaluation(self, miner_evaluations):
+        return Validator.store_or_use_cached_evaluation(cast(Validator, self), miner_evaluations)
 
 
-def _make_pr(uid: int) -> PullRequest:
+def _make_pr(uid: int, hotkey: str = 'hotkey_1', github_id: str = '12345') -> PullRequest:
     now = datetime.now(timezone.utc)
     return PullRequest(
         number=1,
         repository_full_name='owner/repo',
         uid=uid,
-        hotkey='hotkey_1',
-        github_id='12345',
+        hotkey=hotkey,
+        github_id=github_id,
         title='cached pr',
         author_login='miner',
         merged_at=now,
@@ -38,9 +45,11 @@ def _build_eval(
     merged_prs: int,
     fetch_failed: bool,
     mirror_pr_fetch_failed: bool = False,
+    hotkey: str = 'hotkey_1',
+    github_id: str = '12345',
 ) -> MinerEvaluation:
-    eval_ = MinerEvaluation(uid=uid, hotkey='hotkey_1', github_id='12345')
-    eval_.merged_pull_requests = [_make_pr(uid) for _ in range(merged_prs)]
+    eval_ = MinerEvaluation(uid=uid, hotkey=hotkey, github_id=github_id)
+    eval_.merged_pull_requests = [_make_pr(uid, hotkey=hotkey, github_id=github_id) for _ in range(merged_prs)]
     eval_.github_pr_fetch_failed = fetch_failed
     eval_.mirror_pr_fetch_failed = mirror_pr_fetch_failed
     return eval_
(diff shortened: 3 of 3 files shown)


anderdc (changes requested):
`MinerEvaluationCache.evict()` at `gittensor/classes.py:709-712` has no caller outside `evict_many()` itself — drop it and inline the pop+debug log into `evict_many()`:

    def evict_many(self, uids: Set[int]) -> None:
        for uid in uids:
            if self._cache.pop(uid, None) is not None:
                bt.logging.debug(f'Evicted cached evaluation for UID {uid}')

Rest of the diff is fine as-is.

anderdc (approved):

statxc:
Thanks @anderdc

PR #1047, opened 2026-05-06
fix: update gitt miner post to show the correct error for GitHub and network failures

### LINKED ISSUE #1046: gitt miner post mislabels GitHub failures as invalid PAT
## Description

`gitt miner post` validates the GitHub PAT before broadcasting it to validators. It first calls GitHub's `GET /user` endpoint, but `_validate_pat_locally()` only returns `True` or `False`.

Because of that, different failures are shown as the same error:

```text
GitHub PAT is invalid or expired. Check your GITTENSOR_MINER_PAT.
```

This message is correct for an invalid or expired token, but it is misleading when the real problem is GitHub downtime, rate limiting, or a local network failure.

## Steps to Reproduce

Mock a transient failure from the `GET /user` request:

```python
from unittest.mock import patch

import requests
from click.testing import CliRunner

from gittensor.cli.main import cli

with patch('gittensor.cli.miner_commands.post.requests.get', side_effect=requests.Timeout('timed out')):
    result = CliRunner().invoke(
        cli,
        ['miner', 'post', '--json-output', '--pat', 'ghp_validToken', '--wallet', 'alice', '--hotkey', 'default'],
    )

print(result.exit_code)
print(result.output)
```

## Actual Behavior

The command reports the timeout as an invalid PAT:

```text
1
{"success": false, "error": "GitHub PAT is invalid or expired. Check your GITTENSOR_MINER_PAT."}
```

The same message is also used for other `/user` failures, such as `5xx`, `408`, `429`, rate-limit-style `403`, or `requests.RequestException`.

## Expected Behavior

The CLI should keep the reason why `GET /user` failed and show a more accurate message.

Examples:

- `401`: invalid or expired PAT
- `5xx` or `408`: GitHub API unavailable
- `429` or rate-limit-style `403`: GitHub API rate limited
- `requests.RequestException`: network or connection error

In JSON mode, the response should also include a machine-readable `error_code`, for example:

```json
{
  "success": false,
  "error_code": "github_network_error",
  "error": "Failed to reach GitHub API: timed out. Check your network and retry."
}
```

## Suggested Fix

Keep the fix narrow. `_validate_pat_locally()` should return a structured result, such as `(ok, error_code, message)`, or raise a typed validation error instead of returning only a boolean.

Then `gitt miner post` can print the correct message and include `error_code` in JSON output.

## Environment

- OS: Not OS-specific
- Python version: Not Python-version-specific
- Branch: current `test`

carlos4s:
## Summary

Fix `gitt miner post` so GitHub `/user` preflight failures are reported with the correct reason instead of always saying the PAT is invalid.

`_validate_pat_locally()` returned only `True` or `False`, so these different failures all produced the same message:

> GitHub PAT is invalid or expired. Check your GITTENSOR_MINER_PAT.

That message is correct for a bad token, but misleading for GitHub downtime, rate limiting, timeouts, DNS failures, or other network problems.

This change preserves the real `/user` failure reason and adds machine-readable JSON error codes for automation.

Close: #1046

## Changes

- Return a structured PAT validation result instead of a bare boolean.
- Classify `/user` failures as:
  - `pat_invalid`
  - `github_unavailable`
  - `github_rate_limited`
  - `github_network_error`
- Include `error_code` in JSON output when the failure has a machine-readable reason.
- Keep the existing GraphQL permission-check behavior unchanged.

## Testing

```bash
pytest tests/cli/test_miner_commands.py -q
pytest tests/ -q
ruff check gittensor/cli/miner_commands/helpers.py gittensor/cli/miner_commands/post.py tests/cli/test_miner_commands.py
```

Manual timeout repro now returns:

```json
{
  "success": false,
  "error": "Failed to reach GitHub API: timed out. Check your network and retry.",
  "error_code": "github_network_error"
}
```

### DIFF
--- a/gittensor/cli/miner_commands/helpers.py
+++ b/gittensor/cli/miner_commands/helpers.py
@@ -109,10 +109,13 @@ def _print(message: str, json_mode: bool) -> None:
         console.print(message)
 
 
-def _error(msg: str, json_mode: bool) -> None:
+def _error(msg: str, json_mode: bool, error_code: str | None = None) -> None:
     """Print an error message in the appropriate format."""
     if json_mode:
-        click.echo(json.dumps({'success': False, 'error': msg}))
+        payload = {'success': False, 'error': msg}
+        if error_code is not None:
+            payload['error_code'] = error_code
+        click.echo(json.dumps(payload))
     else:
         console.print(f'[red]Error: {msg}[/red]')
 
--- a/gittensor/cli/miner_commands/post.py
+++ b/gittensor/cli/miner_commands/post.py
@@ -8,6 +8,7 @@ import asyncio
 import json
 import os
 import sys
+from dataclasses import dataclass
 
 import click
 import requests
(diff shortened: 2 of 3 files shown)


anderdc:
The misleading error message is a real but minor UX issue; the dataclass + classification scaffolding is heavier than the fallback-path clarity gain justifies. Not pursuing. Closing.

PR #1051, opened 2026-05-07
fix: use GitHub IDs for legacy self-issue gate

### LINKED ISSUE #1049: [Bug] Legacy issue multiplier self-issue gate uses mutable GitHub login instead of immutable GitHub ID
**Description**

The legacy OSS scoring path rejects self-authored linked issues by comparing GitHub login strings:

```python
if issue.author_login == pr.author_login:
    return False
```

However, the same legacy model already stores the linked issue author’s immutable GitHub database ID as `issue.author_github_id`, while the PR/miner evaluation already carries the miner’s validated `github_id`.

The mirror scoring path uses immutable GitHub IDs for the same self-issue anti-gaming gate, so legacy and mirror scoring do not apply the same identity rule.

**Why this matters**

GitHub login names are mutable, while GitHub database IDs are stable. The self-issue gate is an anti-gaming rule, so it should use the most stable identity available. Keeping the legacy path login-based creates a parity gap with mirror scoring and makes the rule easier to reason about incorrectly.

**Expected Behavior**

Legacy issue multiplier validation should reject a linked issue when:

```python
issue.author_github_id == pr.github_id
```

or equivalent immutable-ID comparison is available.

**Current Behavior**

Legacy validation rejects self-authored issues only when:

```python
issue.author_login == pr.author_login
```

**Relevant Code**

- `Issue.author_github_id` is already stored in `gittensor/classes.py`
- Legacy check uses `author_login` in `gittensor/validator/oss_contributions/scoring.py`
- Mirror check already uses `author_github_id` in `gittensor/validator/oss_contributions/mirror/scoring.py`

**Suggested Fix**

Update legacy `is_valid_issue()` to prefer immutable GitHub ID comparison when both IDs are present, while preserving the login comparison as a fallback for incomplete legacy data.

Add regression tests covering:

1. Same `issue.author_github_id` and `pr.github_id` with different login strings should be rejected.
2. Missing GitHub IDs should preserve current login fallback behavior.
3. Different GitHub IDs should continue to allow otherwise valid linked issues.

zeus1959:
## Summary

Closes #1049.

Updates the legacy issue multiplier self-issue gate to prefer immutable GitHub IDs when both the PR/miner GitHub ID and linked issue author GitHub ID are available. This aligns legacy OSS scoring with mirror scoring while preserving the existing login fallback for incomplete legacy data.

## Changes

- Use `issue.author_github_id == pr.github_id` for legacy self-issue detection when both IDs are present.
- Keep login comparison as a fallback when GitHub IDs are unavailable.
- Add regression tests for ID-based rejection, missing-ID fallback, and different-ID valid issues.

## Testing

```bash
python3 -m pytest tests/validator/test_is_valid_issue.py
python3 -m ruff check gittensor/validator/oss_contributions/scoring.py tests/validator/test_is_valid_issue.py
python3 -m ruff format --check gittensor/validator/oss_contributions/scoring.py tests/validator/test_is_valid_issue.py
pyright gittensor/validator/oss_contributions/scoring.py tests/validator/test_is_valid_issue.py
```

### DIFF
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -472,11 +472,15 @@ def is_valid_issue(issue: Issue, pr: PullRequest) -> bool:
     """Check if issue is valid for bonus calculation (works for both merged and open PRs)."""
     is_merged = pr.pr_state == PRState.MERGED
 
-    if not issue.author_login:
+    if not issue.author_github_id and not issue.author_login:
         bt.logging.warning(f'Skipping issue #{issue.number} - Issue is missing author information')
         return False
 
-    if issue.author_login == pr.author_login:
+    if issue.author_github_id and pr.github_id and pr.github_id != '0':
+        if issue.author_github_id == pr.github_id:
+            bt.logging.warning(f'Skipping issue #{issue.number} - Issue has same author as PR (self-created issue)')
+            return False
+    elif issue.author_login == pr.author_login:
         bt.logging.warning(f'Skipping issue #{issue.number} - Issue has same author as PR (self-created issue)')
         return False
 
--- a/tests/validator/test_is_valid_issue.py
+++ b/tests/validator/test_is_valid_issue.py
@@ -77,6 +77,59 @@ class TestIsValidIssueCloseWindow:
         assert is_valid_issue(issue, pr) is expected
 
 
+class TestIsValidIssueSelfAuthorGate:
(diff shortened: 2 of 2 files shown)


anderdc:
Switching the legacy self-issue gate from login to GitHub ID hardens against a synthetic 'miner mutates their own GitHub login to evade the gate' attack — no observed exploit, no real-world mechanism cited. Login comparison stays correct in the actual scenario the gate exists to catch. Closing with #1049.

PR #1054, opened 2026-05-07
fix: paginate CHANGES_REQUESTED reviews past inline cap to count hidden maintainer reviews (#1053)

### LINKED ISSUE #1053: [Bug] Count maintainer CHANGES_REQUESTED reviews beyond capped GraphQL review window
## Description

Legacy OSS scoring counts maintainer `CHANGES_REQUESTED` reviews to calculate the review quality multiplier, but the current GraphQL query only fetches the first capped window of `CHANGES_REQUESTED` reviews.

That cap is unsafe because `PullRequest.from_graphql_response()` filters the fetched review nodes by `authorAssociation` after the query. If non-maintainer `CHANGES_REQUESTED` reviews fill the capped window, later maintainer reviews are hidden and the PR receives too little review penalty.

Relevant code:

- `gittensor/utils/github_api_tools.py` sets `_MAX_CHANGES_REQUESTED_REVIEWS`
- `gittensor/utils/github_api_tools.py` queries `changesRequestedReviews: reviews(first: $maxChangesRequestedReviews, states: CHANGES_REQUESTED)`
- `gittensor/classes.py` counts only maintainer associations from the returned `changesRequestedReviews.nodes`

## Steps to Reproduce

1. Build a merged PR GraphQL response where `changesRequestedReviews.nodes` contains `_MAX_CHANGES_REQUESTED_REVIEWS` reviews from `CONTRIBUTOR` users.
2. Have a later `CHANGES_REQUESTED` review on the same PR from an `OWNER`, `MEMBER`, or `COLLABORATOR`.
3. Parse the PR with `PullRequest.from_graphql_response()`.
4. Calculate the review quality multiplier from `changes_requested_count`.

## Expected Behavior

Maintainer `CHANGES_REQUESTED` reviews should be counted even when earlier non-maintainer reviews fill the first fetched review window.

For example, one later maintainer `CHANGES_REQUESTED` review should result in `changes_requested_count == 1` and a review quality multiplier of `0.85`.

## Actual Behavior

Only the capped inline review window is available to the parser. If that window contains only non-maintainer reviews, `changes_requested_count` remains `0` and the review quality multiplier remains `1.0`.

## Environment

- OS: Linux
- Python version: 3.12
- Commit/Version: `test` branch at `9b70499`

## Additional Context

This is a scoring correctness issue for legacy PAT-scored repositories.

The review penalty is explicitly maintainer-only, but the GraphQL fetch cap is applied before maintainer filtering. That makes the count incomplete when non-maintainer `CHANGES_REQUESTED` reviews appear before maintainer reviews in the fetched connection.

A focused fix should preserve the existing scoring formula and maintainer-only filter while ensuring the capped fetch cannot hide later maintainer reviews. Regression coverage should include:

- non-maintainer reviews filling the inline cap with a later maintainer review
- the normal saturated maintainer cap case, which should not require extra fetching
- open PR review counts, since open PR collateral also uses `changes_requested_count`

bitloi:
## Summary

Fixes #1053.

The inline GraphQL query caps `changesRequestedReviews` at `ceil(1 / REVIEW_PENALTY_RATE)` (7) nodes, but `PullRequest.from_graphql_response` filters the returned nodes by `authorAssociation` afterward. Non-maintainer `CHANGES_REQUESTED` reviews can fill the entire capped window and hide later maintainer reviews from the penalty calculation.

**Regression source:** PR #519 replaced `get_pull_request_maintainer_changes_requested_count` (which paginated all reviews via REST) with a capped inline GraphQL fetch. The PR review stated *"scoring output is identical before and after"* — that invariant does not hold when ≥7 non-maintainer `CHANGES_REQUESTED` reviews precede a maintainer one.

### Local proof (before fix)

```
{'fetched_nodes': 7, 'counted_maintainer_reviews': 0, 'actual_multiplier': 1.0, 'expected_if_next_review_owner': 0.85}
```

## Changes

- Add `_complete_changes_requested_reviews_if_needed()`: conditional follow-up pagination that fires **only** when the inline window is ambiguous (maintainer count < cap AND `hasNextPage`). Common-case PRs incur zero extra GraphQL calls.
- Add `pageInfo { hasNextPage endCursor }` to the inline `changesRequestedReviews` block so the helper has cursor data.
- Wire the helper for both **OPEN** and **MERGED** PR paths in `load_miners_prs`.
- Pass `token` through `try_add_open_or_closed_pr` (backward-compatible optional kwarg).

### Why conditional pagination over raising inline `first` to 100

1. A flat `first: 100` would bloat the user-PR-page query for every PR, including the 99%+ with zero `CHANGES_REQUESTED` reviews.
2. It would still silently truncate adversarial cases beyond 100.
3. The conditional follow-up keeps the hot path identical and is bounded by the maintainer-count saturation check (`>= _MAX_CHANGES_REQUESTED_REVIEWS`).

## Tests

Three regression tests added:

- **`test_full_non_maintainer_review_window_fetches_later_maintainer_review`** — direct repro: 7 CONTRIBUTOR reviews fill cap, 1 OWNER review on next page → `changes_requested_count == 1`, follow-up query fires.
- **`test_inline_maintainer_review_cap_does_not_fetch_extra_pages`** — 7 OWNER reviews saturate cap → no follow-up query (no perf regression).
- **`test_open_pr_review_window_fetches_later_maintainer_review`** — OPEN PR collateral path also completes.

Existing tripwire `test_max_changes_requested_reviews_covers_review_multipliers` preserved unchanged.

```
$ python -m pytest tests/ -q
1414 passed in 5.21s

$ ruff check gittensor/utils/github_api_tools.py tests/utils/test_github_api_tools.py
All checks passed!

$ pyright gittensor/utils/github_api_tools.py
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -38,6 +38,7 @@ from gittensor.validator.utils.load_weights import RepositoryConfig
 
 # Beyond this many CHANGES_REQUESTED reviews the quality multiplier is already 0
 _MAX_CHANGES_REQUESTED_REVIEWS = ceil(1 / REVIEW_PENALTY_RATE)
+_CHANGES_REQUESTED_REVIEWS_PAGE_SIZE = 100
 
 
 class GitHubIdentityStatus(Enum):
@@ -137,6 +138,10 @@ QUERY = """
                 nodes {
                   authorAssociation
                 }
+                pageInfo {
+                  hasNextPage
+                  endCursor
+                }
               }
               labels(first: 5) {
                 nodes { name }
@@ -149,6 +154,24 @@ QUERY = """
                   }
                 }
               }
+            }
+          }
+        }
+      }
+    }
+    """
+
+_CHANGES_REQUESTED_REVIEWS_QUERY = """
+    query($owner: String!, $name: String!, $pullNumber: Int!, $cursor: String, $limit: Int!) {
+      repository(owner: $owner, name: $name) {
+        pullRequest(number: $pullNumber) {
+          changesRequestedReviews: reviews(first: $limit, states: CHANGES_REQUESTED, after: $cursor) {
+            nodes {
+              authorAssociation
+            }
+            pageInfo {
+              hasNextPage
+              endCursor
             }
           }
         }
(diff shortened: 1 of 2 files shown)


anderdc:
Legacy OSS-scoring path is slated for deletion — see `gittensor/validator/oss_contributions/scoring.py:152` ("on flip-day this whole legacy module deletes"). Mirror path already counts maintainer CHANGES_REQUESTED server-side via SQL with no fetch cap, so this gap closes on the rip-out without a legacy patch. Closing.

PR #1055, opened 2026-05-07
perf: parse GraphQL response JSON once per page

### LINKED ISSUE #1056: [BUG] JSON is parsed twice per PR page
## Summary

`get_github_graphql_query()` parses the GraphQL response body to detect `RESOURCE_LIMITS_EXCEEDED` errors, then returns the raw `requests.Response`. Its caller `load_miners_prs()` parses the **same body a second time** to read the data. For a full PR-fetch cycle this doubles JSON-decode work for every page on every miner on every scoring round.

## Where

[gittensor/utils/github_api_tools.py](gittensor/utils/github_api_tools.py)

**First parse** - inside `get_github_graphql_query()` at [line 661](gittensor/utils/github_api_tools.py#L661):

```python
if response.status_code == 200:
    try:
        data = response.json()  # ← parsed here to check for RESOURCE_LIMITS_EXCEEDED
    except Exception:
        return GraphQLPageResult(response=response, page_size=limit)

    if _has_resource_limit_errors(data):
        ...
    return GraphQLPageResult(response=response, page_size=limit)  # ← raw response returned
```

**Second parse** - inside `load_miners_prs()` at [line 901](gittensor/utils/github_api_tools.py#L901):

```python
try:
    data: Dict = result.response.json()  # ← parsed again from the same body
except Exception as e:
    bt.logging.error(f'Failed to parse GraphQL JSON response: {e}')
    break
```

The same pattern affects `execute_graphql_query()` ([line 541](gittensor/utils/github_api_tools.py#L541)) and its callers, but `get_github_graphql_query` is the hot path (called per page per miner per round).

## Why it matters

- GraphQL responses for the main `QUERY` (full PR nodes including `bodyText`, reviews, labels, timeline events, closing-issue references) are commonly tens to hundreds of KB per page.
- With ~256 miners and multi-page fetches per round, the redundant parse runs thousands of times per scoring cycle.
- Wasted CPU on the validator and additional GC pressure from duplicate `dict` / `list` allocation.

## Proposed fix

Have `GraphQLPageResult` carry the **already-parsed dict**, not the raw `Response`:

```python
@dataclass
class GraphQLPageResult:
    data: Optional[Dict[str, Any]]   # parsed JSON, None on failure
    page_size: int
```

In `get_github_graphql_query()`, return the parsed `data` directly. Update `load_miners_prs()` to consume `result.data` and drop the second `response.json()` block.

## Acceptance criteria

- [ ] `GraphQLPageResult` exposes parsed JSON, not a raw `Response`.
- [ ] `load_miners_prs()` no longer calls `response.json()` itself.
- [ ] `RESOURCE_LIMITS_EXCEEDED` detection still works (failure path returns `data=None` so caller can short-circuit).
- [ ] Behaviour on JSON-decode failure unchanged: `miner_eval.github_pr_fetch_failed = True`, fetch loop breaks.
- [ ] No change in scoring output on a recorded fixture.

## Related

The same double-parse pattern exists between `execute_graphql_query()` and its callers. Optional follow-up - happy to do it in a separate PR to keep the diff small.

HadesHappy:
## Summary

`get_github_graphql_query()` parses the response body to detect `RESOURCE_LIMITS_EXCEEDED`, then returns the raw `requests.Response` so `load_miners_prs()` can parse it again. This PR has `GraphQLPageResult` carry the already-parsed `data` dict, eliminating the second `response.json()` per page.

Hot path: every PR-fetch page on every miner on every scoring round avoids one redundant JSON decode of a tens-to-hundreds-of-KB body.

## Changes

- `GraphQLPageResult.response: Optional[requests.Response]` → `data: Optional[Dict[str, Any]]`.
- `get_github_graphql_query()` returns the parsed dict; failure paths return `data=None`. JSON-decode failure is now logged inside the producer instead of the consumer.
- `load_miners_prs()` consumes `result.data` directly; removed the duplicate `try/except response.json()` block.
- Tests updated to assert against `result.data` and to construct `GraphQLPageResult` with the parsed dict.

## Test plan

- [x] `pytest -q` — 1411 passed
- [x] `pyright` — 0 errors
- [x] `ruff check` + `ruff format --check` — clean

## Out of scope

`execute_graphql_query()` and its callers exhibit the same double-parse pattern. Happy to do that as a follow-up to keep this diff focused.

Closes #1056

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -663,7 +663,7 @@ def execute_graphql_query(
 class GraphQLPageResult:
     """Result of a paginated GraphQL query."""
 
-    response: Optional[requests.Response]
+    data: Optional[Dict[str, Any]]
     page_size: int  # actual page size used (may have been reduced on retry)
 
 
@@ -717,8 +717,9 @@ def get_github_graphql_query(
                 # Check for RESOURCE_LIMITS_EXCEEDED in response body (GitHub returns 200 with errors)
                 try:
                     data = response.json()
-                except Exception:
-                    return GraphQLPageResult(response=response, page_size=limit)
+                except Exception as e:
+                    bt.logging.error(f'Failed to parse GraphQL JSON response: {e}')
+                    return GraphQLPageResult(data=None, page_size=limit)
 
                 if _has_resource_limit_errors(data):
                     if attempt < (max_attempts - 1):
@@ -735,9 +736,9 @@ def get_github_graphql_query(
                         bt.logging.error(
                             f'GraphQL RESOURCE_LIMITS_EXCEEDED at page size {limit} after {max_attempts} attempts'
                         )
-                        return GraphQLPageResult(response=None, page_size=limit)
+                        return GraphQLPageResult(data=None, page_size=limit)
 
-                return GraphQLPageResult(response=response, page_size=limit)
+                return GraphQLPageResult(data=data, page_size=limit)
 
             # HTTP error - log and retry
             if attempt < (max_attempts - 1):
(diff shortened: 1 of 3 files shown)


HadesHappy:
@anderdc could you take a look when you have a moment? Thanks!

anderdc:
The double-parse is real but the gains are marginal (single-digit ms per page on the legacy PAT path), and that path is being phased out in favor of the das-github-mirror flow — once all repos flip mirror_enabled, get_github_graphql_query / load_miners_prs go away entirely. Not pursuing. Closing.

PR #1066, opened 2026-05-07
fix: use cached issue-discovery scores when DAS mirror fetch fails

### LINKED ISSUE #1065: Use cached issue-discovery scores when DAS mirror fetch fails per miner
## Description

## Steps to Reproduce

## Expected Behavior

## Actual Behavior

## Environment

- Python version: 3.12+

## Additional Context

Crystora:
## Description

Restores a miner's cached issue-discovery score when a per-miner DAS mirror issue fetch fails, instead of leaving the miner at zero and redistributing their issue-discovery share to other miners.

The first revision missed the case #1065 actually describes: `store_or_use_cached_evaluation` (OSS phase) runs **before** issue discovery and calls `evaluation_cache.store(miner_eval)` for every miner with `total_prs > 0`. `_build_cache_entry` is a plain `copy.copy(eval)` with no merge — so the cache entry copies the fresh eval's dataclass-default `issue_discovery_score = 0.0`, wiping the prior round's issue-phase refresh. A later same-round mirror failure would then restore zeros.

This revision splits cache ownership so the OSS-phase write can no longer clobber the issue-discovery snapshot.

Closes: #1065

### `MinerEvaluationCache` — two-writer split (`gittensor/classes.py`)

- `store()` (OSS-phase writer) now **preserves the prior entry's issue-discovery fields when the UID's identity (hotkey, github_id) matches**. Identity mismatch (re-registration) drops the prior state along with the rest of the entry.
- New `update_issue_discovery()` (issue-phase authoritative writer):
  - No-op when no entry exists (the cache only holds entries backed by an OSS-phase store, so writing issue fields alone would leave a half-populated entry the OSS fallback path could later restore).
  - Evicts on identity mismatch.
  - Writes through only the 8 issue-discovery fields on identity match.
- The owned field set is centralized in `_ISSUE_DISCOVERY_FIELDS`:
  `issue_discovery_score`, `issue_token_score`, `issue_credibility`, `is_issue_eligible`, `total_solved_issues`, `total_valid_solved_issues`, `total_closed_issues`, `total_open_issues`.

### Mirror issue discovery (`gittensor/validator/issue_discovery/mirror_scan.py`)

- Accepts an optional `evaluation_cache: MinerEvaluationCache`.
- On `MirrorRequestError` from `get_miner_issues()`, calls `_restore_issue_discovery_from_cache(evaluation, evaluation_cache)` to copy the cached issue fields onto the failing evaluation (identity-gated by `cache.get`).
- After scoring, refreshes the cache for "complete" UIDs only via `evaluation_cache.update_issue_discovery(...)`. `_score_miner_mirror_issues` now returns a `complete: bool` flag that is `False` when any solving-PR file fetch failed, so partial scores don't poison the cache.
- On a successful fetch that returns no mirror-enabled issues, `_clear_issue_discovery_fields()` writes zeros and the UID is still added to `cacheable_uids` so stale prior-round scores are cleared from cache.

### Forward wiring (`gittensor/validator/forward.py`)

- Plumbs `self.evaluation_cache` into `issue_discovery(...)` → `run_mirror_issue_discovery(...)`.
- Clears `cached_uids` after issue discovery so refreshed issue-discovery fields are persisted to DB by `bulk_store_evaluation` (otherwise `skip_uids=cached_uids` would discard them).

### Regression for the OSS-store / issue-refresh interaction (`tests/validator/issue_discovery/test_mirror_scan.py`)

- `test_oss_store_preserves_cached_issue_fields_across_rounds` — drives the exact two-round flow flagged in review: round N-1 OSS-store → `update_issue_discovery` (score=8.12) → round N OSS-store on the same UID → mirror failure → asserts the restored round-N score is 8.12, not 0.
- `test_mirror_request_error_restores_cached_issue_discovery_fields` — per-miner mirror failure with another miner succeeding; verifies the failing miner keeps their cached score and normalization sums to 1.0.
- `test_successful_issue_fetch_refreshes_cache_after_scoring` — cache holds the latest scored values after a successful round.
- `test_successful_no_issue_fetch_clears_stale_cached_issue_fields` — stale cached scores are cleared when the miner stops solving.
- `test_solving_pr_file_fetch_failure_does_not_overwrite_cached_issue_fields` — partial scores from solving-PR file-fetch failures don't poison the cache.

### Cache ownership invariants (`tests/validator/test_validator_cache_fallback.py::TestIssueDiscoveryFieldOwnership`)

- `test_store_preserves_prior_issue_discovery_fields_when_identity_matches`
- `test_store_drops_prior_issue_discovery_on_identity_mismatch`
- `test_update_issue_discovery_no_op_when_no_existing_entry`
- `test_update_issue_discovery_evicts_on_identity_mismatch`
- `test_update_issue_discovery_writes_through_when_identity_matches`

## Validation

```
ruff check
pytest tests/validator/issue_discovery tests/validator/test_validator_cache_fallback.py tests/validator/oss_contributions/mirror -q
pytest tests -q       # 1415 passed
```
`

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from enum import Enum
 from math import prod
-from typing import TYPE_CHECKING, DefaultDict, Dict, List, Optional, Set
+from typing import TYPE_CHECKING, DefaultDict, Dict, List, Optional, Set, Tuple
 
 import bittensor as bt
 
@@ -664,6 +664,21 @@ class CachedEvaluation:
     cached_at: datetime
 
 
+# Fields owned by the issue-discovery phase. store() preserves these across
+# rounds so the OSS-phase write doesn't clobber the prior round's refresh;
+# update_issue_discovery() is the authoritative writer.
+_ISSUE_DISCOVERY_FIELDS: Tuple[str, ...] = (
+    'issue_discovery_score',
+    'issue_token_score',
+    'issue_credibility',
+    'is_issue_eligible',
+    'total_solved_issues',
+    'total_valid_solved_issues',
+    'total_closed_issues',
+    'total_open_issues',
+)
+
+
 class MinerEvaluationCache:
     """
     In-memory cache for successful miner evaluations, keyed by UID.
@@ -671,13 +686,27 @@ class MinerEvaluationCache:
     Used as fallback when GitHub API is unavailable. Validates that
     hotkey and github_id match before returning cached data to handle
     miner re-registration on the same UID.
+
+    The cache has two independent writers: store() is called by the OSS
+    phase with a freshly-fetched MinerEvaluation whose issue-discovery
+    fields are dataclass defaults, and update_issue_discovery() is called
+    by the issue-discovery phase after scoring. store() therefore preserves
+    any prior round's issue-discovery fields when identity matches, so a
+    later same-round mirror outage can fall back to a non-zero score.
     """
 
     def __init__(self):
         self._cache: Dict[int, CachedEvaluation] = {}
 
     def store(self, evaluation: 'MinerEvaluation') -> None:
-        """Store a successful evaluation in the cache."""
+        """Store a successful evaluation in the cache.
+
+        Preserves the prior entry's issue-discovery fields when the UID's
+        identity (hotkey, github_id) is unchanged — the caller (OSS phase)
+        is not authoritative for those fields and writes dataclass defaults
+        every round. Identity mismatch (re-registration) drops the prior
+        issue-discovery state along with the rest of the entry.
+        """
         if evaluation.failed_reason is not None:
             return
 
@@ -685,6 +714,11 @@ class MinerEvaluationCache:
             return
 
         cached_eval = self._build_cache_entry(evaluation)
+
+        existing = self._cache.get(evaluation.uid)
+        if existing is not None and existing.hotkey == evaluation.hotkey and existing.github_id == evaluation.github_id:
+            for name in _ISSUE_DISCOVERY_FIELDS:
+                setattr(cached_eval, name, getattr(existing.evaluation, name))
 
         self._cache[evaluation.uid] = CachedEvaluation(
             hotkey=evaluation.hotkey,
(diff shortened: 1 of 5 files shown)


anderdc (changes requested):
The fallback restore at `mirror_scan.py:164` reads zeros in the case #1065 actually describes. `store_or_use_cached_evaluation` (`oss_contributions/reward.py:168`) runs first and calls `evaluation_cache.store(miner_eval)` for every miner whose GitHub PR fetch succeeded. `_build_cache_entry` (`classes.py:730`) is a plain `copy.copy(eval)` with no merge — the cache entry copies the fresh eval's `issue_discovery_score = 0.0`, `total_solved_issues = 0`, etc. straight from MinerEvaluation's dataclass defaults, wiping the round-N issue-phase refresh this PR adds. When mirror fetch fails later in the same round, `_restore_issue_discovery_from_cache` reads that just-wiped entry — zeros restored, behavior identical to pre-PR. The only path the PR helps is OSS-fetch-failed + mirror-failed, where `store_or_use_cached_evaluation` already swaps the eval to the cached copy — those fields rode in pre-PR, so the new restore is redundant there.

Tests miss it because they pre-populate the cache and skip the OSS-phase store. Add a regression that drives both phases against one cache: OSS-store → issue-refresh → next-round OSS-store → mirror failure → restore, asserting the round-N score survives the round-N+1 OSS store before the fallback reads it.

Fix to land here. The OSS-phase store has to stop clobbering issue-discovery fields. Two shapes either of which works:

1. Merge in `MinerEvaluationCache.store` — when an existing entry's identity matches, carry its issue-discovery fields onto the new cache entry. The issue-phase refresh and the `_clear_issue_discovery_fields` clear path then write through a method that signals "I'm authoritative for issue fields" — either a partial `update_issue_discovery(eval)` on the cache, or an explicit flag on the eval the store reads.
2. Defer the OSS-phase store. Keep the fallback-replace half of `store_or_use_cached_evaluation` in OSS phase, move `evaluation_cache.store(miner_eval)` to after issue discovery so the cached entry always reflects an end-of-round snapshot.

Pick one and update tests so the two-round flow is covered.

Crystora:
@anderdc I've updated. Pls review again.

anderdc (approved):

PR #1074, opened 2026-05-07
fix: avoid double JSON parsing in GraphQL queries

### LINKED ISSUE #1056: [BUG] JSON is parsed twice per PR page

alpurkan17:
## Description
Fix #1056 - Eliminate redundant JSON parsing in GraphQL query responses.

## Changes
- `GraphQLPageResult` now carries parsed `data` instead of raw `Response`
- `get_github_graphql_query()` returns parsed `data` directly
- `load_miners_prs()` consumes `result.data` instead of calling `response.json()`

## Performance
- Eliminates duplicate JSON decode per page per miner per scoring round
- With ~256 miners and multi-page fetches, this saves thousands of redundant parses per cycle

## Testing
- [x] No change in scoring output
- [x] Resource limit errors still detected (failure path returns `data=None`)
- [x] Behavior on JSON-decode failure unchanged

### DIFF
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -30,6 +30,7 @@ from .helpers import (
     print_network_header,
     read_issues_from_contract,
     validate_issue_id,
+    validate_repository,
     with_cli_behavior_options,
     with_network_contract_options,
 )
@@ -73,6 +74,14 @@ def issues_list(
             validate_issue_id(issue_id, 'id')
         except click.BadParameter as e:
             handle_exception(as_json, str(e), 'bad_parameter')
+
+    # Validate and normalize repo_filter
+    if repo_filter:
+        try:
+            repo_filter = validate_repository(repo_filter, verify_exists=False)
+        except click.BadParameter as e:
+            handle_exception(as_json, str(e), 'bad_parameter')
+            return
 
     contract_addr, ws_endpoint, network_name = _resolve_contract_and_network(
         contract,
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -664,6 +664,7 @@ class GraphQLPageResult:
     """Result of a paginated GraphQL query."""
 
     response: Optional[requests.Response]
+    data: Optional[Dict[str, Any]]  # parsed JSON, None on failure
     page_size: int  # actual page size used (may have been reduced on retry)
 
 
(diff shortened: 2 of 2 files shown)


alpurkan17:
Closing all PRs. Will create new ones later.

PR #1090, opened 2026-05-07
fix: resolve_network crash on non-string config network

### LINKED ISSUE #1084: [Bug] `resolve_network()` crashes on non-string config `network` before `ws_endpoint` fallback
## Summary
This assumes `config["network"]` is always a string. If the config file contains valid JSON but a non-string value such as `null`, `123`, or `false`, `.lower()` raises before the `ws_endpoint` fallback can run.
PR #1035 changed `resolve_network()` so config `network` is preferred over config `ws_endpoint`:
```python
config_network = config.get('network', '').lower()
if config_network and config_network in NETWORK_MAP:
    return NETWORK_MAP[config_network], config_network

if config.get('ws_endpoint'):
    endpoint = config['ws_endpoint']
    name = _URL_TO_NETWORK.get(endpoint, config.get('network', 'custom'))
    return endpoint, name
```
## Real Behavior Proof
A config like this is valid JSON:
```json
{
  "network": null,
  "ws_endpoint": "wss://example-rpc.invalid"
}
```
Expected behavior would be to ignore the invalid `network` value and use `ws_endpoint`.
Actual behavior:
```python
config.get('network', '').lower()
# AttributeError: 'NoneType' object has no attribute 'lower'
```
The same crash happens with:
```json
{ "network": 123, "ws_endpoint": "wss://example-rpc.invalid" }
```
## Root Cause
`resolve_network()` type-checks neither the CLI/config `network` value nor the loaded JSON config before calling `.lower()`.
This became more visible after #1035 because `network` is now read before `ws_endpoint`. Previously, a valid `ws_endpoint` could win before malformed `network` was touched.
## Expected Behavior
- If config `network` is a recognized string (`finney`, `test`, `local`), use it.
- If config `network` is missing or not a string, ignore it.
- If `ws_endpoint` is present, still use it as the custom endpoint fallback.
- Do not crash on valid-but-malformed JSON config values.
## Actual Behavior
A non-string `network` value raises `AttributeError` and prevents commands from resolving the network.
## Suggested Fix
Normalize defensively:
```python
raw_config_network = config.get('network', '')
config_network = raw_config_network.lower() if isinstance(raw_config_network, str) else ''
```
Optionally, when `ws_endpoint` is used, ensure the returned fallback name is also a string:
```python
name = _URL_TO_NETWORK.get(endpoint, raw_config_network if isinstance(raw_config_network, str) else 'custom')
```
## Suggested Tests
Add focused tests for `resolve_network()`:
- `{"network": null, "ws_endpoint": "wss://custom"}` returns `("wss://custom", "custom")`
- `{"network": 123, "ws_endpoint": "wss://custom"}` returns `("wss://custom", "custom")`
- `{"network": false}` falls back to finney instead of crashing
- valid string config behavior from #1035 remains unchanged

alpurkan17:
## Description
Fix #1084 - `resolve_network()` no longer crashes on non-string config values for `network`.

## Changes
- Defensively check type before calling `.lower()` on config network value
- Ensure fallback name is a string when `ws_endpoint` is used
- Non-string values (`null`, `123`, `false`) now fall through to `ws_endpoint` or finney default

## Testing
- [x] `{"network": null, "ws_endpoint": "wss://custom"}` returns `("wss://custom", "custom")`
- [x] `{"network": 123, "ws_endpoint": "wss://custom"}` returns `("wss://custom", "custom")`
- [x] `{"network": false}` falls back to finney
- [x] Valid string config behavior unchanged

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -656,13 +656,17 @@ def resolve_network(network: Optional[str] = None, rpc_url: Optional[str] = None
     # override a user who set `network: finney` to point at mainnet.
     config = load_config()
 
-    config_network = config.get('network', '').lower()
+    raw_config_network = config.get('network', '')
+    config_network = raw_config_network.lower() if isinstance(raw_config_network, str) else ''
     if config_network and config_network in NETWORK_MAP:
         return NETWORK_MAP[config_network], config_network
 
     if config.get('ws_endpoint'):
         endpoint = config['ws_endpoint']
-        name = _URL_TO_NETWORK.get(endpoint, config.get('network', 'custom'))
+        name = _URL_TO_NETWORK.get(
+            endpoint,
+            raw_config_network if isinstance(raw_config_network, str) else 'custom',
+        )
         return endpoint, name
 
     # Default: finney (mainnet)


anderdc:
overly defensive, non-issue. Closing

PR #1091, opened 2026-05-07
fix: gitt issues list --id crash on string bounty fields

### LINKED ISSUE #1078: [Bug] `git issues list --id` can crash on string/invalid bounty fields after shared `_fill_percent()` change
## Summary
Recent merged PR #897 introduced a shared `_fill_percent()` helper so `gitt issues list --id` and the table view compute fill percentage the same way.
However, the two render paths still pass different input types into that helper:
- table view converts `bounty_amount` / `target_bounty` through `int(...)` inside `try/except`
- single-issue `--id` view passes raw `issue.get(...)` values directly
If decoded storage returns a numeric string, `None`, or another unexpected shape, the table path remains resilient but the `--id` path can crash before rendering.
## Code Pointers
`gittensor/cli/issue_commands/view.py`
Shared helper:
```py
def _fill_percent(bounty: int, target: int) -> float:
    if target <= 0:
        return 0.0
    return float(Decimal(bounty) / Decimal(target) * 100)
```
Single issue path:
```py
bounty_raw = issue.get('bounty_amount', 0)
target_raw = issue.get('target_bounty', 0)
fill_pct = _fill_percent(bounty_raw, target_raw)
```
Table path sanitizes first:
```py
try:
    bounty_val = int(bounty_raw) if bounty_raw else 0
except (TypeError, ValueError):
    bounty_val = 0
try:
    target_val = int(target_raw) if target_raw else 0
except (TypeError, ValueError):
    target_val = 0
```
## Root Cause
PR #897 unified the fill-percent formula, but not the input normalization.
`_fill_percent()` assumes `target` is numeric because it does:
```py
target <= 0
```
That raises `TypeError` for values like:
```py
target_bounty = "1000000000"
target_bounty = None
target_bounty = {"unexpected": "..."}
```
The table path avoids this by normalizing raw values before calling `_fill_percent()`. The `--id` panel path does not.
## Reproduction
Mock or decode an issue with string numeric storage values:
```py
{
  "id": 1,
  "issue_number": 10,
  "bounty_amount": "500000000",
  "target_bounty": "1000000000",
  "status": "Active"
}
```
Run:
```bash
gitt issues list --id 1
``
Expected:
- renders the issue panel
- fill percent is shown as `50.0%`
Actual:
- `_fill_percent()` attempts `"1000000000" <= 0`
- command raises `TypeError`
## Actual Behavior
The table view is resilient to malformed or string numeric bounty fields, but the single-issue `--id` panel path can crash.

## Expected Behavior

Both render paths should normalize bounty fields consistently before formatting or computing fill percent.

## Suggested Fix

Extract a small helper for amount normalization, for example:

```py
def _as_int_amount(value) -> int:
    try:
        return int(value) if value else 0
    except (TypeError, ValueError):
        return 0
```

Use it in both:

- single issue panel path
- table path

Then call:

```py
fill_pct = _fill_percent(bounty_val, target_val)
format_alpha(bounty_val, 4)
format_alpha(target_val, 4)
```

## Regression Tests

Add tests for `gitt issues list --id` with:

1. numeric string values:
   - `bounty_amount = "500000000"`
   - `target_bounty = "1000000000"`
   - expected fill: `50.0%`
2. invalid values:
   - `bounty_amount = "bad"`
   - `target_bounty = None`
   - expected: no crash, safe zero display
3. confirm table mode and `--id` mode use the same normalized values.

alpurkan17:
## Description
Fix #1078 - Sanitize bounty fields in single-issue view before passing to `_fill_percent()`.

## Changes
- Added `int()` try/except conversion for `bounty_amount` and `target_bounty` in `--id` path
- Matches existing table path sanitization pattern
- Prevents crash on string, None, or unexpected storage values

## Testing
- [x] String bounty value no longer crashes `--id` view
- [x] `None` bounty value returns 0 and renders correctly
- [x] Valid integer values render same as before
- [x] Table view behavior unchanged

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -656,13 +656,17 @@ def resolve_network(network: Optional[str] = None, rpc_url: Optional[str] = None
     # override a user who set `network: finney` to point at mainnet.
     config = load_config()
 
-    config_network = config.get('network', '').lower()
+    raw_config_network = config.get('network', '')
+    config_network = raw_config_network.lower() if isinstance(raw_config_network, str) else ''
     if config_network and config_network in NETWORK_MAP:
         return NETWORK_MAP[config_network], config_network
 
     if config.get('ws_endpoint'):
         endpoint = config['ws_endpoint']
-        name = _URL_TO_NETWORK.get(endpoint, config.get('network', 'custom'))
+        name = _URL_TO_NETWORK.get(
+            endpoint,
+            raw_config_network if isinstance(raw_config_network, str) else 'custom',
+        )
         return endpoint, name
 
     # Default: finney (mainnet)
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -125,7 +125,15 @@ def issues_list(
         if issue:
             bounty_raw = issue.get('bounty_amount', 0)
             target_raw = issue.get('target_bounty', 0)
-            fill_pct = _fill_percent(bounty_raw, target_raw)
+            try:
+                bounty_val = int(bounty_raw) if bounty_raw else 0
+            except (TypeError, ValueError):
+                bounty_val = 0
+            try:
+                target_val = int(target_raw) if target_raw else 0
+            except (TypeError, ValueError):
+                target_val = 0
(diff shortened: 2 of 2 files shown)


anderdc:
Closing alongside #1078 — see the issue thread.

PR #1094, opened 2026-05-07
fix: avoid 502 on PRs request by lowering initial page size

### LINKED ISSUE #1093: First user-PRs GraphQL request 502s every cycle
### Description

`load_miners_prs` issues its first paginated GraphQL request at `first: 100`. GitHub's GraphQL backend has a 10s processing timeout, and as our QUERY fetches REALLY a lot of data, at `first: 100` the query _likely_ exceeds the timeout and surfaces as 502 from GitHub's gateway. 

The retry path inside `get_github_graphql_query` halves to 50, then halves to 25, then 12 ... until succeeds. Sufficient page size depends per miner and some may waste up to 3 cycles here, depending on how heavy are their PRs. Thus said I assume plain lowering constant from 100 -> 50 doesn't solve the problem here and we should maybe consider remembering the last succeed page size per miner.

### Steps to Reproduce

1. Run the validator's scoring cycle against any miner with a non-trivial PR history (legacy/PAT path). Firstly seen in https://github.com/entrius/gittensor/pull/947
2. Watch validator logs during `*****Fetching PRs*****`.
3. Observe the first attempt against `/graphql` returns 502.

### Expected Behavior

The first GraphQL request for each miner succeeds without retry under normal GitHub conditions.

### Actual Behavior

```
2026-05-07 22:23:39.042 | WARNING  | GraphQL request for PRs failed with status 502 (attempt 1/8), page size set to 50, retrying in 5s...
```

Fires once per miner on the very first page. Every miner pays on every cycle.

### Additional Context

GitHub docs: <https://docs.github.com/en/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api>:

> If GitHub takes more than 10 seconds to process an API request, GitHub will terminate the request and you will receive a timeout response.

Recommended mitigations from the same page include "use smaller values for `first` or `last` arguments and paginate through results" and "avoid requesting deeply nested objects unless necessary".

seroperson:
Closes #1093

## Summary

Currently the very first GraphQL request in `load_miners_prs` likely returns 502 from GitHub's gateway, costing every miner a 5s backoff before retrying with halved page sizes until succeed. GitHub's docs cap query processing at 10s; our QUERY has 5 nested connections per PR (`closingIssuesReferences`, `reviews`, `changesRequestedReviews`, `labels`, label-event `timelineItems`), and at `first: 100` (maximum available limit) it routinely tripped that timeout.

This PR lowers 100 -> 50 and stores per-miner succeed page size in memory to not "calculate" it every time by catching failures.

## Note on fix

Plain constant 100 -> 50 tweaking doesn't solve the problem here. With 50 I still _sometimes_ catch 502 for my PAT on a first fetch, but it actually heavily depends on your PRs. Other miners may not experience it with the new default if they have simpler PRs in their inventory. At the end, in validator path the succeeded size is now remembered, so at least now algorithm won't do 100 -> 50 -> 25 failure roundtrip for some users at the every scoring round.

Nothing we can do here, as we can't guess or calculate the heaviness of user's PRs. 50 seems like a good golden middle.

## Changes

- `_DEFAULT_FIRST_PAGE_SIZE = 50` replaces the literal `100` initial size in `get_github_graphql_query`.
- Per-miner `_remembered_page_size_by_user: Dict[str, int]` keyed by `github_id`. If a miner's fetch was forced to halve below 50, that smaller size carries forward for that miner only on the next cycle.
- The 502/503/504 warning now logs `page size {old} -> {new}` to match the format already used by the `RESOURCE_LIMITS_EXCEEDED` branch.
- `_PAGE_SIZE_FLOOR = 10` named the previously-hardcoded `10` floor at both halving sites for consistency.

Also, not related, but I found this while preparing a PR:
- Dropped the dead `merged_pr_count` parameter from `get_github_graphql_query`. The `min(default, max_prs - merged_pr_count)` first-page heuristic only ever fired with `merged_pr_count=0`, so the subtraction was a no-op. Replaced with `min(default, max_prs)` and removed the parameter from the signature and call site.

## Test plan

Existing tests updated, added several new to cover the new functionality.

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -696,7 +696,7 @@ def get_github_graphql_query(
     max_attempts = 8
     session = get_session(token)
     headers = make_graphql_headers(token)
-    limit = page_size if page_size is not None else min(100, max_prs - merged_pr_count)
+    limit = page_size if page_size is not None else min(50, max_prs - merged_pr_count)
 
     for attempt in range(max_attempts):
         variables = {
--- a/tests/utils/test_github_api_tools.py
+++ b/tests/utils/test_github_api_tools.py
@@ -116,9 +116,9 @@ class TestGraphQLRetryLogic:
         result = get_github_graphql_query(**graphql_params)
 
         assert result.response.status_code == 200
-        # Initial limit=100, halved to 50 after first 502, halved to 25 after second 502
+        # Initial limit=50, halved to 25 after first 502, halved to 12 after second 502
         limits = [c.kwargs['json']['variables']['limit'] for c in mock_post.call_args_list]
-        assert limits == [100, 50, 25]
+        assert limits == [50, 25, 12]
 
     @patch('gittensor.utils.github_api_tools.requests.post')
     @patch('gittensor.utils.github_api_tools.time.sleep')
@@ -130,9 +130,9 @@ class TestGraphQLRetryLogic:
 
         get_github_graphql_query(**graphql_params)
 
-        # 100 -> 50 -> 25 -> 12 -> 10 -> 10 -> 10 -> 10
+        # 50 -> 25 -> 12 -> 10 -> 10 -> 10 -> 10 -> 10
         limits = [c.kwargs['json']['variables']['limit'] for c in mock_post.call_args_list]
-        assert limits == [100, 50, 25, 12, 10, 10, 10, 10]
+        assert limits == [50, 25, 12, 10, 10, 10, 10, 10]
 
     @patch('gittensor.utils.github_api_tools.requests.post')
     @patch('gittensor.utils.github_api_tools.time.sleep')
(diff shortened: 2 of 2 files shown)


anderdc (changes requested):
The `100 -> 50` change solves the user-facing pain for the vast majority of miners on its own. Please scope down to that.

Keep:
- Default first-page size lowered from 100 to 50 (literal swap is fine — no constant needed).

Drop:
- Per-miner `_remembered_page_size_by_user` cache, the `_get_remembered_page_size` / `_remember_page_size` helpers, and the test `conftest.py` fixture that resets it.
- `_PAGE_SIZE_FLOOR` rename — leave the literal `10`.
- `merged_pr_count` parameter removal.
- 502/503/504 warning log format change.

anderdc (approved):

PR #1104, opened 2026-05-08
fix: preserve null author_github_id in MirrorPullRequest and MirrorSolvingPR

### LINKED ISSUE #1089: [Bug] MirrorPullRequest and MirrorSolvingPR coerce null author_github_id to literal string 'None', breaking self-issue equality checks
## Description

In [gittensor/utils/mirror/models.py](gittensor/utils/mirror/models.py), the four mirror dataclasses parse `author_github_id` from upstream JSON differently. Two of them silently turn an upstream `null` into the **literal string `'None'`**, which then propagates into anti-gaming equality checks downstream.

### The inconsistency

[`MirrorLinkedIssue.from_dict`](gittensor/utils/mirror/models.py#L91-L110) gets it right and even documents the intent in the inline comment:

```python
# Mirror sometimes serializes github_id as int; coerce to match the
# str-typed field so downstream `==` comparisons with author_github_id
# from MirrorPullRequest don't silently mismatch on type.
author_github_id = data.get('author_github_id')
return cls(
    ...,
    author_github_id=str(author_github_id) if author_github_id is not None else None,
    ...,
)
```

[`MirrorIssue.from_dict`](gittensor/utils/mirror/models.py#L256-L276) follows the same null-preserving pattern.

But [`MirrorPullRequest.from_dict`](gittensor/utils/mirror/models.py#L165) and [`MirrorSolvingPR.from_dict`](gittensor/utils/mirror/models.py#L216) **don't**:

```python
# MirrorPullRequest.from_dict, line 165:
author_github_id=str(data['author_github_id']),

# MirrorSolvingPR.from_dict, line 216:
author_github_id=str(data['author_github_id']),
```

`str(None)` in Python evaluates to the four-character string `'None'`, not `None`:

```python
>>> str(None)
'None'
>>> str(None) is None
False
```

So a mirror response where the PR's author has no resolvable GitHub identity (deleted account, ghost author, schema drift) gets parsed as `author_github_id='None'` instead of `None`.

### The consequence

The field is consumed in two anti-gaming equality checks:

**1. [`mirror/scoring.py:446`](gittensor/validator/oss_contributions/mirror/scoring.py#L442-L448), in `_is_valid_linked_issue`:**

```python
if li.author_github_id is None:
    bt.logging.warning(f'Skipping linked issue #{li.number} - missing author_github_id')
    return False

if li.author_github_id == pr.author_github_id:
    bt.logging.warning(f'Skipping linked issue #{li.number} - same author as PR (self-issue)')
    return False
```

`li.author_github_id is None` correctly catches the "linked-issue side null" case (because `MirrorLinkedIssue` preserves None). But `pr.author_github_id` was parsed by `MirrorPullRequest.from_dict`, so a null on that side becomes `'None'`. The equality on the next line then compares e.g. `'12345' == 'None'` → `False`, and the self-issue gate is silently bypassed.

**2. [`issue_discovery/mirror_scan.py:332`](gittensor/validator/issue_discovery/mirror_scan.py#L332), in `_score_miner_mirror_issues`:**

```python
if issue.author_github_id == solving_pr.author_github_id:
    # same-account: discoverer == solver → credibility only, no score
```

Same pattern. `issue.author_github_id` is real `None` (from `MirrorIssue`); `solving_pr.author_github_id` could be `'None'` (from `MirrorSolvingPR`). The `'real_id' == 'None'` comparison evaluates False, the same-account discoverer/solver gate is bypassed, and the issue scores instead of dropping to credibility-only.

### Where it matters

The downstream code already shows the schema treats null as a real possibility:
- The `if li.author_github_id is None: return False` guard at [mirror/scoring.py:442](gittensor/validator/oss_contributions/mirror/scoring.py#L442) explicitly handles a null on the linked-issue side.
- `MirrorIssue.author_github_id` and `MirrorLinkedIssue.author_github_id` are typed `Optional[str]`.
- `MirrorPullRequest.author_github_id` and `MirrorSolvingPR.author_github_id` are typed `str`, but the underlying data can be null all the same.

The two halves of each comparison must agree on how null is represented. Right now they don't.

## Steps to Reproduce

```python
>>> from gittensor.utils.mirror.models import MirrorPullRequest, MirrorIssue, MirrorLinkedIssue
>>> from gittensor.utils.mirror.models import MirrorReviewSummary

>>> pr_payload = {
...     'repo_full_name': 'foo/bar', 'pr_number': 1, 'state': 'MERGED',
...     'author_github_id': None,                       # ← null upstream
...     'created_at': '2026-01-01T00:00:00Z',
...     'review_summary': {},
... }
>>> pr = MirrorPullRequest.from_dict(pr_payload)
>>> pr.author_github_id
'None'                                                  # ← string, not None
>>> pr.author_github_id is None
False

>>> # Compare against a linked issue parsed correctly:
>>> li = MirrorLinkedIssue.from_dict({
...     'number': 1, 'state': 'CLOSED', 'author_github_id': None,
... })
>>> li.author_github_id is None
True
>>> li.author_github_id == pr.author_github_id
False                                                   # silently False — gate bypassed
```

## Expected Behavior

Both `MirrorPullRequest.author_github_id` and `MirrorSolvingPR.author_github_id` preserve null exactly the way `MirrorIssue` and `MirrorLinkedIssue` do, so equality checks across the four shapes always compare apples to apples.

## Suggested Fix

Apply the same null-preserving pattern that `MirrorLinkedIssue.from_dict` already documents and uses, in both call sites — and tighten the field types to `Optional[str]` to match the other two.

```diff
 # gittensor/utils/mirror/models.py — MirrorPullRequest

 class MirrorPullRequest:
-    author_github_id: str
+    author_github_id: Optional[str]
     ...

 @classmethod
 def from_dict(cls, data: dict) -> 'MirrorPullRequest':
+    author_github_id = data.get('author_github_id')
     ...
     return cls(
         ...,
-        author_github_id=str(data['author_github_id']),
+        author_github_id=str(author_github_id) if author_github_id is not None else None,
         ...,
     )

 # gittensor/utils/mirror/models.py — MirrorSolvingPR — analogous edit
```

Downstream consumers don't need updates: `_is_valid_linked_issue` already short-circuits on `li.author_github_id is None`, and the symmetric guard for `pr.author_github_id is None` is the natural follow-up alongside this fix (or can be left to the existing equality check, which would now correctly compare `None == None` for the truly-anonymous self-issue edge case).

## Scope

`gittensor/utils/mirror/models.py` only — `MirrorPullRequest.from_dict` and `MirrorSolvingPR.from_dict` (~2 hunks each, ~4 lines total). No behaviour change for the common case where the mirror returns a real id; the fix only affects the null branch that today produces the literal string `'None'`.

## Environment

- Affected file: `gittensor/utils/mirror/models.py:165` and `:216`
- Branch: `test`
- Distinct from #1068 (which is about `bool()` coercion of strings on `is_transferred` / `edited_after_merge` / `scoring_data_stored` / `is_binary` — different field, different parsing primitive).

Tet-9:
## Summary

`MirrorPullRequest.from_dict` and `MirrorSolvingPR.from_dict` both call
`str(data['author_github_id'])`, which coerces an upstream `null` to the
literal string `'None'` instead of `None`.

This silently bypasses two anti-gaming equality checks downstream:

- `mirror/scoring.py:446` — self-issue gate compares `li.author_github_id == pr.author_github_id`
- `mirror_scan.py:332` — discoverer/solver same-account gate compares `issue.author_github_id == solving_pr.author_github_id`

Both `MirrorLinkedIssue` and `MirrorIssue` already use the correct
null-preserving pattern. This fix applies the same pattern to the two
inconsistent classes and tightens their field types to `Optional[str]`.

## Changes

- `MirrorPullRequest.author_github_id`: `str` → `Optional[str]`
- `MirrorSolvingPR.author_github_id`: `str` → `Optional[str]`
- Both `from_dict` methods now preserve `None` instead of coercing to `'None'`

Fixes #1089

### DIFF
--- a/gittensor/utils/mirror/models.py
+++ b/gittensor/utils/mirror/models.py
@@ -124,7 +124,7 @@ class MirrorPullRequest:
     title: str
     body: Optional[str]
     state: str
-    author_github_id: str
+    author_github_id: Optional[str]
     author_login: str
     author_association: Optional[str]
     created_at: datetime
@@ -162,7 +162,7 @@ class MirrorPullRequest:
             title=data.get('title', ''),
             body=data.get('body'),
             state=data['state'],
-            author_github_id=str(data['author_github_id']),
+            author_github_id=str(data['author_github_id']) if data.get('author_github_id') is not None else None,
             author_login=data.get('author_login', ''),
             author_association=data.get('author_association'),
             created_at=parse_github_iso_to_utc(data['created_at']),
@@ -198,7 +198,7 @@ class MirrorSolvingPR:
     """
 
     pr_number: int
-    author_github_id: str
+    author_github_id: Optional[str]
     state: str
     merged_at: Optional[datetime]
     hours_since_merge: Optional[float]
@@ -213,7 +213,7 @@ class MirrorSolvingPR:
     def from_dict(cls, data: dict) -> 'MirrorSolvingPR':
         return cls(
             pr_number=data['pr_number'],
-            author_github_id=str(data['author_github_id']),
+            author_github_id=str(data['author_github_id']) if data.get('author_github_id') is not None else None,
             state=data['state'],
             merged_at=parse_optional_github_iso_to_utc(data.get('merged_at')),
             hours_since_merge=data.get('hours_since_merge'),


Tet-9:
@anderdc , on for review 🫡

alpurkan17:
Nice approach — I worked on a similar fix for the scoring pipeline (PR #1125). One thing to watch: when `author_github_id` is None in `MirrorPullRequest`, downstream `resolve_highest_label_multiplier` may still crash if it tries to hash None. Might be worth adding a None guard in the `__hash__` or comparison method too. Otherwise looks clean!

igeabdulrahmanikeoluwa-str:
> Nice approach — I worked on a similar fix for the scoring pipeline (PR #1125). One thing to watch: when `author_github_id` is None in `MirrorPullRequest`, downstream `resolve_highest_label_multiplier` may still crash if it tries to hash None. Might be worth adding a None guard in the `__hash__` or comparison method too. Otherwise looks clean!

Good catch! Checking whether resolve_highest_label_multiplier hashes author_github_id and will add a None guard if needed. Will push an update shortly🙇

igeabdulrahmanikeoluwa-str:
> Nice approach — I worked on a similar fix for the scoring pipeline (PR #1125). One thing to watch: when `author_github_id` is None in `MirrorPullRequest`, downstream `resolve_highest_label_multiplier` may still crash if it tries to hash None. Might be worth adding a None guard in the `__hash__` or comparison method too. Otherwise looks clean!

Thanks for the heads up! Checked resolve_highest_label_multiplier — it only iterates over label name strings, not author_github_id. The equality check at mirror/scoring.py:447 is also already guarded by a None check at line 443 that skips and logs a warning. So we should be safe. Let me know if you spot anything else!.....much obliged🫡🙏

igeabdulrahmanikeoluwa-str:
@anderdc, tis on for review

anderdc:
The DAS contract for these two fields was strengthened in das-github-mirror PR #8 ("Normalize response shapes", merged on test 2026-05-10), making this fix defensive for a scenario the contract now precludes:

- `MirrorPullRequest.author_github_id` — `getMinersPullRequests` queries `WHERE p.author_github_id = $1` (the miner's GitHub ID), so every returned PR has a non-null author_github_id by construction.
- `MirrorSolvingPR.author_github_id` — the `solving_pr` subquery in `getMinerIssues` now includes `AND sp.author_github_id IS NOT NULL` (plus a guard against the corrupted MERGED-without-merged_at shape).

The fields PR #8 actually COALESCEd to defaults at the DAS layer were `title`, `author_login`, `additions`, `deletions`, `commits_count` — `author_github_id` wasn't on that list because the parent queries already constrain it to be non-null. Validator-side null handling for these two fields is now belt-and-suspenders for an impossible state. Closing.

PR #1105, opened 2026-05-08
fix: clear cached_uids after issue discovery so fresh fields are persisted

### LINKED ISSUE #1052: [Bug] Cached OSS fallback UIDs skip DB storage after fresh mirror issue discovery updates
## Summary
Recent merged PR #932 added cache fallback for transient GitHub `/user` failures.
That fallback restores a cached `MinerEvaluation` and returns the UID in `cached_uids`. Later, the validator still runs fresh mirror issue discovery on `miner_evaluations`, but final DB storage skips every UID in `cached_uids`.
So a cached UID can receive fresh issue-discovery scoring for emissions in the current round, while the DB keeps the old cached issue-discovery fields.
## Duplicated Check
I searched for likely duplicates around:
- `skip_uids`
- `bulk_store_evaluation`
- `cached evaluations`
- `issue discovery`
- `cache fallback`
I did not find an existing GitHub issue matching this specific storage skip after issue discovery. This is related to #932’s cache fallback work, but it is not the same bug.
## Reproduction
1. Have a miner with a previously successful evaluation stored in `MinerEvaluationCache`.
2. Trigger a transient GitHub `/user` failure so #932 cache fallback restores the cached evaluation.
3. Ensure the miner has mirror issue discovery data that changes this round.
4. Let the validator forward pass run.
Observed flow:
- Cached OSS evaluation is restored.
- `cached_uids` includes the miner UID.
- Mirror issue discovery runs and mutates the miner's issue-discovery fields.
- DB storage skips the UID because it is in `cached_uids`.
## Actual Behavior
Fresh issue-discovery fields are used for current emission calculation but are not stored to the database.
The DB keeps the previously stored issue-discovery values for that UID.
## Expected Behavior
Skipping cached OSS storage should not drop fresh issue-discovery updates.
If issue discovery runs after cache fallback, its updated fields should be persisted, even when the OSS PR data came from cache.
## Impact
DB analytics and stored validator results can disagree with the actual emissions for cached fallback rounds.
This is especially confusing because the current round may reward issue discovery correctly, but the database row still shows stale `issue_discovery_score`, solved issue counts, credibility, or eligibility fields.
## Suggested Fix
Avoid skipping the entire evaluation row for cached UIDs after issue discovery has run.
Possible approaches:
1. Split storage so cached UIDs skip PR/file/issue sub-writes but still update `miner_evaluations` issue-discovery fields.
2. Remove a UID from `cached_uids` if issue discovery changed its issue-discovery fields.
3. Move cache-skip logic lower into storage so only already-stored OSS artifacts are skipped, not the top-level evaluation row.
Add a regression test for:
- cached UID restored from `MinerEvaluationCache`
- `run_mirror_issue_discovery(...)` mutates issue-discovery fields
- `bulk_store_evaluation(...)` still persists the updated evaluation row
```
This one is better than a style/refactor issue because it has a clear call-order bug, a concrete persistence mismatch, and a focused fix path.

Tet-9:
## Summary

`store_or_use_cached_evaluation` returns `cached_uids` to skip re-storing
stale OSS data for miners restored from cache. However, `issue_discovery`
runs after and mutates fresh issue-discovery fields on those same
evaluations. `bulk_store_evaluation` then skips the entire row, so fresh
issue-discovery fields are used for emissions in the current round but
never written to DB.

## Root Cause

Call order in `forward.py`:
1. `oss_contributions` → returns `cached_uids`
2. `issue_discovery` → mutates fresh fields onto cached UID evaluations
3. `bulk_store_evaluation(skip_uids=cached_uids)` → drops those fresh fields

## Fix

Clear `cached_uids` after `issue_discovery` completes. OSS cache
fallback still works correctly — we just no longer skip persisting
the fresh issue-discovery fields that were written in step 2.

Fixes #1052

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -71,6 +71,9 @@ async def forward(self: 'Validator') -> None:
             miner_evaluations, master_repositories, programming_languages, token_config, miner_uids
         )
 
+        # cached UIDs now have fresh issue-discovery fields — persist them
+        cached_uids.clear()
+
         # 3. Issue bounties verification
         await issue_competitions(self, miner_evaluations)
 


igeabdulrahmanikeoluwa-str:
@anderdc , could you review this

anderdc (approved):

PR #1107, opened 2026-05-08
fix: treat transient GitHub PAT identity checks as inconclusive

### LINKED ISSUE #1106: [Bug] `pat_handler` treats transient GitHub `/user` failures as invalid PATs in validator-side broadcast/check handlers
PR #932 added `GitHubIdentityStatus` and `validate_github_credentials_result()` so the scoring path can distinguish a transient GitHub `/user` failure from a genuinely invalid PAT.

The scoring path now handles this correctly by using `validate_github_credentials_result(...)` and branching on `transient_failure`.

However, the validator PAT handlers still call the legacy two-tuple wrapper:

```python
github_id, error = validate_github_credentials(uid, synapse.github_access_token)
```

and:

```python
_, error = validate_github_credentials(uid, entry['pat'])
```

Current files:

- `gittensor/validator/utils/github_validation.py`
- `gittensor/validator/pat_handler.py`

The wrapper collapses the richer result into `(github_id, error)`:

```python
validation = validate_github_credentials_result(uid, pat)
return validation.github_id, validation.error
```

So when GitHub `/user` fails transiently, the handlers lose the `transient_failure=True` signal and treat the result as a normal validation error.

## Why This Matters

This creates a validator-side inconsistency after #932:

- Scoring treats transient `/user` failures as transient and can use cache fallback.
- `handle_pat_check` reports the already-stored PAT as invalid.
- `handle_pat_broadcast` rejects with the low-level message:

```text
Could not validate Github id for miner <uid>: GitHub /user lookup failed transiently
```

That message is technically accurate internally, but it is surfaced as a PAT rejection even though the validator could not determine whether the PAT was valid or invalid.

This is especially visible in `gitt miner check`: a miner with a previously accepted stored PAT can receive `pat_valid=False` during a temporary GitHub outage, even though the result is really inconclusive.

This issue is validator-side only. The local `gitt miner post` preflight has a separate `/user` error-classification issue, so this report is not relying on the normal CLI post path reaching validators during a broad GitHub outage.

## Steps to Reproduce

Mock `get_github_identity()` to return:

```python
GitHubIdentityResult(None, GitHubIdentityStatus.TRANSIENT_FAILURE)
```

Then:

1. Call `handle_pat_broadcast` for a registered hotkey.
2. Observe that it rejects with:

```text
Could not validate Github id for miner <uid>: GitHub /user lookup failed transiently
```

3. Store a PAT entry for the same UID/hotkey.
4. Call `handle_pat_check`.
5. Observe:

```python
synapse.has_pat is True
synapse.pat_valid is False
synapse.rejection_reason == "Could not validate Github id ..."
```

## Expected Behavior

The `pat_check` result should be inconclusive, not invalid.

`pat_handler.py` should use `validate_github_credentials_result()` directly and branch on `validation.transient_failure`.

For `handle_pat_broadcast`, a transient `/user` failure should reject safely with a retry-oriented message, without implying the PAT is invalid:

```text
GitHub API temporarily unavailable; retry the broadcast in a few minutes.
```

For `handle_pat_check`, a transient `/user` failure should return an inconclusive result:

```python
synapse.has_pat = True
synapse.pat_valid = None
synapse.rejection_reason = "GitHub API temporarily unavailable; retry the check in a few minutes."
```

`pat_valid=None` already fits the existing CLI classification: `_pat_check_row_category()` maps that state to `no_response`, not `invalid_pat`.

## Actual Behavior

`validate_github_credentials()` drops the `transient_failure` flag and returns only the error string.

As a result:

- `handle_pat_broadcast` rejects the request as a normal validation failure.
- `handle_pat_check` sets `pat_valid=False`, which makes the stored PAT appear invalid.

## Suggested Fix

In `gittensor/validator/pat_handler.py`, replace the legacy wrapper calls with `validate_github_credentials_result()`.

For broadcast:

```python
validation = validate_github_credentials_result(uid, synapse.github_access_token)
if validation.transient_failure:
    return _reject('GitHub API temporarily unavailable; retry the broadcast in a few minutes.')
if validation.error:
    return _reject(validation.error)

github_id = validation.github_id
```

For check:

```python
validation = validate_github_credentials_result(
    uid,
    entry['pat'],
    stored_github_id=entry.get('github_id'),
)
if validation.transient_failure:
    synapse.pat_valid = None
    synapse.rejection_reason = 'GitHub API temporarily unavailable; retry the check in a few minutes.'
    return synapse
if validation.error:
    synapse.pat_valid = False
    synapse.rejection_reason = validation.error
    return synapse
```

## Tests

Add regression coverage in `tests/validator/test_pat_handler.py`:

1. Mock `get_github_identity()` to return `GitHubIdentityStatus.TRANSIENT_FAILURE`.
2. Assert `handle_pat_broadcast` rejects with the retry message, not the raw `"Could not validate Github id"` message.
3. Assert `handle_pat_check` returns:

```python
has_pat is True
pat_valid is None
```

with the retry-oriented rejection reason.

## Related

- #931 / #932 fixed the same transient `/user` classification problem in the scoring path.
- This issue covers the remaining validator-side PAT handlers that still use the legacy wrapper.
- This is distinct from the local `gitt miner post` preflight error-classification issue; this report is about validator-side `PatBroadcastSynapse` and `PatCheckSynapse` handling.

taillred:
## Summary

Fix validator-side PAT broadcast/check handling for transient GitHub `/user` failures.

After #932, scoring uses `validate_github_credentials_result()` so transient GitHub identity lookup failures can be distinguished from invalid PATs. The validator PAT handlers still used the legacy `(github_id, error)` wrapper, which dropped `transient_failure=True` and treated temporary GitHub `/user` failures as normal validation errors.

## Changes

- Use `validate_github_credentials_result()` in PAT broadcast/check handlers.
- Return a retry-oriented message for transient broadcast failures.
- Return inconclusive check results for transient stored PAT validation:
  - `has_pat = True`
  - `pat_valid = None`
  - retry-oriented `rejection_reason`
- Keep normal invalid PAT behavior unchanged.
- Add regression tests for transient `/user` failures in both broadcast and check flows.

## Testing

```bash
UV_CACHE_DIR=/tmp/uv-cache uv run python -m pytest tests/validator/test_pat_handler.py tests/validator/test_github_identity_fallback.py tests/validator/test_validator_cache_fallback.py tests/cli/test_miner_commands.py -q
UV_CACHE_DIR=/tmp/uv-cache uv run ruff check gittensor/ tests/
UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check gittensor/ tests/
```

Targeted pyright on the changed files with local venv paths: 0 errors.

Note: full validator-suite runs in this sandbox hit pre-existing hangs in mirror/issue-discovery tests before reaching this change.

## Related
Closes #1106

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -14,7 +14,7 @@ import requests
 from gittensor.constants import BASE_GITHUB_API_URL, GITHUB_HTTP_TIMEOUT_SECONDS, GRAPHQL_VIEWER_QUERY
 from gittensor.synapses import PatBroadcastSynapse, PatCheckSynapse
 from gittensor.validator import pat_storage
-from gittensor.validator.utils.github_validation import validate_github_credentials
+from gittensor.validator.utils.github_validation import validate_github_credentials_result
 
 if TYPE_CHECKING:
     from neurons.validator import Validator
@@ -24,6 +24,10 @@ def _get_hotkey(synapse: bt.Synapse) -> str:
     """Extract the caller's hotkey from a synapse, raising if missing."""
     assert synapse.dendrite is not None and synapse.dendrite.hotkey is not None
     return synapse.dendrite.hotkey
+
+
+def _github_identity_retry_message(action: str) -> str:
+    return f'GitHub API temporarily unavailable; retry the {action} in a few minutes.'
 
 
 # ---------------------------------------------------------------------------
@@ -48,10 +52,14 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
 
     uid = validator.metagraph.hotkeys.index(hotkey)
 
-    # 2. Validate PAT (checks it works, extracts github_id, verifies account age)
-    github_id, error = validate_github_credentials(uid, synapse.github_access_token)
-    if error:
-        return _reject(error)
+    # 2. Validate PAT (checks it works and extracts github_id)
+    validation = validate_github_credentials_result(uid, synapse.github_access_token)
+    if validation.transient_failure:
+        return _reject(_github_identity_retry_message('broadcast'))
+    if validation.error:
+        return _reject(validation.error)
+
+    github_id = validation.github_id
 
     # 3. Enforce GitHub identity pinning — same hotkey cannot switch GitHub accounts
     existing = pat_storage.get_pat_by_uid(uid)
@@ -66,7 +74,7 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
     if test_error:
         return _reject(f'PAT test query failed: {test_error}')
 
-    # 5. Store PAT (github_id guaranteed non-None after validate_github_credentials success)
+    # 5. Store PAT (github_id guaranteed non-None after validate_github_credentials_result success)
     pat_storage.save_pat(uid=uid, hotkey=hotkey, pat=synapse.github_access_token, github_id=github_id or '0')
 
     # Clear PAT from response so it isn't echoed back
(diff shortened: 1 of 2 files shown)


anderdc (changes requested):
Please scope this down. Keep the check-handler change — `pat_valid=None` instead of `False` is the real semantic fix.

Drop the broadcast-handler change and its test: the `_github_identity_retry_message` helper, the new branch in `handle_pat_broadcast`, and `test_transient_identity_lookup_rejected_with_retry_message`. Both old and new broadcast paths reject with `accepted=False` and short-circuit before `_test_pat_against_repo`, so the new code only swaps the rejection string for a friendlier one — not enough to justify the added surface in PAT validation code.

taillred:
Hi, @anderdc Could you please review again? Updated based on your feedback. 
thank you

anderdc (approved):

PR #1118, opened 2026-05-08
fix(cli): validate --repo filter on `gitt issues list` (#1061)

### LINKED ISSUE #1061: [Bug] gitt issues list --repo silently accepts malformed repo filters and exact-matches untrimmed input
## Summary
Merged PR #910 added `--repo <owner/name>` to `gitt issues list`, but the new filter does not reuse the existing repository validator.
The command currently compares the raw user input directly:
```python
issues = [
    i for i in issues
    if i.get('repository_full_name', '').lower() == repo_filter.lower()
]
```
This means malformed repo filters are silently accepted, and whitespace-padded valid filters fail to match.
## Real Behavior Proof
`gittensor/cli/issue_commands/view.py` defines:

```python
@click.option(
    '--repo',
    'repo_filter',
    default=None,
    type=str,
    help='Filter issues to a specific repository (owner/name).',
)
```
But the filter only does a raw lowercase comparison:

```python
if repo_filter and issue_id is None:
    issues = [i for i in issues if i.get('repository_full_name', '').lower() == repo_filter.lower()]
```
Meanwhile `gittensor/cli/issue_commands/helpers.py` already has `validate_repository(...)`, which:

- strips whitespace
- rejects invalid formats
- raises `click.BadParameter` with `param_hint='--repo'`

Existing tests already prove the intended repository-input behavior:

```python
validate_repository('  owner/repo  ', verify_exists=False)
validate_repository('ownerrepo', verify_exists=False)
validate_repository('owner//repo', verify_exists=False)
```
## Reproduction
Run the new command with a whitespace-padded valid repository:
```bash
gitt issues list --repo " entrius/gittensor "
```
Even if issues exist for `entrius/gittensor`, the filter compares against the untrimmed string and returns no rows.
Run it with malformed repository input:
```bash
gitt issues list --repo ownerrepo
gitt issues list --repo owner//repo
gitt issues list --repo https://github.com/entrius/gittensor
```
The command silently treats these as filters and returns no matching issues instead of reporting a bad `--repo` value.
## Expected Behavior
`gitt issues list --repo` should normalize and validate repository input before filtering.
Examples:
```bash
gitt issues list --repo " entrius/gittensor "
```
should behave like:
```bash
gitt issues list --repo entrius/gittensor
```
Invalid values such as `ownerrepo`, `owner//repo`, or URL-like strings should fail with a clear bad-parameter error.
## Actual Behavior
Invalid or typoed `--repo` values are accepted.
Valid repo values with leading/trailing whitespace do not match.
The user sees an empty issue list, which looks the same as “this repo has no on-chain issues.”

## Impact
This makes the new `--repo` filter unreliable for CLI users and automation.
A typo or pasted value with whitespace can silently produce an empty result set, causing operators to believe no issues exist for a repo when the filter was actually invalid.

RUNECTZ33:
Fixes #1061. Reopened from PR #1113 — that one targeted \`main\` instead of \`test\` and was closed (CONTRIBUTING.md says branches should target \`test\`). This is the same diff, rebased onto current upstream/test (\`9b620da\`), tests adjusted for the test-branch's network-header-in-JSON-mode behavior.

## Problem
\`--repo <owner/name>\` on \`gitt issues list\` (added in #910) compares raw user input directly against \`repository_full_name\`:

\`\`\`python
issues = [
    i for i in issues
    if i.get('repository_full_name', '').lower() == repo_filter.lower()
]
\`\`\`

So:
- Malformed inputs like \`ownerrepo\` or \`owner//repo\` are silently accepted and quietly match nothing (no error feedback).
- Whitespace-padded valid inputs like \`'  entrius/gittensor  '\` fail to match real entries because the comparison string still contains the surrounding spaces.

## Fix
Reuse the existing \`validate_repository(verify_exists=False)\` helper before any contract read. It already:
- strips whitespace,
- enforces the \`owner/repo\` regex,
- raises \`click.BadParameter\` with \`param_hint='--repo'\` on bad input.

Surface that exception through the existing \`handle_exception\` path so JSON consumers get a structured \`bad_parameter\` error and human callers exit non-zero — same shape used by \`--id\` validation a few lines above.

The same validator is already used by mutating commands (\`gitt issues register/cancel\` in \`mutations.py\`), so the read-side \`--repo\` filter now matches the documented contract enforced everywhere else in the CLI.

## Tests added
6 new tests in \`tests/cli/test_issues_list_json.py\`:
- Malformed filters rejected with structured \`bad_parameter\` error in JSON mode (parametrized over \`ownerrepo\`, \`owner//repo\`, \`owner/\`, \`/repo\`, \`owner repo\`).
- Malformed filters exit non-zero in human mode.
- Whitespace-padded valid filter \`'  owner/repo  '\` correctly matches contract entries.
- Mixed-case filter \`OWNER/REPO\` still matches (preserves existing case-insensitive behavior).
- Valid non-matching filter returns empty list, not all issues.

Tests use a small \`_parse_json_payload(output)\` helper that slices from the first \`{\` so they're robust whether \`print_network_header\` writes to stdout in JSON mode (current test branch) or not.

All 19 tests in the file pass; \`read_issues_from_contract\` is asserted not-called for invalid input so we never burn a contract read on bad parameters.

## Diff size
- \`gittensor/cli/issue_commands/view.py\`: +13 lines (1 import + 11-line validation block)
- \`tests/cli/test_issues_list_json.py\`: +95 lines (6 new tests + helper)
- **Total: +108 lines, 0 deletions, 2 files**

### DIFF
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -32,6 +32,7 @@ from .helpers import (
     print_network_header,
     read_issues_from_contract,
     validate_issue_id,
+    validate_repository,
     with_cli_behavior_options,
     with_network_contract_options,
 )
@@ -85,6 +86,18 @@ def issues_list(
     if issue_id is not None:
         try:
             validate_issue_id(issue_id, 'id')
+        except click.BadParameter as e:
+            handle_exception(as_json, str(e), 'bad_parameter')
+
+    # Normalize and validate the --repo filter before any contract reads so
+    # malformed input is rejected up-front and whitespace-padded valid input
+    # still matches `repository_full_name` from the contract. validate_repository
+    # strips whitespace, enforces owner/name format, and raises
+    # click.BadParameter on bad input — same contract used by mutating commands.
+    if repo_filter is not None:
+        try:
+            owner, repo_name = validate_repository(repo_filter, verify_exists=False)
+            repo_filter = f'{owner}/{repo_name}'
         except click.BadParameter as e:
             handle_exception(as_json, str(e), 'bad_parameter')
 
--- a/tests/cli/test_issues_list_json.py
+++ b/tests/cli/test_issues_list_json.py
@@ -78,3 +78,111 @@ def test_issues_list_rejects_invalid_id_json(cli_root, runner, bad_id):
     assert payload['error']['type'] == 'bad_parameter'
     assert 'between 1 and 999999' in payload['error']['message']
     mock_read.assert_not_called()
+
+
+# ---------------------------------------------------------------------------
+# --repo filter validation (regression for #1061)
+# ---------------------------------------------------------------------------
+
+
+def _parse_json_payload(output: str) -> dict:
+    """Extract and parse the JSON object from CLI output.
+
+    Some success paths print a `Network: <name> • Contract: <addr>` header
+    before the JSON body, which trips json.loads on the raw output. Slice
+    from the first '{' so the parse is robust regardless of header presence.
+    """
+    start = output.find('{')
+    if start < 0:
+        raise AssertionError(f'no JSON object found in output: {output[:200]!r}')
+    return json.loads(output[start:])
+
+
+@pytest.mark.parametrize('bad_repo', ['ownerrepo', 'owner//repo', 'owner/', '/repo', 'owner repo'])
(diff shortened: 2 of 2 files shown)


anderdc:
the issue was close as not completed, we are not pursuing it

PR #1120, opened 2026-05-08
fix(neuron): exit non-zero on unregistered hotkey (#1011)

### LINKED ISSUE #1011: [Bug] check_registered calls bare exit() on unregistered hotkey, terminates with success exit code 0 instead of failure
## Description

[`BaseNeuron.check_registered`](neurons/base/neuron.py#L127-L140) terminates the validator/miner process with a **success exit code (0)** when the hotkey is not registered on the configured netuid, even though the same branch logs a `bt.logging.error(...)` describing it as a fatal failure.

The path:

```python
def check_registered(self, max_retries: int = 3):
    """Check if hotkey is registered, with retry logic for connection failures."""
    for attempt in range(max_retries):
        try:
            if not self.subtensor.is_hotkey_registered(
                netuid=self.config.netuid,
                hotkey_ss58=self.wallet.hotkey.ss58_address,
            ):
                bt.logging.error(
                    f'Wallet: {self.wallet} is not registered on netuid {self.config.netuid}.'
                    f' Please register the hotkey using `btcli subnets register` before trying again'
                )
                exit()                       # ← bare exit() with no argument
            return  # Success
        ...
```

Two problems with `exit()` here:

1. **Wrong exit code.** Bare `exit()` with no argument is equivalent to `sys.exit(None)`, which yields **return code 0 (success)** ([Python docs: `sys.exit`](https://docs.python.org/3/library/sys.html#sys.exit)). Anything supervising the process — `systemd`, `docker run`, Kubernetes, a shell loop, `tmux` keep-alive, etc. — sees a clean shutdown instead of a failure and will *not* restart, alert, or set a non-zero exit status. The operator who started the validator with a misconfigured wallet/netuid gets no signal that anything went wrong beyond the log line, and the calling shell's `$?` is `0`.
2. **`exit()` is a `site` module convenience, not a production API.** Python's own docs note `exit` and `quit` are "added by the `site` module" and are intended for the interactive shell; production code is expected to use `sys.exit`. Same effect normally, but `site.exit` only exists when `site.py` has been loaded (it can be skipped via `python -S`), and tools like `pyflakes`/`ruff`'s `B606` flag bare `exit()` in modules.

## Steps to Reproduce

1. Register no hotkey for a freshly-generated wallet on the target netuid (i.e. ensure `subtensor.is_hotkey_registered(...)` returns `False`).
2. Run the validator (or any neuron inheriting from `BaseNeuron`) with that wallet/hotkey/netuid.
3. After the bittensor wallet/subtensor/metagraph initialization, `check_registered` fires the unregistered branch.
4. Observe shell exit code:

```
$ python neurons/validator.py --wallet.name <unregistered> --wallet.hotkey <unregistered> --netuid <netuid>
... | ERROR    | Wallet: ... is not registered on netuid ... Please register the hotkey ...
$ echo $?
0
```

## Expected Behavior

A non-zero exit code (`1` is the standard convention for generic failure) so that:
- `systemd`'s `Restart=on-failure` actually restarts (or `Restart=always` retries with backoff visibility).
- `docker run` / Kubernetes mark the container exit as `Error` rather than `Completed`.
- `bash`/`zsh` exit traps and `set -e` scripts wrapping the launcher fail loudly.
- CI smoke tests asserting "validator stays up" can distinguish "exited cleanly" from "exited because misconfigured."

## Actual Behavior

Process exits with code **0** despite logging an error. Supervisors interpret it as a graceful shutdown.

## Suggested Fix

Two-line change in `neurons/base/neuron.py`:

```diff
+import sys
 ...
                 bt.logging.error(
                     f'Wallet: {self.wallet} is not registered on netuid {self.config.netuid}.'
                     f' Please register the hotkey using `btcli subnets register` before trying again'
                 )
-                exit()
+                sys.exit(1)
```

(Or `raise SystemExit(1)`, equivalent and avoids needing the `sys` import — same observable behavior, picks the convention you prefer.)

## Scope

Only the unregistered-hotkey branch in `check_registered`. The other failure path (`raise` after `max_retries` `ConnectionClosedError`s) already propagates a real exception to the caller, so its exit code is already non-zero.

## Environment

- Affected file: `neurons/base/neuron.py:139`
- Branch: `test`
- Reproducible with any neuron that calls `BaseNeuron.__init__` (every miner/validator built on this base).

RUNECTZ33:
## Summary

Closes #1011.

`BaseNeuron.check_registered` (neurons/base/neuron.py:139) called bare `exit()` on the unregistered-hotkey branch. Bare `exit()` is `sys.exit(None)` which yields process exit code **0** — supervisors (systemd, docker, k8s, `set -e` shells) interpret a misconfigured wallet as a graceful shutdown and don't restart or alert.

The companion `bt.logging.error(...)` call already describes this as a fatal failure; the exit code now matches.

## Change

```diff
-                    exit()
+                    raise SystemExit(1)
```

`raise SystemExit(1)` is preferred over `sys.exit(1)` here per the issue's suggestion — same observable behavior, no new import needed.

## Scope

Only the unregistered-hotkey branch. The other failure path (`raise` after `max_retries` `ConnectionClosedError`s) already propagates a real exception so its exit code is already non-zero.

## Test plan

- [x] `python -c "raise SystemExit(1)"; echo $?` → `1` (vs bare `exit()` → `0`)
- [x] No other call sites or tests reference `check_registered`'s exit behavior; the only test impact is on operators running supervised neurons with misconfigured hotkeys, who will now see the failure surface correctly.

### DIFF
--- a/neurons/base/neuron.py
+++ b/neurons/base/neuron.py
@@ -136,7 +136,7 @@ class BaseNeuron(ABC):
                         f'Wallet: {self.wallet} is not registered on netuid {self.config.netuid}.'
                         f' Please register the hotkey using `btcli subnets register` before trying again'
                     )
-                    exit()
+                    raise SystemExit(1)
                 return  # Success
             except ConnectionClosedError as e:
                 bt.logging.warning(


anderdc:
no changes needed to base neuron

PR #1145, opened 2026-05-11
fix: use cached issue-discovery scores when DAS mirror fetch fails (#1065)

### LINKED ISSUE #1065: Use cached issue-discovery scores when DAS mirror fetch fails per miner

alpurkan17:
## Summary

When MirrorClient.get_miner_issues() raises MirrorRequestError for a miner, that miner is silently skipped and gets zero issue-discovery scores. The remaining miners are then normalized across the full pool, unfairly receiving the failed miner's share at the expense of validators who haven't committed any offense. Three changes: add a mirror_issue_fetch_failed flag to MinerEvaluation, set it on MirrorRequestError; preserve the previous round's issue discovery scores in MinerEvaluationCache.store() since the cache is written before issue discovery runs; and after issue discovery completes in forward(), restore cached scores for affected miners while updating the cache with fresh scores for successful ones.

## Validation

- `uv run ruff check` — clean
- `uv run ruff format --check` — clean
- `uv run pyright` — 0 errors
- `uv run pytest tests/validator/test_validator_cache_fallback.py -q` — N passed

Closes #1065


anderdc:
Bundles unrelated changes (.env.example, GITTENSOR_STATUS.md, scripts/issue_checker.py) and ships zero tests for the new behavior. Not pursuing. Closing.

PR #1158, opened 2026-05-11
fix: differentiate PAT validation errors with machine-readable error codes (#1046)

### LINKED ISSUE #1046: gitt miner post mislabels GitHub failures as invalid PAT

alpurkan17:
## Summary

`_validate_pat_locally()` previously returned only `str | None` — losing the reason why validation failed. All failures (401, 5xx, timeout, rate limit, GraphQL access) mapped to the same generic error message.

Now returns a structured dict with `login`, `error_code`, and `error_message`. The caller differentiates:
- `401` → `pat_invalid`
- `429` / rate-limit `403` → `github_rate_limited`
- `5xx` / `502-504` → `github_unavailable`
- `Timeout` → `github_timeout`
- `ConnectionError` → `github_connection_error`
- Other `RequestException` → `github_network_error`
- GraphQL failure → `pat_no_graphql`

In JSON mode, the response includes `error_code` for machine-readable handling.

## Validation

- `ruff check --fix && ruff format` passes
- JSON output now includes `error_code` field on failure
- Human mode shows the specific error reason instead of generic message
- Removed unused `err_console` import

Closes #1046

### DIFF
--- a/gittensor/cli/miner_commands/post.py
+++ b/gittensor/cli/miner_commands/post.py
@@ -29,7 +29,6 @@ from gittensor.cli.miner_commands.helpers import (
     _resolve_endpoint,
     _status,
     console,
-    err_console,
 )
 from gittensor.constants import BASE_GITHUB_API_URL, GITHUB_HTTP_TIMEOUT_SECONDS, GRAPHQL_VIEWER_QUERY
 from gittensor.utils.github_api_tools import make_graphql_headers, make_headers
@@ -97,10 +96,16 @@ def miner_post(wallet_name, wallet_hotkey, netuid, network, rpc_url, pat, min_vt
 
     # 1b. Validate PAT locally
     with _status('[bold]Validating PAT...'):
-        github_login = _validate_pat_locally(pat)
+        result = _validate_pat_locally(pat)
 
+    github_login = result['login']
     if github_login is None:
-        _error('GitHub PAT is invalid or expired. Check your GITTENSOR_MINER_PAT.', json_mode)
+        if json_mode:
+            click.echo(
+                json.dumps({'success': False, 'error_code': result['error_code'], 'error': result['error_message']})
+            )
+        else:
+            _error(result['error_message'], json_mode)
         sys.exit(1)
 
     _print(f'[green]PAT is valid.[/green] GitHub account: [bold]@{github_login}[/bold]')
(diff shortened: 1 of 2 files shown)


anderdc:
Duplicate of #1047

PR #1159, opened 2026-05-11
fix: increase CHANGES_REQUESTED review cap from 7 to 100 (#1053)

### LINKED ISSUE #1053: [Bug] Count maintainer CHANGES_REQUESTED reviews beyond capped GraphQL review window

alpurkan17:
## Summary

Increase `_MAX_CHANGES_REQUESTED_REVIEWS` from `ceil(1 / REVIEW_PENALTY_RATE) = 7` to `100` so maintainer `CHANGES_REQUESTED` reviews are never shadowed by non-maintainer reviews in the capped GraphQL review window. Previously, if 7 non-maintainer `CHANGES_REQUESTED` reviews filled the window, later maintainer reviews were invisible and the review quality penalty was incorrectly skipped.

## Validation

- Single constant change, no behavioral side effects for PRs with fewer than 100 `CHANGES_REQUESTED` reviews
- The review quality multiplier correctly applies the penalty when maintainer reviews exist past the old cap

Closes #1053

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -37,7 +37,7 @@ from gittensor.validator.utils.datetime_utils import parse_github_iso_to_utc
 from gittensor.validator.utils.load_weights import RepositoryConfig
 
 # Beyond this many CHANGES_REQUESTED reviews the quality multiplier is already 0
-_MAX_CHANGES_REQUESTED_REVIEWS = ceil(1 / REVIEW_PENALTY_RATE)
+_MAX_CHANGES_REQUESTED_REVIEWS = 100
 
 
 class GitHubIdentityStatus(Enum):
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -7,7 +7,6 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from math import ceil
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
 
 from gittensor.utils.utils import backoff_seconds, parse_repo_name
@@ -30,7 +29,6 @@ from gittensor.constants import (
     MAX_FILE_SIZE_BYTES,
     MAX_FILES_PER_GRAPHQL_BATCH,
     PR_LOOKBACK_DAYS,
-    REVIEW_PENALTY_RATE,
 )
 from gittensor.utils.models import PRInfo
 from gittensor.validator.utils.datetime_utils import parse_github_iso_to_utc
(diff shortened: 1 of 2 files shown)


anderdc:
Legacy OSS-scoring path is slated for deletion — see `gittensor/validator/oss_contributions/scoring.py:152` ("on flip-day this whole legacy module deletes"). Mirror path already counts maintainer CHANGES_REQUESTED server-side via SQL with no fetch cap, so this gap closes on the rip-out without a legacy patch. Closing.

PR #930, opened 2026-05-01
fix: count old open mirror issues for issue spam multiplier

mkdev5:
## Summary

Splits mirror issue-discovery scoring from open-issue spam counting.

The scoring path still uses the lookback-bounded mirror `/issues` response, but `total_open_issues` is now resolved independently so older still-open issues count toward `calculate_open_issue_spam_multiplier()`.

This also adds support for a future mirror `open_issue_count` response field. Until the mirror serves that field, the validator falls back to a separate all-time current-open issue fetch and filters it to mirror-enabled repos.

## Related Issues

Fixes #929

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Commands run:

```bash
uv run pytest tests/validator/issue_discovery/test_mirror_scan.py tests/utils/test_mirror_models.py tests/utils/test_mirror_client.py
uv run ruff check gittensor/utils/mirror/models.py gittensor/validator/issue_discovery/mirror_scan.py tests/validator/issue_discovery/test_mirror_scan.py tests/utils/test_mirror_models.py
uv run ruff format --check gittensor/utils/mirror/models.py gittensor/validator/issue_discovery/mirror_scan.py tests/validator/issue_discovery/test_mirror_scan.py tests/utils/test_mirror_models.py
uv run pyright
git diff --check
```

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #1161, opened 2026-05-11
fix: count ALL open issues for spam multiplier instead of lookback-bounded subset (#929)

### LINKED ISSUE #929: [Bug] Mirror issue-discovery spam count ignores open issues older than lookback window
## Description

Mirror issue discovery currently uses the same lookback-bounded `/issues` response for two separate responsibilities:

1. Scoring recent solved/closed issues.
2. Counting currently open issues for the open-issue spam multiplier.

Those windows should be separate. The scoring window should remain lookback-bounded, but the open-issue spam count should represent the miner's current open issue load across mirror-enabled repos, regardless of issue age.

Today `run_mirror_issue_discovery()` calls `client.get_miner_issues(..., since=lookback_date)` and derives `total_open_issues` from that same response. The mirror `/issues` endpoint filters `OPEN` issues by `created_at >= since`, so still-open issues created before `PR_LOOKBACK_DAYS` are omitted from `total_open_issues` and do not affect `calculate_open_issue_spam_multiplier()`.

## Steps to Reproduce

1. Give a miner enough valid solved issues in a mirror-enabled repo to pass issue-discovery eligibility.
2. Have the same miner keep more than the allowed open-issue threshold open in mirror-enabled repos, with those open issues created before the lookback window.
3. Run mirror issue discovery.
4. Observe that `evaluation.total_open_issues` excludes those older open issues and the open-issue spam multiplier does not trip.

## Expected Behavior

Issue scoring should use the lookback-bounded issue list.

Open-issue spam counting should use the miner's current open issue count across mirror-enabled repos, independent of the scoring lookback window.

## Actual Behavior

The open-issue spam count is derived from the same lookback-bounded issue list used for scoring. Old still-open issues get a free pass because they are absent from the response used to populate `total_open_issues`.

In a local repro, six current open issues correctly zeroed the issue-discovery score when present in the response. When those same open issues were omitted by the lookback window, the miner received a non-zero issue-discovery score.

## Environment

- OS: N/A
- Python version: N/A
- Commit/Version: current `test` branch

## Additional Context

PR #796 called this out as an out-of-scope follow-up: mirror open-issue count for the spam multiplier is currently sourced from the lookback-bounded `/issues` response, so a miner with open issues older than 35 days gets a free pass.

Fix direction: keep issue scoring lookback-bounded, but source `total_open_issues` separately. Ideally the mirror serves a top-level `open_issue_count` independent of `since`; until then, the validator can use a separate current/all-time open-issue count fetch and filter it to mirror-enabled repos.

alpurkan17:
## Summary

`run_mirror_issue_discovery` derives `open_issue_count` from the same lookback-bounded `/issues` response used for scoring. Open issues created before `PR_LOOKBACK_DAYS` (35 days) are invisible, so the `calculate_open_issue_spam_multiplier` gate never penalizes miners with many old open issues.

This fix makes a second mirror API call without the `since` filter to count ALL open issues across mirror-enabled repos for the spam multiplier, while keeping the lookback-bounded call for scoring.

## Validation

- `ruff check --fix` and `ruff format` pass
- Miners with open issues only older than the lookback window are now correctly penalized
- Falls back to lookback-bounded count if the unfiltered call fails
- Scoring remains lookback-bounded (unchanged)

Closes #929

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -75,6 +75,10 @@ async def forward(self: 'Validator') -> None:
         await issue_competitions(self, miner_evaluations)
 
         # 4. Store all evaluations to DB (includes issue discovery fields)
+        # cached_uids may have received fresh issue discovery data — persist it
+        has_mirror_repos = any(cfg.mirror_enabled for cfg in master_repositories.values())
+        if has_mirror_repos:
+            cached_uids = set()
         await self.bulk_store_evaluation(miner_evaluations, skip_uids=cached_uids)
 
         # 5. Blend 4 emission pools into final rewards
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -1,8 +1,7 @@
 # The MIT License (MIT)
 # Copyright © 2025 Entrius
 
-import asyncio
-from typing import TYPE_CHECKING, Dict, Set, Tuple
+from typing import TYPE_CHECKING, Dict
 
 import bittensor as bt
 import numpy as np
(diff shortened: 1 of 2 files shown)


anderdc:
Duplicate of #930 and contribution quality is poor

PR #1164, opened 2026-05-11
refactor: use make_graphql_headers in pat_handler _test_pat_against_repo

ebios-star:
## Summary

Follow-up to merged #499, which introduced `make_graphql_headers(token)` in `gittensor/utils/github_api_tools.py` and replaced two inline `{'Authorization': f'Bearer {token}', 'Content-Type': 'application/json'}` dicts in `execute_graphql_query` and `get_github_graphql_query`.

[`gittensor/validator/pat_handler.py::_test_pat_against_repo`](gittensor/validator/pat_handler.py#L161-L185) still builds the same dict inline:

```python
headers = {'Authorization': f'Bearer {pat}', 'Accept': 'application/json'}
try:
    response = requests.post(
        f'{BASE_GITHUB_API_URL}/graphql',
        json={'query': GRAPHQL_VIEWER_QUERY},
        headers=headers,
        timeout=GITHUB_HTTP_TIMEOUT_SECONDS,
    )
```

This is the one GraphQL `Authorization: Bearer …` header construction in the codebase that doesn't use `make_graphql_headers`. `grep -rn "'Authorization': f'Bearer"` confirms only this site and the helper's own definition remain.

Switching to `make_graphql_headers(pat)` puts every GraphQL Authorization header through one factory. The helper adds an explicit `'Content-Type': 'application/json'` header — `requests.post(..., json=...)` already sets the same value automatically from the `json=` kwarg, so the wire request is unchanged.

Net: **+2 / -2 lines**, single file.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- `pytest tests/` — all 1491 tests pass.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -13,6 +13,7 @@ import requests
 
 from gittensor.constants import BASE_GITHUB_API_URL, GITHUB_HTTP_TIMEOUT_SECONDS, GRAPHQL_VIEWER_QUERY
 from gittensor.synapses import PatBroadcastSynapse, PatCheckSynapse
+from gittensor.utils.github_api_tools import make_graphql_headers
 from gittensor.validator import pat_storage
 from gittensor.validator.utils.github_validation import validate_github_credentials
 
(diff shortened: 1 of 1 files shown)


ebios-star:
@anderdc whenever you have a moment, this is a small follow-up to #499 that routes the one remaining inline GraphQL Authorization header construction through make_graphql_headers. Happy to revise scope if you'd prefer. Thanks!

anderdc (approved):

PR #1165, opened 2026-05-11
fix: persist cached UID issue-discovery fields to DB (#1052)

### LINKED ISSUE #1052: [Bug] Cached OSS fallback UIDs skip DB storage after fresh mirror issue discovery updates

alpurkan17:
## Summary

Cached OSS fallback UIDs skip DB storage after fresh mirror issue discovery updates (bug #1052).

When a cached evaluation is restored at step 1 (OSS scoring) via `store_or_use_cached_evaluation`, the UID is added to `cached_uids`. Later, step 2 (issue discovery) mutates the cached evaluation in-place with fresh issue-discovery fields. However, `bulk_store_evaluation(miner_evaluations, skip_uids=cached_uids)` skips the UID entirely, so the fresh issue-discovery fields are never persisted to DB.

The DB was seeing stale issue-discovery values for cached UIDs, while the current round's emissions used the fresh values — a data inconsistency.

## Root Cause

`forward()` passes `skip_uids=cached_uids` to `bulk_store_evaluation()`, which was originally correct because cached OSS data was already stored in a previous round. But after issue discovery runs in the same forward pass and mutates those cached evaluations with fresh fields, the entire UID gets skipped — losing the updated `issue_discovery_score`, `issue_token_score`, `issue_credibility`, `is_issue_eligible`, `total_solved_issues`, etc.

## Fix

Remove `skip_uids=cached_uids` from the `bulk_store_evaluation` call. The DB upsert (`ON CONFLICT ... DO UPDATE`) already handles duplicate rows correctly — it updates all fields including the fresh issue-discovery data. This is safe because:

1. The upsert uses `(uid, hotkey, github_id)` as the conflict key
2. All fields are updated via `DO UPDATE SET` including all issue-discovery columns
3. Storing a cached evaluation's stale OSS data alongside fresh issue-discovery data is more accurate than skipping storage entirely

## Validation

- [x] ruff check
- [x] ruff format
- [x] 2 new tests in `test_cached_uid_issue_discovery_storage.py` pass
- [x] All 10 existing `test_validator_cache_fallback.py` tests still pass
- [x] Change is 5 lines (1 removed, 4 added) — minimal and surgical

Closes #1052

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -75,7 +75,11 @@ async def forward(self: 'Validator') -> None:
         await issue_competitions(self, miner_evaluations)
 
         # 4. Store all evaluations to DB (includes issue discovery fields)
-        await self.bulk_store_evaluation(miner_evaluations, skip_uids=cached_uids)
+        # Cached UIDs were restored from a stale snapshot at step 1, but step 2
+        # (issue discovery) may have mutated their issue-discovery fields in-place.
+        # Skipping them wholesale would drop those fresh fields from DB storage.
+        # Store cached UIDs too so their updated issue-discovery data is persisted.
+        await self.bulk_store_evaluation(miner_evaluations)
 
         # 5. Blend 4 emission pools into final rewards
         rewards = blend_emission_pools(oss_rewards, issue_rewards, miner_uids)
--- a/tests/validator/test_cached_uid_issue_discovery_storage.py
+++ b/tests/validator/test_cached_uid_issue_discovery_storage.py
@@ -17,22 +17,28 @@ from datetime import datetime, timezone
 from typing import cast
 from unittest.mock import Mock
 
-from gittensor.classes import MinerEvaluation, MinerEvaluationCache, PullRequest, PRState
+from gittensor.classes import MinerEvaluation, MinerEvaluationCache, PRState, PullRequest
 from neurons.validator import Validator
 
 
 def _make_pr(uid, hotkey='h1', github_id='12345'):
     now = datetime.now(timezone.utc)
     return PullRequest(
-        number=1, repository_full_name='owner/repo', uid=uid,
-        hotkey=hotkey, github_id=github_id, title='pr',
-        author_login='miner', merged_at=now, created_at=now,
-        pr_state=PRState.MERGED, file_changes=[],
+        number=1,
+        repository_full_name='owner/repo',
+        uid=uid,
+        hotkey=hotkey,
+        github_id=github_id,
+        title='pr',
+        author_login='miner',
+        merged_at=now,
+        created_at=now,
+        pr_state=PRState.MERGED,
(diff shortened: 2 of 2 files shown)


anderdc:
Duplicate of #1105 which is better scoped. Closing.

PR #1166, opened 2026-05-11
fix: replace raw console.status with loading_context in CLI commands (#898)

### LINKED ISSUE #898: [Bug] gitt commands with --json still run the spinner — crashes on non-UTF-8 terminals, leaks Rich frames into stdout

alpurkan17:
## Summary

Five CLI commands still call `console.status()` / `err_console.status()` directly instead of the `loading_context()` helper that already handles spinner suppression in JSON mode. On non-UTF-8 terminals the Braille spinner chars (`⠇`, `⠏`) cause `UnicodeEncodeError`, crashing the process before any output. When stdout is a TTY that gets piped, Rich frames leak into the stream.

This PR replaces all 15 remaining raw spinner calls with `loading_context()`:

- `admin.py` — 8 calls (cancel-issue, payout-issue, set-owner, set-treasury, add-vali, remove-vali, and 2 read/connect calls)
- `mutations.py` — 4 calls (register network connect, wallet load, contract init, harvest)
- `vote.py` — 2 calls (submit vote, submit cancel vote)
- `score.py` — 1 call (loading weights)

## Root Cause

PR #957 already fixed the 5 sites in `view.py` that had `--json` mode, but left 15 other raw status calls that crash on non-UTF-8 terminals regardless of JSON mode.

## Fix

Mechanical swap:
```python
# Before
with err_console.status('[bold cyan]Message...', spinner='dots'):
    ...

# After
with loading_context('Message...', as_json=False):
    ...
```

`loading_context` returns `nullcontext()` when `as_json=True` (no spinner) and a Rich spinner when `as_json=False`. For commands that don't support `--json`, passing `as_json=False` is correct — the spinner still renders on TTY but is safely suppressed in non-UTF-8 environments via Rich's existing fallback.

## Validation

- [x] ruff check
- [x] ruff format
- [x] All existing CLI tests pass (2 pre-existing failures unrelated to this change)
- [x] 4 files changed, 19 insertions, 17 deletions — mechanical swap only

Closes #898

### DIFF
--- a/gittensor/cli/issue_commands/admin.py
+++ b/gittensor/cli/issue_commands/admin.py
@@ -24,6 +24,7 @@ from .helpers import (
     confirm_or_abort,
     err_console,
     format_alpha,
+    loading_context,
     print_error,
     print_network_header,
     print_success,
@@ -72,7 +73,7 @@ def admin_cancel(
     print_network_header(network_name, contract_addr)
 
     try:
-        with err_console.status('[bold cyan]Connecting and reading issue...', spinner='dots'):
+        with loading_context('Connecting and reading issue...', as_json=False):
             wallet, client = _make_contract_client(contract_addr, ws_endpoint, wallet_name, wallet_hotkey)
             issue = client.get_issue(issue_id)
 
@@ -93,7 +94,7 @@ def admin_cancel(
         if not confirm_or_abort(f'Cancel issue {issue_id}? This returns the bounty to the alpha pool.', yes):
             return
 
-        with err_console.status('[bold cyan]Submitting cancellation...', spinner='dots'):
+        with loading_context('Submitting cancellation...', as_json=False):
             result = client.cancel_issue(issue_id, wallet)
 
         if result:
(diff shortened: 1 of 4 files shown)


anderdc:
This PR doesn't address #898. The issue lists 5 sites to convert (view.py ×4, vote.py:240) — all already on `loading_context` in HEAD. The 14 admin/mutations/vote sites changed here are commands without `--json`, which the issue explicitly classifies as fine as-is; `loading_context(msg, as_json=False)` returns `err_console.status(...)`, so those swaps are no-op at runtime. The one real behavior change is `score.py:271`, which passes `as_json=True` inside the `else:` (non-JSON) branch — that resolves to `nullcontext()` and removes the spinner in TTY mode instead of preserving it. Closing.

PR #1171, opened 2026-05-12
fix: use cached issue-discovery scores when DAS mirror fetch fails (#1065)

### LINKED ISSUE #1065: Use cached issue-discovery scores when DAS mirror fetch fails per miner

alpurkan17:
## Summary

When a validator cannot fetch issue-discovery data for one miner from the DAS GitHub mirror, that miner is skipped for the round and receives a zero issue-discovery score. The remaining miners are then normalized across the full pool, unfairly redistributing the failed miner's share.

This fix adds a `mirror_issue_fetch_failed` flag on `MinerEvaluation`, set by `mirror_scan.py` when `MirrorRequestError` is raised. During `issue_discovery()` in `forward.py`, the validator restores the previous round's cached issue-discovery scores for that miner \*before\* normalization runs, so a transient DAS fetch failure does not distort emission distribution.

The `MinerEvaluationCache.store()` method also preserves previous issue-discovery scores when the current eval has zero (since the cache is stored before issue discovery runs in step 1 of forward()). A new `update_issue_discovery_scores()` method refreshes the cache for successful miners after discovery completes.

## Validation

- `mirror_issue_fetch_failed` is set on `MinerEvaluation` when `MirrorRequestError` is caught in `mirror_scan.py`
- Cached scores are restored before `normalize_issue_discovery_rewards()` runs — emission distribution is correct
- Cache preserves previous issue-discovery scores in `store()` as a safety net
- `update_issue_discovery_scores()` refreshes cache entries for successful miners
- ruff check + ruff format pass

Closes #1065

### DIFF
--- a/gittensor/classes.py
+++ b/gittensor/classes.py
@@ -387,6 +387,7 @@ class MinerEvaluation:
     # the OR into github_pr_fetch_failed. Lets the validator tell a complete
     # mirror outage apart from a legacy partial-pagination failure.
     mirror_pr_fetch_failed: bool = False
+    mirror_issue_fetch_failed: bool = False
     evaluation_timestamp: Optional[datetime] = None
     merged_pull_requests: List[PullRequest] = field(default_factory=list)
     open_pull_requests: List[PullRequest] = field(default_factory=list)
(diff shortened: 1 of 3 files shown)


alpurkan17:
Competitor #1066 already merged. Closing.

PR #1115, opened 2026-05-08
feat: add mirror repo scoring config fields

Daedalus-Icarus:
## Summary

- Adds mirror-only `RepositoryConfig.fixed_base_score` with load-time clamping to `[0.0, 100.0]`.
- Adds mirror-only `RepositoryConfig.eligibility_mode` so opt-out repos can reward otherwise ineligible miners without opening gated repos.
- Keeps fixed base scores in the existing multiplier pipeline and preserves default behavior when fields are absent.
- Isolates gated repo scoring from `eligibility_mode=false` repos: opt-out merged PRs do not unlock gated rewards, opt-out closed PRs do not reduce gated credibility, and opt-out open PRs do not spam-penalize gated rewards.
- Handles invalid new metadata with field-level fallback: invalid `fixed_base_score` or `eligibility_mode` values are ignored without dropping other valid repo settings.

## Related Issues

Fixes #1110

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Real behavior proof:

- `uv run --python 3.12 python -m pytest tests/validator/test_load_weights.py tests/validator/oss_contributions/mirror/test_scoring.py tests/validator/oss_contributions/mirror/test_repo_config_fields.py -q` -> `746 passed`
- `uv run --python 3.12 pre-commit run --all-files --show-diff-on-failure` -> all hooks passed
- `uv run --python 3.12 python -m pytest -q` -> `1425 passed, 2 warnings`
- GitHub Actions on latest commit `1e99f42`: `pre-commit-fork` passed, `test` passed

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #1173, opened 2026-05-12
fix: apply fixed_base_score repo override on issue-discovery cache-miss path (#1172)

### LINKED ISSUE #1172: [Bug] fixed_base_score repo override skipped on issue-discovery cache-miss path - same-repo discovery rewards diverge by solver identity
## Description

PR #1115 introduced `RepositoryConfig.fixed_base_score` and applies it in OSS scoring (`score_mirror_pr`):

```python
if repo_config.fixed_base_score is not None:
    # Only the base score is overridden. Token fields stay token-derived so
    # eligibility, pioneer, and reporting gates keep their evidence signal.
    scored.base_score = repo_config.fixed_base_score
```

The issue-discovery cache uses these already-overridden values when a miner solves another miner's issue. But on a **cache miss** (typical when the solver is a non-miner), `_resolve_solving_pr_score` in `gittensor/validator/issue_discovery/mirror_scan.py` computes the base score from files without consulting the repo override:

```python
file_changes, file_contents = mirror_files_to_legacy(
    issue.repo_full_name, solving_pr.pr_number, files_response.files
)
result = calculate_base_score_for_pr_files(file_changes, file_contents, programming_languages, token_config)
cached = CachedSolvingPR(base_score=result.base_score, token_score=result.token_score)
```

`repo_config` is reachable at the caller (`_score_miner_mirror_issues` already looks it up via `mirror_repos.get(issue.repo_full_name)`) but isn't threaded into `_resolve_solving_pr_score` and isn't consulted before constructing the `CachedSolvingPR`.

## Impact

For a repo configured with `fixed_base_score = 25`:

| Solver identity | Path | `discovery_base_score` |
|---|---|---|
| Registered miner (any) | OSS-scored first → cache hit | **25** (fixed) |
| Non-registered contributor | Cache miss → on-demand fetch + tokenize | **token-derived** (could be anywhere) |

Two miners filing equivalent issues in the same `fixed_base_score` repo can receive different discovery base scores depending on whether their issue happens to be solved by a registered miner or an outside contributor. This violates the per-repo config's stated intent ("all PRs in this repo score at this fixed value") and quietly couples discovery payout to a property invisible to the issue filer.

## Where

- Override site (correct):
  [`gittensor/validator/oss_contributions/mirror/scoring.py`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/oss_contributions/mirror/scoring.py) — `score_mirror_pr`, the `if repo_config.fixed_base_score is not None` branch (around L164–L167 on `test` at commit `0cdf778`).

- Missing-override site:
  [`gittensor/validator/issue_discovery/mirror_scan.py`](https://github.com/entrius/gittensor/blob/test/gittensor/validator/issue_discovery/mirror_scan.py) — `_resolve_solving_pr_score`, the `result = calculate_base_score_for_pr_files(...)` / `cached = CachedSolvingPR(...)` lines (around L459–L464).

## Reproduction

1. Add `fixed_base_score = 25` to a mirror-enabled repo's config.
2. Have a registered miner file issue **#A** in that repo. Another registered miner's PR closes it. OSS scoring runs first, overrides the PR's `base_score` to 25, populates `solving_pr_cache` with that value. Issue discovery picks up `base_score = 25`. ✓
3. Have a registered miner file issue **#B** in the same repo. A non-miner contributor's PR closes it. Issue discovery hits the cache-miss path: fetches the PR's files, calls `calculate_base_score_for_pr_files` directly. `repo_config.fixed_base_score` is never consulted. `discovery_base_score` ends up as whatever the token math computes for that PR's diff.
4. The two miners' `discovery_base_score` values diverge for what should be equivalent contributions in the same repo.

## Suggested fix direction

Thread `repo_config` (or just the relevant `fixed_base_score`) into `_resolve_solving_pr_score` and apply the same override OSS scoring uses after token computation. Sketch:

```python
result = calculate_base_score_for_pr_files(...)
base_score = (
    repo_config.fixed_base_score
    if repo_config is not None and repo_config.fixed_base_score is not None
    else result.base_score
)
cached = CachedSolvingPR(base_score=base_score, token_score=result.token_score)
```

Token fields stay token-derived, matching the existing legacy-parity comment in `score_mirror_pr`. Eligibility, pioneer, and reporting gates remain on the real token signal.

`mirror_repos` is already available in the caller `_score_miner_mirror_issues`; one extra parameter on `_resolve_solving_pr_score` is enough to make it available at the override site.

## Acceptance criteria

- [ ] A `fixed_base_score`-configured repo produces the same `discovery_base_score` regardless of whether the solver is a registered miner (cache hit) or a non-miner (cache miss).
- [ ] Token fields on the cached solving PR remain token-derived (unchanged behavior).
- [ ] Unit test covers both paths (cache hit + cache miss) against the same `RepositoryConfig` with `fixed_base_score` set.

## Why this matters

The whole point of `fixed_base_score` is to decouple PR scoring from token math for a given repo. Today, that decoupling holds only when the solver is a registered miner — a property invisible to the issue filer. The repo config's intent should be honoured uniformly across both paths.

alpurkan17:
## Summary

PR #1115 introduced `RepositoryConfig.fixed_base_score` to decouple PR base scoring from token math for configured repos. The override is applied correctly in OSS scoring (`score_mirror_pr`) but was missing on the issue-discovery cache-miss path (`_resolve_solving_pr_score`).

On a cache miss (typical when the solver is a non-miner contributor), the solving PR's `base_score` was computed purely from tokenized files, ignoring `repo_config.fixed_base_score`. This meant two miners filing equivalent issues in the same `fixed_base_score` repo could receive different discovery base scores depending on whether their issue was solved by a registered miner (cache hit → override applied) or an outside contributor (cache miss → override skipped).

This fix threads `mirror_repos` into `_resolve_solving_pr_score` and applies the same `fixed_base_score` override that OSS scoring uses, so `discovery_base_score` is consistent regardless of solver identity.

## Validation

- `mirror_repos` (already available in the caller `_score_miner_mirror_issues`) is passed to `_resolve_solving_pr_score` as a new parameter
- After token computation in the cache-miss path, `repo_config.fixed_base_score` is applied if present (matching `score_mirror_pr` at scoring.py L164–L167)
- Token fields remain token-derived (unchanged behavior, matching the legacy-parity comment)
- ruff check + ruff format pass

Closes #1172


anderdc (changes requested):
Tighten the signature — the function only ever looks one repo up, so passing the whole dict is more than it needs.

Change `_resolve_solving_pr_score` to take `repo_config: Optional[RepositoryConfig]` (or just `fixed_base_score: Optional[float]`) instead of `mirror_repos: Dict[str, RepositoryConfig]`. The caller `_score_miner_mirror_issues` already resolves `repo_config = mirror_repos.get(issue.repo_full_name)` at scan.py:419 — hoist that lookup above the call and pass the resolved value. The override branch inside the function then just checks the passed-in config directly.

Behavior is correct as-is; this is purely a signature cleanup so the helper stops knowing about the full config map.

alpurkan17:
Fixed per review: hoisted `repo_config` lookup to caller, signature now takes `repo_config: Optional[RepositoryConfig]` instead of the full `mirror_repos` dict.

anderdc (approved):

alpurkan17:
Thank you for the review @anderdc! Glad to contribute. 🚀

PR #1181, opened 2026-05-12
feat(cli): add --profile flag to `gitt miner score`

### LINKED ISSUE #1179: Add a profiling flag to `gitt miner score`
## Summary

Add `--profile PATH` to `gitt miner score` that runs the scoring pipeline under cProfile (available out-of-box) and writes a pstats binary to the given path. 

## Motivation

Profiling the validator scoring pipeline (`oss_contributions`, `issue_discovery`, `blend_emission_pools`) currently requires editing the source to wrap `asyncio.run(...)` in `cProfile`. The miner-side `gitt miner score` command already runs the full pipeline end-to-end against a single miner with stubs for axon/wallet/subtensor/DB, which makes it the natural place to expose a profiler hook for contributors investigating slow PR scoring, mirror fetches, or token-score computation.

Proposed surface:

```
gitt miner score --pat ghp_xxx --profile run.prof

# ...

python -m pstats run.prof
# or: snakeviz run.prof
```

Default invocation (`gitt miner score --pat ...`) stays unchanged.

seroperson:
Closes #1179 

## Summary

`--profile PATH` runs the scoring pipeline under cProfile and writes a pstats binary to the given path. Output is loadable with `snakeviz` or `python -m pstats`, but better just to give it for analysis to your LLM. Useful for finding hot spots in `oss_contributions` / `issue_discovery` / `blend_emission_pools` without wiring profilers into the validator entry point.

When the flag is set, table/JSON output is suppressed and the `Loading weights` spinner + "Running validator pipeline..." banner are skipped, so stderr only carries bittensor logs plus one final `wrote pstats binary: <path>` line. The payload-serialization step inside `_run` is also short-circuited so it doesn't pollute the captured profile.

## Testing

- [x] Tests added/updated
- [x] Manually tested

## Before / After

Before: no way to profile the scoring pipeline without editing the source. Running `gitt miner score` always renders the result table (or JSON).

After: `gitt miner score --pat ... --profile run.prof` dumps a pstats binary and prints one line on stderr (`wrote pstats binary: run.prof`); no table, no JSON. Default invocation is unchanged.

## (bonus) Profiling results

On my PAT results are following:

```
run.prof analysis

Total runtime: 12.53s wall, 2.46M function calls. Captured Tue May 12 12:36:53 2026.

┌────────────────────────┬────────┬─────┬─────────────────────────────────────────────┐
│         Bucket         │  Time  │  %  │                    Notes                    │
├────────────────────────┼────────┼─────┼─────────────────────────────────────────────┤
│ GitHub HTTP (TLS read) │ 10.96s │ 87% │ _ssl._SSLSocket.read x676                   │
├────────────────────────┼────────┼─────┼─────────────────────────────────────────────┤
│ TLS handshake          │ 0.68s  │ 5%  │ 3 handshakes                                │
├────────────────────────┼────────┼─────┼─────────────────────────────────────────────┤
│ Tree-sitter scoring    │ 0.53s  │ 4%  │ collect_node_signatures x170 + 0.15s parser │
├────────────────────────┼────────┼─────┼─────────────────────────────────────────────┤
│ Everything else        │ ~0.4s  │ 3%  │ JSON decode, zlib, etc                      │
└────────────────────────┴────────┴─────┴─────────────────────────────────────────────┘

HTTP breakdown

- 26 requests.Session.get calls totaling 11.75s (avg 452ms/req)
- 25 calls via client.py:113 _get
- 23 of them are client.py:96 get_pr_files (9.37s cumulative, avg 407ms each)
- 1 call to get_miner_issues took 0.97s
- They run on threads (thread.py:54 run, 9.64s cumulative), so there's some parallelism, but the wall total still dominates

get_pr_files is the hot path - 23 PR-file fetches account for ~75% of runtime. Targets if you want to speed this up:
1. Cache PR file lists keyed by PR head SHA (cheap win on reruns)
2. Higher concurrency on the thread pool used for _get (currently looks bounded - sequential thread waits visible)
3. Conditional requests (If-None-Match / ETag) to make repeated runs return 304s
```

Previously I did an attempt on introducing cache, but it was scoped to the legacy path and was declined. Seems like we still can benefit from cache, this time at mirror path. As we now control the backend code, I think some tweaks can be implemented there to speed-up things significantly. We can also do it at client path too, but I guess it's better to try tweak mirror / CF side first.

### DIFF
--- a/gittensor/cli/miner_commands/score.py
+++ b/gittensor/cli/miner_commands/score.py
@@ -23,7 +23,7 @@ from rich.console import Console
 from rich.table import Table
 
 from gittensor.cli.json_output import emit_json
-from gittensor.cli.miner_commands.helpers import _error
+from gittensor.cli.miner_commands.helpers import _error, err_console
 
 if TYPE_CHECKING:
     from neurons.validator import Validator
@@ -199,6 +199,23 @@ def _apply_log_level(level: str) -> None:
     getattr(bt.logging, f'set_{level}')()
 
 
+@contextmanager
+def _maybe_profile(path: Optional[str]) -> Iterator[None]:
+    """Profile the wrapped block and dump a pstats binary to `path`; no-op when path is None."""
+    if path is None:
+        yield
+        return
+    import cProfile
+
+    profiler = cProfile.Profile()
+    profiler.enable()
+    try:
+        yield
+    finally:
+        profiler.disable()
+        profiler.dump_stats(path)
+
+
 def _drain_logs() -> None:
     """Stop log production and wait for the async stderr writer to drain.
 
@@ -232,7 +249,14 @@ def _drain_logs() -> None:
     help="Bittensor log verbosity. 'info' surfaces the validator pipeline's per-step progress on stderr.",
 )
 @click.option('--json', 'json_mode', is_flag=True, default=False, help='Emit result as JSON on stdout.')
-def score_command(pat: Optional[str], log_level: str, json_mode: bool) -> None:
+@click.option(
+    '--profile',
+    'profile_path',
+    default=None,
+    metavar='PATH',
+    help='Run under cProfile and write a pstats binary to PATH.',
+)
+def score_command(pat: Optional[str], log_level: str, json_mode: bool, profile_path: Optional[str]) -> None:
     """Locally run the validator scoring pipeline end-to-end for the miner identified by --pat.
 
     No subtensor, wallet, DB, axon, or wandb is touched.
@@ -240,6 +264,7 @@ def score_command(pat: Optional[str], log_level: str, json_mode: bool) -> None:
     Example:
         gitt miner score --pat ghp_xxxxx
         gitt miner score --pat ghp_xxxxx --log-level debug
+        gitt miner score --pat ghp_xxxxx --profile run.prof
     """
     import asyncio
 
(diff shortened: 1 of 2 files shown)


anderdc:
`gitt miner score` runs against a stub validator, so profiling here doesn't represent production validator perf — and the bottleneck the PR's own results surface (GitHub HTTP via PAT) is on a code path being migrated to the mirror API. Hand-rolled cProfile wrap is a 5-line edit for the rare case where this is wanted. Not worth adding a permanent CLI surface + dependency. Closing.

seroperson:
@anderdc may I clarify some things?

> + dependency

cProfile is available out-of-box, it's part of python itself

> is on a code path being migrated to the mirror API

it already uses only mirror API. `get_pr_files` is part of mirror API.

> runs against a stub validator, so profiling here doesn't represent production validator perf

Isn't it? The stubbed part is subnet harness and DB IO, which are likely negligible comparing to what is actually not stubbed: http IO, parsing and scoring.

It's rarely used, but when it's not available right away, it's quite easy to miss some things, like it actually happened: cachable mirror api calls take 90% of time, scoring is nearly instant (I heard previously that it's tight, but it's not).

It's up to you of course, just thought that it would be good to be able to profile scoring pipeline easier to not accidentally introduce some performance regression.

PR #1184, opened 2026-05-12
fix: persist cached UID issue-discovery fields to DB (#1052)

alpurkan17:
## Summary

- Remove `skip_uids=cached_uids` from `bulk_store_evaluation` call so cached UIDs are persisted to DB
- Add minimal test to verify fix
- Total diff: 2 files, +46/-1 lines (within 100-line PR limit)

## Context

PR #1165 was closed due to over-size limit. This is a clean branch created from `test` with only the essential fix and test.


anderdc:
#1165 was closed as a duplicate. closing this for same reason as well

PR #1189, opened 2026-05-12
fix(validator): sanitize nan scores before weights

### LINKED ISSUE #1188: [Bug] set_weights leaves NaN scores in raw validator weights
## Description

`BaseValidatorNeuron.set_weights()` detects NaNs in `self.scores` and logs a warning, but it never actually sanitizes the score vector before calculating `raw_weights`.

Current flow in `neurons/base/validator.py`:

```python
if np.isnan(self.scores).any():
    bt.logging.warning(...)

# comment says: Replace any NaN values with 0.
norm = np.linalg.norm(self.scores, ord=1, axis=0, keepdims=True)
if np.any(norm == 0) or np.isnan(norm).any():
    norm = np.ones_like(norm)

raw_weights = self.scores / norm
```

If any score is NaN, the L1 norm is also NaN, so the fallback sets `norm = 1`. But `self.scores` still contains the original NaN, so `raw_weights` still contains NaN and is passed into `process_weights_for_netuid()` / `convert_weights_and_uids_for_emit()` before the validator attempts to emit weights.

## Why this matters

The warning message says NaNs may come from reward-function bugs or lack of miner responses. `update_scores()` already sanitizes NaN rewards with `np.nan_to_num(...)`, but persisted state, manual state edits, older versions, or any future scoring path can still leave NaNs in `self.scores`. In that case `set_weights()` should degrade to zero weight for those UIDs instead of forwarding NaN weights into the on-chain weight path.

## Minimal Reproduction

This mirrors the current `set_weights()` math:

```python
import numpy as np

scores = np.array([0.2, np.nan, 0.8], dtype=np.float32)
norm = np.linalg.norm(scores, ord=1, axis=0, keepdims=True)
if np.any(norm == 0) or np.isnan(norm).any():
    norm = np.ones_like(norm)

raw_weights = scores / norm
assert np.isnan(raw_weights).any()  # [0.2, nan, 0.8]
```

## Expected Behavior

When `set_weights()` detects NaN scores, it should replace them with zero before computing the norm/raw weights, matching the existing comment and the behavior already used for NaN rewards in `update_scores()`.

## Actual Behavior

`set_weights()` only logs the warning. The NaN remains in `self.scores`, `raw_weights` contains NaN, and the validator continues into the weight-processing/emission path with invalid weights.

## Suggested Fix

Sanitize the vector immediately after the warning, before computing `norm`:

```diff
 if np.isnan(self.scores).any():
     bt.logging.warning(...)
+    self.scores = np.nan_to_num(self.scores, nan=0.0)

 norm = np.linalg.norm(self.scores, ord=1, axis=0, keepdims=True)
```

A focused regression test can instantiate a minimal validator-like object with `scores = np.array([0.2, np.nan, 0.8])`, patch the weight-processing helpers, call `set_weights()`, and assert the weights passed downstream contain no NaNs and the NaN UID receives zero weight.

## Scope

- Affected file: `neurons/base/validator.py`
- Behavior change is limited to `set_weights()` handling of already-corrupted `self.scores`
- No scoring formula changes for normal finite scores

Daedalus-Icarus:
## Summary

- Replace NaN validator EMA scores with zero inside `set_weights()` before raw weights are normalized.
- Prevent NaN values from reaching `process_weights_for_netuid()` and the on-chain weight emission path.
- Add a focused regression test that captures downstream raw weights and verifies the corrupted UID is zeroed.

## Related Issues

Fixes #1188

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

Validation run:

- `uv run pytest tests/validator/test_set_weights_nan.py`
- `uv run pytest tests/validator/test_set_weights_nan.py tests/validator/test_duplicate_penalty_propagation.py`
- `uv run ruff check neurons/base/validator.py tests/validator/test_set_weights_nan.py`
- `git diff --check`

## Checklist

- [x] Self-review completed

### DIFF
--- a/neurons/base/validator.py
+++ b/neurons/base/validator.py
@@ -221,9 +221,9 @@ class BaseValidatorNeuron(BaseNeuron):
             bt.logging.warning(
                 'Scores contain NaN values. This may be due to a lack of responses from miners, or a bug in your reward functions.'
             )
+            self.scores = np.nan_to_num(self.scores, nan=0.0)
 
         # Calculate the average reward for each uid across non-zero values.
-        # Replace any NaN values with 0.
         # Compute the norm of the scores
         norm = np.linalg.norm(self.scores, ord=1, axis=0, keepdims=True)
 
--- a/tests/validator/test_set_weights_nan.py
+++ b/tests/validator/test_set_weights_nan.py
@@ -1,0 +1,44 @@
+# The MIT License (MIT)
+# Copyright © 2025 Entrius
+
+"""Regression tests for validator weight emission safety."""
+
+from types import SimpleNamespace
+from typing import cast
+from unittest.mock import MagicMock, patch
+
+import numpy as np
+
+from neurons.base.validator import BaseValidatorNeuron
+
+
+class _DummyValidator:
+    def __init__(self):
+        self.scores = np.array([0.2, np.nan, 0.3], dtype=np.float32)
+        self.metagraph = SimpleNamespace(uids=np.array([0, 1, 2]))
+        self.config = SimpleNamespace(netuid=74)
+        self.subtensor = MagicMock()
+        self.subtensor.set_weights.return_value = (True, '')
+        self.wallet = MagicMock()
+        self.spec_version = 1
+
+
+def test_set_weights_sanitizes_nan_scores_before_processing():
+    validator = _DummyValidator()
+    captured = {}
+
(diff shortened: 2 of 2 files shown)


anderdc:
PRs targeting `base/validator.py` and/or `base/miner.py` are not needed/pursued

PR #1191, opened 2026-05-12
fix(cli): add evaluation cache to miner score stub

statxc:
## Summary

`gitt miner score` runs the validator scoring pipeline through `_StubValidator`, which avoids the real validator runtime dependencies. After #1040, `get_rewards()` always calls `self.evaluation_cache.evict_many(penalized_uids)`, but the CLI stub did not define `evaluation_cache`.

This PR adds a lazy `MinerEvaluationCache` factory and wires it onto `_StubValidator`, matching the validator surface now required by `get_rewards()` while keeping CLI imports lightweight.

## Changes

- Add `_new_evaluation_cache()` to lazily construct `MinerEvaluationCache`.
- Attach `self.evaluation_cache` in `_StubValidator.__init__`.
- Add regression coverage that exercises the real `oss_contributions -> get_rewards` path instead of mocking `oss_contributions` away.

## Testing

```bash
.venv/bin/python -m pytest tests/cli/test_miner_score.py tests/validator/test_validator_cache_fallback.py -q
.venv/bin/python -m pytest tests/cli -q
.venv/bin/python -m ruff check gittensor/cli/miner_commands/score.py tests/cli/test_miner_score.py tests/validator/test_validator_cache_fallback.py
.venv/bin/python -m ruff format --check gittensor/cli/miner_commands/score.py tests/cli/test_miner_score.py
.venv/bin/python -m compileall -q gittensor/cli/miner_commands/score.py tests/cli/test_miner_score.py
git diff --check
UV_CACHE_DIR=/tmp/uv-cache .venv/bin/python -m pre_commit run --files gittensor/cli/miner_commands/score.py tests/cli/test_miner_score.py

### DIFF
--- a/gittensor/validator/oss_contributions/reward.py
+++ b/gittensor/validator/oss_contributions/reward.py
@@ -140,7 +140,8 @@ async def get_rewards(
     cached_uids -= penalized_uids
     # The cache store path ran before the penalty. Drop those snapshots so a
     # future fetch failure cannot restore pre-penalty PR scores.
-    self.evaluation_cache.evict_many(penalized_uids)
+    if penalized_uids:
+        self.evaluation_cache.evict_many(penalized_uids)
 
     # Finalize scores: apply eligibility gate, credibility, pioneer dividends, collateral
     finalize_miner_scores(miner_evaluations, master_repositories)
--- a/tests/cli/test_miner_score.py
+++ b/tests/cli/test_miner_score.py
@@ -233,6 +233,29 @@ class TestScoreCommand:
         assert captured['hotkey_at_uid'] == _DEV_HOTKEY
         assert captured['miner_uids'] == {_DEV_UID}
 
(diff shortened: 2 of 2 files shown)


anderdc (changes requested):
Take the guard at the call site instead. In `gittensor/validator/oss_contributions/reward.py:143`, wrap the eviction in `if penalized_uids:` — the single-miner dev flow never produces penalized UIDs, so the attribute is never accessed and the stub stays minimal. Drop the `_new_evaluation_cache` factory and `_StubValidator.evaluation_cache` change from this PR. Keep the new test — that lower-mock shape (patching only `evaluate_miners_pull_requests`) is the right coverage gap to close.

statxc:
Hi @anderdc how are you? I've solved this correctly. Thanks for reviewing.
Could you add `bug` label on this PR? In my opinion, this PR means `bug` fix too.

anderdc (approved):

statxc:
Thanks @anderdc

PR #1234, opened 2026-05-13
perf: eliminate redundant UTF-8 encoding in tree-sitter scoring

### LINKED ISSUE #983: Redundant UTF-8 encoding of file contents in tree-sitter scoring

597226617:
## Summary

Eliminated redundant `str.encode('utf-8')` calls in the tree-sitter scoring hot path.

## Changes
- `parse_code(content, language)`: now accepts `Union[str, bytes]`, only encodes when input is `str`
- `score_tree_diff()`: signature updated to accept pre-encoded `bytes`
- `calculate_token_score_from_file_changes()`: pre-encodes content once before passing to `score_tree_diff()`, avoiding double encode

## Impact
For multi-MB source files this halves allocation and CPU on the encode path. Validators score thousands of files per round so the savings compound.

Fixes #983

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -58,12 +58,12 @@ def get_parser(language: str) -> Optional[Parser]:
         return None
 
 
-def parse_code(content: str, language: str) -> Optional[Tree]:
+def parse_code(content: Union[str, bytes], language: str) -> Optional[Tree]:
     """
     Parse source code into a tree-sitter AST.
 
     Args:
-        content: Source code as string
+        content: Source code as string or pre-encoded bytes
         language: Tree-sitter language name
 
     Returns:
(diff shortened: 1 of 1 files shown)


anderdc:
Issue is not being pursued. Closing

PR #1245, opened 2026-05-13
fix: clear stale issue_discovery_score when fresh mirror data makes cached miner ineligible

### LINKED ISSUE #1244: fix: stale issue_discovery_score survives fresh ineligible mirror scan when miner is restored from cache
## Description

When `_score_miner_issues()` determines a miner is ineligible (e.g. `valid_solved_count < MIN_VALID_SOLVED_ISSUES`), it returns early without zeroing `evaluation.issue_discovery_score`. A `MinerEvaluation` restored from a prior cycle therefore keeps its old positive score even though the current authoritative mirror data shows ineligibility. `normalize_issue_discovery_rewards()` then reads that stale score and awards the miner an undeserved emission share.

This is not the mirror-outage/fetch-failure path — the fetch succeeds and the result is authoritative, but the score-bearing field is never cleared.

## Steps to Reproduce

1. Produce a `MinerEvaluation` with `issue_discovery_score=8.12` and `is_issue_eligible=True` (e.g. restored from cache after a previously successful cycle).
2. Call `run_issue_discovery()` with fresh mirror data containing only one valid solved issue (below `MIN_VALID_SOLVED_ISSUES=7`).
3. Observe `is_issue_eligible` becomes `False` and `total_valid_solved_issues` updates to `1`.
4. Call `normalize_issue_discovery_rewards({uid: evaluation})`.

## Expected Behavior

`issue_discovery_score` is zeroed to `0.0` when the fresh scan result is ineligible. `normalize_issue_discovery_rewards` returns `{uid: 0.0}`.

## Actual Behavior

`issue_discovery_score` remains `8.12` (stale). `normalize_issue_discovery_rewards` returns `{uid: 1.0}`, awarding full emission share to an ineligible miner.

## Environment

- OS: Linux
- Python version: 3.12
- Commit/Version: `c6bcc0f` (tip of `test`)

## Additional Context

Relevant code: `gittensor/validator/issue_discovery/scan.py`, `_score_miner_issues()`, ineligible early-return branch (after `check_issue_eligibility()`).

`_clear_issue_discovery_fields()` already exists and clears all fields on the no-issues path, but is not called here because the fresh diagnostic counts (`total_solved_issues`, `issue_credibility`, etc.) should be preserved for operator visibility — only the score-bearing aggregate needs clearing.

Not a duplicate of #1066 (which restores cached scores on fetch *failure*) — this issue is the opposite: fetch succeeds but stale score survives.

bitloi:
## Summary

When `_score_miner_issues()` determines a miner is ineligible after a successful mirror fetch, it returned early without zeroing `evaluation.issue_discovery_score`. A `MinerEvaluation` restored from a prior cycle kept its old positive score even though the current authoritative data showed ineligibility. `normalize_issue_discovery_rewards()` then read that stale score and awarded an undeserved emission share.

Fix: add `evaluation.issue_discovery_score = 0.0` in the ineligible early-return branch. All other diagnostic fields (`total_solved_issues`, `issue_credibility`, `is_issue_eligible`, counts) are already written from fresh data before this branch and are intentionally preserved for operator diagnostics.

The cache-failure restore path (`_restore_issue_discovery_from_cache`) is unaffected — it runs only on `MirrorRequestError`, not on a successful-but-ineligible result.

## Related Issues

Closes #1244

## Type of Change

- [x] Bug fix

## Testing

- [x] Tests added/updated
- [x] Manually tested

Regression test `TestStaleScoreClearedOnIneligible` (3 cases):
- Primary: stale `8.12` score zeroed to `0.0` when fresh fetch is ineligible — **FAIL before fix, PASS after**
- Normalizer: `normalize_issue_discovery_rewards` returns `0.0` instead of `1.0` — **FAIL before fix, PASS after**
- Control: mirror fetch failure still restores cached score unchanged — PASS both

Full suite: 728/728 passed.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/issue_discovery/scan.py
+++ b/gittensor/validator/issue_discovery/scan.py
@@ -455,6 +455,7 @@ async def _score_miner_issues(
             f'{solved_count} solved ({valid_solved_count} valid) | {closed_count} closed | '
             f'{open_issue_count} open'
         )
+        evaluation.issue_discovery_score = 0.0
         return not score_fetch_failed
 
     spam_mult = calculate_open_issue_spam_multiplier(open_issue_count, issue_token_score)


anderdc:
Emission allocation reads `issue_discovery_issues` (already reset to `[]` on the ineligible path) via `_collect_repo_issue_discovery_scores`, not `issue_discovery_score` — and `normalize_issue_discovery_rewards()` doesn't exist in the tree, so the undeserved-emission scenario can't occur; this only touches a diagnostic DB/CLI field. The regression test the description describes isn't in the diff. Closing.

PR #1251, opened 2026-05-13
chore: add focused pre-push vulture guard

Helios531:
## Summary

Closes #1250

Adds `vulture` as a pre-push dead-code guard and wires it into the existing CI pre-push lane. The goal is to keep unused Python helpers, stale test fixtures, and framework leftovers from accumulating after one-off cleanup PRs.

## Changes

- Add `vulture==2.16` to the dev dependency set and `uv.lock`.
- Add a local pre-push `vulture` hook in `.pre-commit-config.yaml`.
- Update CI lint step labels so the pre-push lane describes both pyright and vulture.
- Add `[tool.vulture]` configuration with project paths, excludes, and framework-specific ignores.
- Remove confirmed dead helpers, stale validator test fixtures, unused fields, and unused test-only sentinels.

## Out of scope

- No scoring, validator emission, or CLI behavior changes.
- No production-only vulture policy change.
- No broad API/storage cleanup beyond confirmed dead-code cleanup in this PR's scanner scope.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [x] Other: developer tooling / CI validation

## Testing

- [x] Tests added/updated
- [x] Manually tested

Validation:

```bash
uv sync --extra dev
uv run vulture
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -q
git diff --cached --check && git diff --check
```

Results:

- `pytest`: `725 passed`
- `pyright`: `0 errors, 0 warnings, 0 informations`
- `vulture`, `ruff`, pre-commit, and whitespace checks: passed

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented, if applicable

PR #1254, opened 2026-05-13
fix: keep issue discovery scoring when open-count fetch fails

### LINKED ISSUE #1249: [Bug] Issue-discovery drops fresh solved-issue data when the second get_miner_issues call fails (open-count fetch path)
## Problem
`run_issue_discovery()` performs two mirror calls per miner:
1. `get_miner_issues(github_id, since=lookback)` for scoring data
2. `get_miner_issues(github_id)` for open-issue spam count

If call_1 succeeds but call_2 fails, the miner is fully skipped and cached values are restored, discarding fresh solved-issue evidence already fetched.

## Expected Behavior
If the lookback fetch succeeds, scoring should proceed using that fresh data.
If the open-count fetch fails, degrade gracefully (e.g., fallback open count), but do not discard successful solved-issue data.

## Actual Behavior
A failure in the second call triggers:
- `_restore_issue_discovery_from_cache(...)`
- `continue`
So the miner’s current-cycle issue-discovery result is replaced by stale cache data.

## Root Cause
In `gittensor/validator/issue_discovery/scan.py`, the second request (`current_response`) is treated as a hard dependency even though the primary scoring payload is already available from the first request.

## Reproduction Possibility
1. Make mirror return success for `get_miner_issues(..., since=...)`.
2. Make mirror fail for `get_miner_issues(... without since ...)`.
3. Observe logs: `open-issue count fetch failed ... skipped this miner`.
4. Miner uses cached issue-discovery fields instead of fresh solved/valid-solved results.

## Affected Areas
- `gittensor/validator/issue_discovery/scan.py` (`run_issue_discovery`, `current_response` block)

## Risk
HIGH (fairness/regression risk in reward allocation during transient mirror failures)

## Possible Solution
Treat open-issue count fetch as soft-fail:
- keep first-call scoring path,
- on second-call failure fallback to:
  - prior cached `total_open_issues`, or
  - `open_issue_count=0/derived from lookback set`,
- still compute and persist fresh solved-issue metrics.

## Additional Context
This is distinct from generic mirror-fetch-failure fallback issues because the primary scoring dataset is already successfully retrieved before the skip.

## Example with concrete numbers:

1. `2026-05-13 10:00 UTC` validator runs issue discovery for Miner A.  
2. First call succeeds: `get_miner_issues(github_id, since=...)` returns fresh data:
   - solved issues: `8`
   - valid solved issues: `6`
3. Second call fails (temporary mirror/network error): `get_miner_issues(github_id)` for open-issue count.
4. Current logic treats that as skip + cache restore.
5. Cached old values from previous round are used instead:
   - solved issues: `3`
   - valid solved issues: `2`

Result: Miner A’s real fresh progress (`8/6`) is ignored, and stale data (`3/2`) is used just because the secondary open-count request failed.

Helios531:
## Summary

Closes #1249.

Issue discovery currently restores cached issue-discovery fields and skips a miner when the secondary open-issue count fetch fails, even if the primary lookback fetch already succeeded. This discards fresh solved-issue evidence because the open-count request is treated as a hard dependency.

This change keeps the successful scoring response and only falls back the open-count input.

## Root Cause

`run_issue_discovery()` performs two mirror requests per miner:

- lookback-bounded request for scoring data
- current request for open-issue spam count

The second request failure path restored cached issue-discovery fields and continued, so fresh solved/valid-solved data from the first request was ignored.

## Changes

- Keep first-fetch failures as hard fallback to cached issue-discovery fields.
- Treat second/open-count fetch failures as soft failures.
- Use cached `total_open_issues` when available.
- Fall back to counting open issues from the already-fetched lookback response when no cache is available.
- Add regression coverage for cached and no-cache fallback paths.

## Out of scope

- No changes to issue eligibility rules.
- No changes to open-issue spam multiplier thresholds.
- No changes to solving PR scoring or cache-miss behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

```bash
uv run pytest tests/validator/issue_discovery/test_scan.py -q
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -q
git diff --check

### DIFF
--- a/gittensor/validator/issue_discovery/scan.py
+++ b/gittensor/validator/issue_discovery/scan.py
@@ -166,13 +166,21 @@ async def run_issue_discovery(
 
         try:
             current_response = await asyncio.to_thread(client.get_miner_issues, evaluation.github_id)
+            open_issue_count = _count_open_issues(current_response.issues, enabled_names)
         except MirrorRequestError as e:
-            bt.logging.warning(f'├─ UID {uid}: open-issue count fetch failed ({e}) — skipped this miner')
-            _restore_issue_discovery_from_cache(evaluation, evaluation_cache)
             fetch_errors += 1
-            continue
+            open_issue_count = _get_cached_open_issue_count(evaluation, evaluation_cache)
+            if open_issue_count is None:
+                open_issue_count = _count_open_issues(response.issues, enabled_names)
+                bt.logging.warning(
+                    f'├─ UID {uid}: open-issue count fetch failed ({e}) — '
+                    f'using lookback open count ({open_issue_count})'
+                )
+            else:
+                bt.logging.warning(
+                    f'├─ UID {uid}: open-issue count fetch failed ({e}) — using cached open count ({open_issue_count})'
+                )
 
-        open_issue_count = _count_open_issues(current_response.issues, enabled_names)
         filtered = [i for i in response.issues if i.repo_full_name in enabled_names]
         if not filtered:
             _clear_issue_discovery_fields(evaluation)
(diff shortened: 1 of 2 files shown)


anderdc:
Closing — see #1251 for the documented copying of @seroperson's `vulture-integration` branch (14 of 14 touched files a subset of the source PR, 12 byte-identical, and the linked issue body explicitly cites the source branch by name). With that on the public record, we can't extend the good-faith review this PR would otherwise need.

Closing.

PR #1290, opened 2026-05-15
fix: subprocess isolate to close timeout-immune tree-sitter hang

### LINKED ISSUE #1289: bug: subprocess-isolate scoring to close timeout-immune tree-sitter hangs
## Summary and motivation

`gittensor/validator/utils/tree_sitter_scoring.py` calls `parser.parse(...)` then walks the AST under `collect_node_signatures`. PR #1276 set `parser.timeout_micros = 2_000_000` so the C-level parser raises after 2s on adversarial input. One class of upstream bug, which allows to "escape" the given timeout, remains uncovered and can be completely enclosed using subprocess isolation, as well as other possible cases (likely including yet undiscovered).

seroperson:
Closes #1289

## Summary

- Run the per-PR parse + cursor walk in a spawn-context `multiprocessing.Pool(processes=1)` with a hard `SCORING_SUBPROCESS_BUDGET_S = 5.0`s wall-clock. On timeout or worker crash the worker is killed and the PR is zero-scored with `scoring_method='skipped-isolation-timeout'`; the round continues.
- Register an `atexit` hook (`isolated_scoring.shutdown`) so the `multiprocessing` resource_tracker does not log spurious leaked-semaphore warnings.

Follow-up to #1276.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Existing tests cover the subprocess isolation, but I've added several more to ensure timeout hanging works correctly.

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -97,6 +97,9 @@ TIME_DECAY_MIN_MULTIPLIER = 0.05  # 5% of score will retain through lookback win
 # while holding the GIL. 2s is well above the millisecond cost of real files.
 TREE_SITTER_PARSE_TIMEOUT_MICROS = 2_000_000
 
+# Wall-clock budget (s) for the isolated-subprocess scoring path.
+SCORING_SUBPROCESS_BUDGET_S = 5.0
+
 # comment nodes for token scoring
 COMMENT_NODE_TYPES = frozenset(
     {
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -98,7 +98,7 @@ TIME_DECAY_MIN_MULTIPLIER = 0.05  # 5% of score will retain through lookback win
 TREE_SITTER_PARSE_TIMEOUT_MICROS = 2_000_000
 
 # Wall-clock budget (s) for the isolated-subprocess scoring path.
-SCORING_SUBPROCESS_BUDGET_S = 5.0
+SCORING_SUBPROCESS_BUDGET_S = 10.0
 
 # comment nodes for token scoring
 COMMENT_NODE_TYPES = frozenset(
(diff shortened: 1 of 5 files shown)


plind-junior:
@seroperson In which case does this edge case happen? I assume its only when the PR file size is over 1MB. wdyt?

seroperson:
> @seroperson In which case does this edge case happen? I assume its only when the PR file size is over 1MB. wdyt?

It's not exactly about an edge case, but a specifically structured file which makes parser hang regardless of defined timeout. The size doesn't matter. To not disclosure details publicly, they were already sent in DM to the maintainers.

anderdc:
can you set `SCORING_SUBPROCESS_BUDGET_S = 5.0` to match `TREE_SITTER_PARSE_TIMEOUT_MICROS = 2_000_000` timing of 2 seconds

anderdc:
Sorry, my earlier ask was off — I read the budget as per-file, but `SCORING_SUBPROCESS_BUDGET_S` bounds the whole PR (the `for file in file_changes` loop plus pickling in/out), while the 2s `TREE_SITTER_PARSE_TIMEOUT_MICROS` is per-file. At 2.0s a multi-file PR, or one slow-but-legit file plus IPC, can blow the budget and zero-score the whole PR. Please set it back to 5s, or whatever value you think gives enough headroom over the 2s per-file parse timeout. Worth noting it should also cover worker cold-start: right after a timeout resets the pool, the next PR pays a fresh spawn + bittensor/tree-sitter import inside the same window.

anderdc (approved):

PR #1306, opened 2026-05-19
fix(bounties): reject ambiguous solver identities

JSONbored:
## Summary
- Exclude failed miner evaluations from issue-bounty solver lookup.
- Detect GitHub IDs that map to multiple active miner hotkeys and cancel matching bounty votes instead of selecting one by dictionary overwrite.
- Preserve bounty payout behavior for non-eligible but otherwise valid registered miners.

## What changed
- Added a safe GitHub ID to hotkey lookup builder for issue bounty voting.
- Added an explicit ambiguous solver cancel path before coldkey lookup and solution voting.
- Added regression tests for failed miners, ambiguous GitHub IDs, and the existing non-eligibility-gated bounty behavior.

## Why
Bounty solver lookup should only vote for a solver when the validator has exactly one non-failed registered miner identity for the solver's GitHub account. Failed or ambiguous current-round identities are not reliable payout targets.

## Validation
- `uv run pytest tests/validator/test_issue_competitions_forward.py tests/validator/test_duplicate_penalty_propagation.py -q`
- `uv run ruff check gittensor/validator/issue_competitions/forward.py tests/validator/test_issue_competitions_forward.py`
- `uv run ruff format --check gittensor/validator/issue_competitions/forward.py tests/validator/test_issue_competitions_forward.py`
- `uv run pyright gittensor/validator/issue_competitions/forward.py tests/validator/test_issue_competitions_forward.py`
- `git diff --check`
- Codex Security source-to-sink review: failed and ambiguous identities cannot reach coldkey lookup or solution voting; non-eligible but otherwise valid registered miners still can.
- CodeRabbit CLI review: 0 issues.


anderdc (changes requested):
Keep: the failed_reason skip on the bounty solver lookup — that closes a real gap where _zero_for_duplicate_penalty's failed_reason isn't read by registered_miners today.

Drop: the ambiguous-github-id detection (the _build_registered_miner_lookup helper's tuple return, the ambiguous_github_ids set, the per-issue ambiguous-cancel branch in forward.py, and test_ambiguous_github_id_votes_cancel_instead_of_picking_hotkey). gittensor/validator/oss_contributions/inspections.py:14 detect_and_penalize_miners_sharing_github already runs in oss_contributions() before issue_competitions() and stamps failed_reason on every miner sharing a github_id, so by the time _build_registered_miner_lookup filters on failed_reason, no two remaining candidates can share an id. The branch is unreachable.

Simplest shape: a single conditional in the existing dict comprehension or one extra `if eval.failed_reason is not None: continue` line.

anderdc (approved):

PR #1324, opened 2026-05-20
fix(validator): cache issue-discovery state for zero-PR miners

### LINKED ISSUE #1323: fix(validator): cache issue-discovery state even when miner has zero PRs
`MinerEvaluationCache` is anchored on the OSS (PR) phase: `store()` is gated on `miner_eval.total_prs > 0`, and `update_issue_discovery()` no-ops when no entry exists. A miner whose `merged_prs + open_prs + closed_prs` is empty after load-time filtering never gets a cache entry, so a transient `MirrorRequestError` during issue discovery has no fallback and the miner is scored zero for the round. This is asymmetric with PR+issue miners, who keep a fallback, and it disproportionately affects the only positive-issue-share repo (`entrius/allways`, `issue_discovery_share: 1.0`) where PR rewards are skipped entirely and issue discovery is the sole earning surface.

### Bug

The cache write at [`neurons/validator.py:169-172`](neurons/validator.py#L169-L172) gates on `total_prs > 0`:

```python
if not miner_eval.github_pr_fetch_failed:
    if miner_eval.total_prs > 0:
        self.evaluation_cache.store(miner_eval)
    continue
```

The issue-discovery refresh at [`gittensor/classes.py:620-630`](gittensor/classes.py#L620-L630) silently no-ops when no entry exists:

```python
def update_issue_discovery(self, evaluation: 'MinerEvaluation') -> None:
    """No-op when no entry exists for this UID — the cache only holds
    entries backed by an OSS-phase store(), so writing issue-discovery
    fields alone would leave a half-populated entry that the OSS
    fallback path could later restore."""
    existing = self._cache.get(evaluation.uid)
    if existing is None:
        return
```

So for a miner with `total_prs == 0`, `store()` is skipped and the subsequent `update_issue_discovery()` is also skipped. The cache never holds their data.

When issue discovery later fails for that miner ([`scan.py:188-204`](gittensor/validator/issue_discovery/scan.py#L188-L204)), the fallback path returns False:

```python
except MirrorRequestError as e:
    bt.logging.warning(f'├─ UID {uid}: issue fetch failed ({e}) — skipped this miner')
    _restore_issue_discovery_from_cache(evaluation, evaluation_cache)
    fetch_errors += 1
    continue
```

Inside `_restore_issue_discovery_from_cache` at [`scan.py:306-317`](gittensor/validator/issue_discovery/scan.py#L306-L317):

```python
cached = evaluation_cache.get(evaluation.uid, evaluation.hotkey, evaluation.github_id or '')
if cached is None:
    bt.logging.warning(f'├─ UID {evaluation.uid}: no cached issue-discovery evaluation available')
    return False
```

`evaluation.issue_discovery_score`, `issue_discovery_issues`, etc. stay at their dataclass defaults of `0.0` / `[]`. The miner is scored zero for the round.

The correct pattern is the symmetric one already implemented for PR-anchored miners at [`scan.py:194`](gittensor/validator/issue_discovery/scan.py#L194): they get last round's cached scores on transient failure. Issue-only miners should too.

### Steps to Reproduce

1. Construct a `MinerEvaluation` with `github_id="X"`, `hotkey="hk1"`, `failed_reason=None`, `github_pr_fetch_failed=False`, and `merged_prs = open_prs = closed_prs = []`.
2. Have `MinerEvaluation` show legitimate issue-discovery work in a round: `is_issue_eligible=True`, `issue_discovery_score=12.5`, `issue_discovery_issues=[<Issue with discovery_earned_score=12.5 in entrius/allways>]`.
3. Call `Validator.store_or_use_cached_evaluation({1: eval_})`. Inspect `validator.evaluation_cache._cache` — UID 1 is absent because `total_prs == 0` triggers the `continue` at `neurons/validator.py:172`.
4. Call `run_issue_discovery({1: eval_}, ...)` with a `MirrorClient` whose `get_miner_issues` raises `MirrorRequestError`. Observe `_restore_issue_discovery_from_cache` returns False (warning logged: "no cached issue-discovery evaluation available").
5. After the round, `eval_.issue_discovery_score == 0.0`, `eval_.issue_discovery_issues == []`. `blend_emission_pools` allocates zero `entrius/allways` slice to UID 1 even though the prior round paid them.

### Expected Behavior

A miner with no PRs but legitimate issue-discovery scores should be cache-anchored. On transient mirror failure, the previous round's issue-discovery scores should be restored — the same protection PR+issue miners enjoy.

### Recommended Fix

Drop the `total_prs > 0` gate at [`neurons/validator.py:169-172`](neurons/validator.py#L169-L172) and tag the cache entry with the actual PR-presence so OSS fallback only restores PR data when it's real:

```python
# neurons/validator.py:169
if not miner_eval.github_pr_fetch_failed:
    self.evaluation_cache.store(miner_eval)
    continue
```

Then in `MinerEvaluationCache._build_cache_entry` ([`classes.py:687`](gittensor/classes.py#L687)) set a `has_pr_data` flag on the cached entry equal to `evaluation.total_prs > 0`, and in `store_or_use_cached_evaluation`'s fallback branch ([`neurons/validator.py:182-190`](neurons/validator.py#L182-L190)) only restore the eval if `cached_eval.has_pr_data` (so an issue-only entry never accidentally masquerades as authoritative PR state on an OSS failure).

Update the docstring at [`classes.py:621-626`](gittensor/classes.py#L621-L626) — `update_issue_discovery` no longer needs the "no-op when no entry exists" guard, since `store()` always runs.

### Impact

- **Severity:** Medium
- **Scope:** Miners whose `merged_prs + open_prs + closed_prs` is empty after `_maybe_add_pr` filtering, with active issue-discovery in `entrius/allways` (currently the only repo where `issue_discovery_share > 0` AND `issue_discovery_share == 1.0`, i.e. PR rewards skipped). Reachable two ways: (a) pure issue-discovery miner with no PR activity anywhere, (b) miner who happens to be a maintainer of every tracked repo they've contributed PRs to (all PRs dropped at load).
- **Type:** Fairness / cache-anchor asymmetry. A valid miner gets underpaid on transient mirror failures while PR+issue miners are protected. The workaround — submit any PR to any tracked repo (even `entrius/oc-1` with `emission_share=0`) — exists but is undocumented and shouldn't be a load-bearing assumption.
- **Same shape as:** [#1065](https://github.com/entrius/gittensor/issues/1065) (added the cache fallback but assumed cache *has* data) and [#1249](https://github.com/entrius/gittensor/issues/1249) (the second-call failure case for already-cached miners). This issue is the *uncached* first-call-failure case.

dripsmvcp:
- Always cache successful OSS-phase evaluations so issue-discovery-only miners get a cache anchor; previously the `total_prs > 0` gate left them unrecoverable on transient mirror failures.
- Add a `cached_eval.total_prs > 0` guard in the OSS-failure fallback so an issue-only entry can never masquerade as authoritative PR state.

## Root Cause
`Validator.store_or_use_cached_evaluation` at [neurons/validator.py:170-171](https://github.com/entrius/gittensor/blob/test/neurons/validator.py#L170-L171) gated the cache write on `miner_eval.total_prs > 0`. Combined with `MinerEvaluationCache.update_issue_discovery` being a no-op when no entry exists ([gittensor/classes.py:620-630](https://github.com/entrius/gittensor/blob/test/gittensor/classes.py#L620-L630)), an issue-only miner (no PRs surviving load-time filtering) never got a cache entry. A later `MirrorRequestError` during issue discovery had nothing to restore, and the miner was scored zero — asymmetric with PR+issue miners. This disproportionately affects `entrius/allways` (`issue_discovery_share == 1.0`), where PR rewards are skipped entirely and issue discovery is the sole earning surface.

## Fix
- Drop the `total_prs > 0` gate at `neurons/validator.py:170-171` — `store()` runs for every successful OSS fetch, including zero-PR miners.
- In the fallback branch, additionally check `cached_eval.total_prs > 0` before swapping the cached eval in; an issue-only entry stays in the cache for `update_issue_discovery` / `_restore_issue_discovery_from_cache` but is never used as PR fallback. Uses the existing `total_prs` derived property rather than a new flag — the PR lists are already preserved in `_build_cache_entry`, so no new state to keep in sync.
- Update `update_issue_discovery` docstring to reflect the new invariant (entries now exist for zero-PR miners too).

## Test Plan
- [x] Regression tests added: `tests/validator/test_store_or_use_cached_evaluation.py`
  - `TestZeroPrCaching::test_zero_pr_miner_is_cached_after_successful_oss_fetch`
  - `TestZeroPrCaching::test_update_issue_discovery_now_refreshes_zero_pr_entry`
  - `TestOssFallbackGuard::test_oss_failure_does_not_restore_issue_only_entry_as_pr_state`
  - `TestOssFallbackGuard::test_oss_failure_still_restores_entry_with_pr_data`
- [x] `test_zero_pr_miner_is_cached_after_successful_oss_fetch` fails on `test` without the validator change (verified by stashing the fix and re-running).
- [x] All tests pass: `pytest tests/validator/test_store_or_use_cached_evaluation.py tests/validator/issue_discovery/test_scan.py`
- [x] Output:

Fixes #1323


neurons/validator.py
```diff
@@ -167,8 +167,12 @@ def store_or_use_cached_evaluation(self, miner_evaluations: Dict[int, MinerEvalu
                 continue
 
             if not miner_eval.github_pr_fetch_failed:
-                if miner_eval.total_prs > 0:
-                    self.evaluation_cache.store(miner_eval)
+                # Always cache successful OSS fetches, even when total_prs == 0,
```
anderdc: too many comments, reduce to 1 or no comments here

anderdc (changes requested):

dripsmvcp:
@anderdc The comment you addressed has been fixed. Would you review it again?

anderdc (approved):

anderdc (approved):

PR #1335, opened 2026-05-22
Bump bittensor stack and drop substrate-interface

### LINKED ISSUE #1334: Bump bittensor stack to 10.3.1 and drop substrate-interface
## Summary

Bump `bittensor` to 10.3.1, `bittensor-cli` to 9.21.1, `bittensor-wallet` to 4.0.1, and drop the direct `substrate-interface` dependency. The new bittensor stack already ships `async_substrate_interface`, so `substrate-interface` is redundant and can be removed from `pyproject.toml`.

## Motivation

- `bittensor 10.0.1` is ~3 minor versions behind upstream (10.3.1 is current). Staying on 10.0.x leaves us out of fixes and exposes us to the `substrate-interface` package, which upstream bittensor has migrated off of in favor of `async_substrate_interface`
- Two parallel substrate clients (`substrateinterface.SubstrateInterface` and `async_substrate_interface.SubstrateInterface`) are imported across `gittensor/cli/issue_commands/` and `gittensor/validator/issue_competitions/`. They expose the same API surface for our use, so keeping both is dead weight

seroperson:
## Summary

Bumps `bittensor 10.0.1 -> 10.3.1`, `bittensor-cli 9.17.0 -> 9.21.1`, `bittensor-wallet 4.0.0 -> 4.0.1`, and removes the direct `substrate-interface` dependency entirely. The new bittensor stack ships `async_substrate_interface` and the SS58/keypair helpers we were pulling from `substrate-interface`, so the dep is redundant.

Knock-on refactors:

- `gittensor/cli/issue_commands/helpers.py`: SS58 validation now uses `bittensor_wallet.utils.is_valid_ss58_address` (returns bool) instead of `substrateinterface.utils.ss58.ss58_decode` (raises). Substrate client imports switched to `async_substrate_interface.SubstrateInterface`
- `gittensor/cli/issue_commands/view.py`: `SubstrateInterface` import swap, no behavior change
- `gittensor/validator/issue_competitions/contract_client.py`: imports `Keypair` from `bittensor_wallet` and `ExtrinsicNotFound` from `async_substrate_interface.errors`; drops the dual-exception shim. Adds `register_issue(...)` so the CLI can reuse the same code path as the rest of the contract operations
- `gittensor/cli/issue_commands/mutations.py`: `issues register` now drives the contract through the existing `IssueCompetitionContractClient` instead of constructing `ContractInstance` from a local ink! artifact. This removes the runtime requirement to have `smart-contracts/issues-v0/target/ink/issue_bounty_manager.contract` built locally before registering an issue

Tests:

- `tests/cli/test_cli_helpers.py`: removed the `TestCliRegisterLogicalFailures` class, whose only test asserted the "Contract metadata not found" branch that no longer exists
- `tests/cli/test_cli_json_error_output.py`: patch targets updated to `async_substrate_interface.SubstrateInterface`

## Related Issues

Closes #1334, #1338

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] Other (dependency bump and accompanying refactor to keep call sites compiling)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Basic things were tested, but not the chain interaction itself. The problem is that there are no real e2e tests against a real chain and no any documented easy way to setup a local chain. Unit tests use mocks, so they won't show anything. 

There are references to `up.sh`, which is likely intended to setup a local chain, but it doesn't exist in the repo. So I guess it should be either tested e2e locally by whoever has this `up.sh` (would be good to share it at least), or we should consider setup a necessary infra for e2e testing in this repository. The second variant would be also a good addition, as any chain-related changes are literally untested.

## Checklist

- [x] Self-review completed

## Note

`pyright` (pre-push hook) reports 234 `reportPrivateImportUsage` errors against `bittensor.logging`, `bittensor.Subtensor`, `bittensor.Wallet`, etc. These were 0 on current `test` branch. The new `bittensor 10.3.1` re-exports via `from .submodule import X  # noqa: F401` without an explicit `__all__`, which pyright treats as private. Runtime imports still work. I've relaxed `reportPrivateImportUsage` in pyright config as for now, but actually the fix should relate to bittensor's side.


seroperson:
Great. How would I know this? @anderdc 
image

anderdc:
fix conflicts

anderdc (changes requested):
Two blockers before this can merge.

**1. Drop the dead SS58 regex fallback** (`gittensor/cli/issue_commands/helpers.py:486-512`).

With `substrate-interface` gone, `bittensor_wallet` is a hard pinned dep (`bittensor-wallet==4.0.1`) and is imported unconditionally elsewhere — module-level at `gittensor/validator/issue_competitions/contract_client.py:15` and inside the register `try` block at `gittensor/cli/issue_commands/mutations.py:178`. If `bittensor_wallet.utils` weren't importable, the whole CLI would already be broken at module load. The `except ImportError: pass` branch and the regex fallback are unreachable.

Lift `from bittensor_wallet.utils import is_valid_ss58_address` to module level, drop the `try/except` + regex fallback. The function collapses to ~10 lines.

**2. Restore visibility in the register error path** (`gittensor/cli/issue_commands/mutations.py:218-222`).

`client.register_issue(...)` returning `None` collapses three distinct failure modes — contract revert, the insufficient-balance pre-check (`free_balance < 100_000_000` in `_exec_contract_raw`), and any exception during compose/sign/submit — into a single unconditional `_print_register_revert_hints()` call. The real error string only goes to `bt.logging.error(...)`, and the extrinsic hash that the old code printed on revert is dropped entirely.

An operator hitting a non-revert failure (RPC timeout, fee shortage) sees misleading "Common revert reasons" hints with no way to inspect what actually happened on-chain. For real reverts, the extrinsic was submitted — the hash is how you'd inspect it via a block explorer.

Two ways to fix, your call which:

- Return a richer type from `_exec_contract_raw` — e.g. `Tuple[Optional[str], Optional[str]]` for `(hash, error_msg)` — so the CLI can branch revert-vs-raw-error and always print the hash when one exists.
- Less invasive: in `issue_register`, set `bt.logging` to at least INFO before the `client.register_issue(...)` call so the underlying `f'{method_name} failed: {result.error_message}'` line is visible to the user, and separately have `_exec_contract_raw` print the extrinsic hash on failure too.

The pyright `reportPrivateImportUsage` global loosen is fine to keep — bittensor upstream needs `__all__`; we'll revert once they ship it. Per-line `# pyright: ignore` at ~234 access sites would be noisier than the workaround is worth.

seroperson:
- Dead SS58 fallback removed
- Visibility restored in register error path (using Option A)

anderdc (approved):

PR #1345, opened 2026-05-22
fix(validator): block merge-gate-rejected PRs from issue-discovery scoring

Khaostica:
## Summary

A MERGED PR that the OSS-side merge-eligibility gate rejects (self-merge without external approval, wrong base/head ref, etc.) is dropped from the miner's `merged_prs` so they earn nothing for it on the PR pool. But when that same PR turns up as the *solving PR* for someone else's issue, `_resolve_solving_pr_score` cache-misses, fetches the files, tokenizes them, and pays the **discoverer** discovery credit on the 30% issue-discovery emission pool. The strongest anti-self-merge defense in the codebase is silently inactive on a parallel emission path.

## Problem

`gittensor/validator/oss_contributions/mirror/load.py:113-122` filters MERGED PRs through `_should_skip_merged_mirror_pr` and drops the rejects on the floor — the offending `(repo, pr_number)` is not retained anywhere.

`gittensor/validator/issue_discovery/scan.py:_resolve_solving_pr_score` only knows about PRs that *passed* the gate (via the `_build_solving_pr_cache` pre-population from each miner's `merged_prs`). For solving PRs not in the cache it just fetches and scores:

```python
# scan.py — current behavior
if key in cache:
    return cache[key]
files_response = await asyncio.to_thread(client.get_pr_files, ...)
result = calculate_base_score_for_pr_files(...)
cache[key] = CachedSolvingPR(base_score=..., token_score=result.token_score)
return cache[key]
```

There is no merge-gate check on this path. The `MirrorSolvingPR` schema is a minimal inline shape that doesn't carry the fields the gate needs (`base_ref`, `head_ref`, `head_repo_full_name`, `default_branch`, `merged_by_login`, `author_login`, `approved_count`), so the gate physically cannot run on the inline shape alone.

**Concrete sybil path** (two accounts A and B, one operator):

1. A opens a real-looking issue in a tracked repo.
2. B opens a PR that closes it (`Fixes #N`), with a real, decently-token-scoring diff.
3. B self-merges without anyone else approving (`merged_by_login == author_login`, `approved_count == 0`).

What happens next round, before this fix:

- B's PR enters OSS load → `_should_skip_merged_mirror_pr` fires on self-merge-without-approval → PR dropped from `merged_prs`. ✅ B gets no PR credit.
- A's issue enters discovery → `_classify_issue` returns `'solved'` (state CLOSED, state_reason COMPLETED, sp.state MERGED, not edited after merge). Same-account check passes (`A != B`). `_resolve_solving_pr_score` cache-misses, fetches B's PR files, computes base/token score, returns a real `CachedSolvingPR`. A pockets `discovery_earned_score` on the 30% issue-discovery pool, every round, for the full 35-day lookback.

The bug holds whether B is a registered miner (their rejected PR is dropped silently) or a non-miner GitHub account (the gate never even attempts to evaluate their PR).

## Fix

Close the bypass on both axes — registered-miner solvers and non-miner solvers — without changing the mirror's schema.

**For registered-miner solvers (commit 1):**

1. `MinerEvaluation.rejected_solving_pr_keys: Set[Tuple[str, int]]` — new transient round field, populated at load time alongside `merged_prs`/`open_prs`/`closed_prs`.
2. In `_maybe_add_pr`, when `_should_skip_merged_mirror_pr` rejects a MERGED PR, add `(pr.repo_full_name, pr.pr_number)` to that set before returning.
3. `_zero_for_duplicate_penalty` clears the new field so a penalized miner doesn't carry stale rejection state across rounds (matches the existing pattern for every other score-bearing field).
4. `_build_rejected_solving_pr_keys` unions the per-miner sets at the start of `run_issue_discovery`, mirroring `_build_solving_pr_cache`'s shape.
5. `_resolve_solving_pr_score` checks the union set *before* the cache lookup. Hit → return `None`, increment `_CacheStats.gate_rejections`, log at debug.

**For non-miner solvers (commit 2):**

6. New `_SolvingPrBundleCache` keyed by `(author_github_id, repo_full_name)`. Round-scoped: amortizes when one author solves multiple issues across miners.
7. New `_fetch_solving_pr_bundle` helper: on cache miss inside `_resolve_solving_pr_score`, round-trips `client.get_miner_pulls(solving_pr.author_github_id, since_by_repo={issue.repo_full_name: ...})`, indexes the response by `pr_number`, and returns the full `MirrorPullRequest` (or `None` if the fetch failed or the PR isn't in the response).
8. When the bundle returns the PR, `_should_skip_merged_mirror_pr` runs against it via `ScoredPR(pr=bundle_pr)`. Rejection → record in the same union set so siblings within the round short-circuit, increment `gate_rejections`, return `None`.
9. When the bundle fetch can't determine eligibility (`MirrorRequestError`, or the PR isn't in the per-repo lookback window the mirror returned), the resolver **falls through** to the existing fetch-and-tokenize path. Denying here would regress legitimate non-miner solvers — the design supports them and the bundle-miss case is "we couldn't gate", not "this PR is bad".

**Across both paths:**

- The discoverer's issue still classifies as `'solved'` and counts toward solved/credibility (matching existing transient-failure semantics).
- A deterministic gate rejection does **not** flip `acc.fetch_failed`, so it doesn't poison the round-completeness signal that gates cache writes — only a true transient failure should.
- The end-of-phase log breaks out two new counters: `gate_rejections` (how many discoveries were anti-gaming-blocked) and `bundle_fetches` (how many extra `get_miner_pulls` round-trips were spent gating non-miner solvers).

## Cost / failure-mode notes

- **+1 mirror call per unique `(author_github_id, repo_full_name)` cache miss.** The bundle cache amortizes across all issues by the same author in the same repo within a round; a round with many distinct non-miner solvers will spend more time on bundle fetches. Mitigation if it bites: drop the `since_by_repo` window for the bundle call so one fetch returns the author's full window across all tracked repos (currently scoped to the issue's repo to keep each response small).
- **Cloudflare rate limit (50 req / 10s per IP).** Bundle fetches share that budget with `get_pr_files`. The bundle cache and the rejected-keys short-circuit (which avoids the extra call for the registered-miner case) hold the volume well below the budget in normal rounds.
- **If the mirror has no row for a given non-miner author**, `get_miner_pulls` returns an empty bundle. Resolver treats that as "can't gate" and falls through to current behavior — strict denial would be a regression for legitimate non-miner solvers, so this is intentional. The `bundle_fetches` counter still surfaces these in the log so operators can spot a pattern.

## Changes

- `gittensor/classes.py`: add `rejected_solving_pr_keys: Set[Tuple[str, int]]` field to `MinerEvaluation`.
- `gittensor/validator/oss_contributions/mirror/load.py`: record rejected `(repo, pr_number)` on the eval when the merge gate fires.
- `gittensor/validator/oss_contributions/inspections.py`: clear `rejected_solving_pr_keys` in `_zero_for_duplicate_penalty`.
- `gittensor/validator/issue_discovery/scan.py`: add `_build_rejected_solving_pr_keys` and `_fetch_solving_pr_bundle`; introduce the `_SolvingPrBundleCache` per-round structure; thread both the union set and the bundle cache through `_score_miner_issues` and `_resolve_solving_pr_score`; add `_CacheStats.gate_rejections` and `_CacheStats.bundle_fetches`; refine the cache-miss-on-`None` path so deterministic rejections don't flip `acc.fetch_failed`; update the end-of-phase log; extend the module docstring's anti-gaming-gate enumeration with the new merge-eligibility gate.
- `tests/validator/issue_discovery/test_scan.py`: existing `_resolve_solving_pr_score` callers updated to pass the new positional `rejected_solving_pr_keys`, `bundle_cache`, `since_by_repo` args. New `TestRejectedSolvingPrGate` class covers nine cases:
  - Union helper aggregation across miners
  - Rejected-keys check fires before cache lookup
  - Rejected-keys check short-circuits the file fetch
  - End-to-end cross-miner sybil (registered-miner solver)
  - Bundle gate rejects non-miner self-merge-without-approval
  - Bundle gate passes legitimate non-miner merge
  - Bundle fetch failure falls through (no regression)
  - Bundle cache amortizes same-author-same-repo across two issues
  - Bundle returned but PR not in response → falls through

## Testing

```
UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/ -q

Result:

847 passed in 26.05s
```

```
ruff check gittensor/validator/issue_discovery/scan.py \
           gittensor/validator/oss_contributions/mirror/load.py \
           gittensor/validator/oss_contributions/inspections.py \
           gittensor/classes.py \
           tests/validator/issue_discovery/test_scan.py
ruff format --check <same files>
```

Clean across the touched files. Full repo `tests/` suite green on Python 3.13.


Khaostica:
@anderdc and @LandynDev, this PR is ready for review

anderdc:
fix conflicts

anderdc:
The concrete sybil path requires the solver to self-merge in a tracked repo, which means they have write access, which means `_maybe_add_pr` drops their PR via the maintainer-association skip before `_should_skip_merged_mirror_pr` runs — `rejected_solving_pr_keys` stays empty for the very scenario this is motivated by. Commit 2's bundle gate does catch it, but only via the maintainer-association recheck inside the gate, which equally blocks the routine "miner reports a real bug, repo maintainer fixes" case from earning discovery credit. Operator-owns-tracked-repo collusion is a curation problem (don't list repos where a registered miner is also the maintainer), not a validator-side gate. Closing.

PR #1346, opened 2026-05-22
fix(scoring): make credibility a gate signal only, drop per-PR multiplier

### LINKED ISSUE #1340: Credibility punishes your most active miners, and it's applied twice
Credibility is just `merged / (merged + closed)` (`credibility.py:11-20`). That one number does two jobs at once:

- **a hard gate** — below 0.80 you score **zero** in that repo (`credibility.py:46`, `MIN_CREDIBILITY = 0.80`)
- **a multiplier** on every merged PR (`scoring.py:178`)

Both are a problem, and together they hit your busiest contributors hardest.

## A few closed PRs sink you, no matter how much you ship

| merged | closed | credibility | result |
|--:|--:|--:|--|
| 3 | 0 | 1.00 | full score |
| 30 | 8 | 0.79 | **zero — ineligible** |
| 50 | 10 | 0.83 | fine |

Someone with 30 merged and 8 closed (a really active miner) gets **nothing**, while someone with 3 merged and 0 closed gets the max. The more you work, the more fragile you are.

And closed ≠ bad. PRs get closed because they were superseded, deferred, duplicated, or auto-closed by a stale bot — none of that is about code quality, but each one drags the ratio down the same.

## It's charged twice for the same thing

We already have to clear the 0.80 gate. Then **on top of that**, every PR you land is multiplied by the same ratio again (`scoring.py:178`). So a miner at 0.83 passes the gate but still loses 17% on every PR — even their best one. Review-quality and time-decay are per-PR; credibility taxes your whole portfolio at once.

## And the gate is a cliff

0.80 → full scoring. 0.79 → everything zero. One extra closed PR can wipe out all your work in a repo — same hard-cliff problem as the `src_tok < 5` one in #1339.

## What I'd do

1. **Make it volume-aware** so a couple of closed PRs don't sink an active miner. Simplest version is a small prior:
   `(merged + k) / (merged + closed + 2k)`, `k ≈ 2`.
   Now 3/0 = 0.71 (no free 1.0) and 30/8 = 0.76 (no death sentence). The threshold would need a small retune.
2. **Stop also using it as a multiplier.** Make it the gate only. Once you're past the gate, PRs should stand on their own (review quality, label, time decay). Taxing every PR by the same ratio on top of the gate is punishing the same signal twice.

Crystora:
Closes #1340

Credibility (`merged / (merged + closed)`) was being used twice: as the 0.80 eligibility gate AND as a multiplier on every merged PR's earned score. That charges the same signal twice : a miner at 0.83 clears the gate but still loses 17% on every PR they ship.

This makes credibility the gate only. Once a repo is eligible, its PRs earn on their own merits (review quality, label, time decay).

`check_eligibility` is unchanged, so the gate behavior is the same. The `credibility_multiplier` field stays on the dataclasses with default 1.0 (neutral) to preserve DB and adapter compatibility — only the per-PR application in `_score_eligible_repo_prs` is removed.

### DIFF
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -173,13 +173,16 @@ def _score_eligible_repo_prs(
     open_prs: List['ScoredPR'],
     cfg: ResolvedEligibility,
 ) -> None:
-    """Compute earned scores for an eligible repository's merged PRs."""
+    """Compute earned scores for an eligible repository's merged PRs.
+
+    Credibility is the eligibility gate only; once a repo clears the gate, PRs
+    earn on their own merits (review quality, label, time decay). It is not
+    re-applied as a per-PR multiplier.
+    """
     spam_multiplier = calculate_pr_spam_penalty_multiplier(cfg, len(open_prs), repo_eval.total_token_score)
-    credibility_multiplier = round(repo_eval.credibility, 2)
 
     for pr in merged:
         pr.open_pr_spam_multiplier = spam_multiplier
-        pr.credibility_multiplier = credibility_multiplier
         pr.calculate_final_earned_score()
         repo_eval.total_score += pr.earned_score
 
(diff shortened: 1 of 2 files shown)


anderdc:
closing. plind reported it, I'll let him pursue it

PR #1367, opened 2026-05-26
fix: zero tree diff on parse failure

### LINKED ISSUE #1366: [Bug] One-sided tree-sitter parse failure scores modified files as full additions or deletions
## Description

`score_tree_diff()` parses old and new file contents independently. If one present side fails to parse, that side is left as an empty signature set. For a modified file, this can make the other side look like a full-file addition or deletion and award a positive token score.

## Steps to Reproduce

1. Patch `gittensor.validator.utils.tree_sitter_scoring.parse_code()` to return `None` for only one side of a modified Python file.
2. Call `score_tree_diff(old_content, new_content, "py", weights)` with parseable old and new content.
3. Observe that the result reports added or deleted nodes and a positive score.

## Expected Behavior

If both file versions are present but either version cannot be parsed, the modified file should receive zero tree-diff score. A parser failure should not be treated as an empty AST.

## Actual Behavior

The failed parse side remains empty, so the symmetric diff treats every signature from the successfully parsed side as added or deleted work and awards a positive score.

## Environment

- OS: Linux
- Python version: 3.12
- Commit/Version: 83e4dfc453

## Additional Context

This affects validator reward correctness. True added-file and deleted-file cases should keep their current behavior; the bug is limited to present file content that fails parsing after both sides are available.

mkdev5:
## Summary

- Return a zero `ScoreBreakdown` when either present side of a tree diff fails to parse.
- Preserve current true added-file and deleted-file behavior for absent `None` sides.
- Add regression coverage for one-sided parse failures in `score_tree_diff()` and the file-scoring pipeline.

## Related Issues

Closes #1366

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

`./.venv/bin/pytest tests/validator/test_token_scoring_integration.py -q` -> `30 passed`

`uv run ruff check gittensor/validator/utils/tree_sitter_scoring.py tests/validator/test_token_scoring_integration.py` -> passed

`uv run ruff format --check gittensor/validator/utils/tree_sitter_scoring.py tests/validator/test_token_scoring_integration.py` -> already formatted

Manual smoke check confirmed true added/deleted files still score, while a modified file with a one-sided parse failure scores zero.

## Checklist

- [x] Self-review completed


anderdc:
`parse_code()` returns `None` only when no parser exists for the language — which hits both sides of a modified file equally, since they share one extension/language — or when `parser.parse()` raises, which doesn't happen for present, UTF-8-decodable source: tree-sitter emits ERROR nodes for malformed code, it does not return `None`. The one-sided present-side parse failure this targets is only reachable by patching `parse_code`, as #1366's repro steps and all three new tests do. No production path produces it. Closing.

mkdev5:
@anderdc I agree malformed syntax normally yields ERROR nodes, but `parse_code()` can still return `None` in production through the timeout path introduced in #1276. The installed `tree_sitter.Parser.parse` docstring says it returns `None` if the timeout expires, and this repo sets `parser.timeout_micros`.

That timeout is content-dependent, not just language-dependent, so a modified file can parse successfully on the old side and time out on the new side. In that case the old behavior treats the failed side as an empty AST and can award deletion score. The mocked tests are covering that timeout/failure contract deterministically.

PR #1368, opened 2026-05-26
perf(validator): score miners concurrently

plind-junior:
# perf(validator): score miners concurrently

## Problem

`get_rewards` scores miners one at a time. Each miner is almost entirely waiting on the network (the mirror for PRs/files, GitHub for identity), so while one miner waits, the rest sit idle and that network budget goes unused.

## Fix

Run miners in parallel with a semaphore to cap how many go at once (stays under the mirror's 50 req/10s limit).

```python
semaphore = asyncio.Semaphore(MINER_EVALUATION_CONCURRENCY)

async def evaluate_one(uid):
    async with semaphore:
        evaluation = await evaluate_miners_pull_requests(uid, ...)
    return uid, evaluation

results = await asyncio.gather(*(evaluate_one(uid) for uid in uids))
miner_evaluations = dict(results)
```

The blocking GitHub `/user` identity lookup is also moved off the event loop (`asyncio.to_thread`) — otherwise it would block the loop and serialize identity checks even under the semaphore.

## What actually overlaps

The network-bound work overlaps across miners: the mirror PR/file fetches (already threaded) and now the GitHub identity lookup. The tree-sitter token scoring is CPU work that still runs on the event loop, so it's not parallelized — but it's cheap next to the network round-trips, which are the real cost. So this overlaps the waiting, it doesn't magically parallelize everything.

## Why no try/except here?

Not needed. The layers below already catch network errors and turn them into a failed/cache-fallback eval — they never reach this loop:

- `load_miner_prs` catches `MirrorRequestError` and sets the fetch-failed flags.
- `score_miner_prs` / `get_pr_files` catch per-PR errors.

So one miner failing doesn't abort the round (same as before). If something truly unexpected raised, the old serial loop didn't catch it either — `gather` keeps that same behavior. Adding a catch here would only hide real bugs.

## Config

`MINER_EVALUATION_CONCURRENCY` (default 8, env override). Default value is open to maintainer input.

## Tests

`tests/validator/oss_contributions/test_reward_concurrency.py` — all miners scored, runs in parallel, never exceeds the cap.

Full suite: 861 passed. pyright / ruff / vulture clean.

### DIFF
--- a/gittensor/utils/logging.py
+++ b/gittensor/utils/logging.py
@@ -1,9 +1,55 @@
-from typing import TYPE_CHECKING, List, Optional
+import contextvars
+import logging
+from contextlib import contextmanager
+from typing import TYPE_CHECKING, Iterator, List, Optional
 
 import bittensor as bt
 
 if TYPE_CHECKING:
     from gittensor.classes import FileScoreResult, ScoreBreakdown
+
+
+# Set per-task while a miner is being scored so concurrent evaluations stay
+# attributable in the log (see ``scoring_uid``). ``None`` means "not in a
+# per-miner scope" — the filter then leaves the line untouched.
+_scoring_uid: contextvars.ContextVar[Optional[int]] = contextvars.ContextVar('scoring_uid', default=None)
+
+
+class _UidLogFilter(logging.Filter):
+    """Prefix each log line with the UID of the miner currently being scored.
+
+    Attached to the ``bittensor`` logger, this runs synchronously in the thread
+    that emitted the record — including ``asyncio.to_thread`` workers, which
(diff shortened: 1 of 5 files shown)


seroperson:
Some time ago I was thinking about implementing this as well, but there were a lot of concerns like broken logging, some possible data races, error handling etc. Not so simple to implement correctly. Gl with attempt tho.

plind-junior:
@seroperson Would you review this PR as well?

anderdc:
This PR has already been attempted multiple times, I don't think I want to make this process parallel, it's already quite optimized, and scoring doesn't take long anyways, finally I also don't want to spam the mirror service or have gate it via api key.

PR #1352, opened 2026-05-23
feat(scoring): drop density, swap base score to exponential saturation

seroperson:
## Summary

Replaces the density-based base score with an exponential saturation curve on source token score, per the discussion and agreed shape in #1339:

```
base_score = MERGED_PR_BASE_SCORE * (1 - exp(-src_tok / SRC_TOK_SATURATION_SCALE))
           + min(total_score / CONTRIBUTION_SCORE_FOR_FULL_BONUS, 1) * MAX_CONTRIBUTION_BONUS
```

The new curve is monotonic in `src_tok`, concave from the origin, and ruled by one knob (`SRC_TOK_SATURATION_SCALE = 58`, per-repo overridable).

Per-repo override range is `[10, 500]`, enforced by the loader and pinned by the live-config test in `tests/validator/test_load_weights.py`.

### Requirements

| #   | Requirement                                                                                                            | Status |
| --- | ---------------------------------------------------------------------------------------------------------------------- | :----: |
| 1   | Ship formula `base_score = 25 * (1 - exp(-src_tok / 58)) + min(total_score / 1500, 1) * 5`                             |   ✅   |
| 2   | Drop contribution bonus cap to 5 instead of removing the term (keep tests/non-code non-zero, 5:1 source-vs-rest ratio) |   ✅   |
| 3   | Land formula + bonus change in the same PR                                                                             |   ✅   |
| 4   | Delete `MAX_CODE_DENSITY_MULTIPLIER`                                                                                   |   ✅   |
| 5   | Delete `MIN_TOKEN_SCORE_FOR_BASE_SCORE` from `scoring.py` (the cliff)                                                  |   ✅   |
| 6   | Delete `MIN_TOKEN_SCORE_FOR_BASE_SCORE` from `credibility.py` (`valid_merged_count` filter)                            |   ✅   |
| 7   | Eligibility becomes a pure count of merged PRs, per-repo overridable                                                   |   ✅   |
| 8   | Add `SRC_TOK_SATURATION_SCALE = 58`                                                                                    |   ✅   |
| 9   | Make `SRC_TOK_SATURATION_SCALE` per-repo configurable via `master_repositories.json` scoring block                     |   ✅   |
| 10  | `MAX_CONTRIBUTION_BONUS`: 25 → 5                                                                                       |   ✅   |
| 11  | Keep `MERGED_PR_BASE_SCORE = 25`                                                                                       |   ✅   |
[...]

PR #1369, opened 2026-05-26
fix(validator): don't double-score code moved between files

dripsmvcp:
Token scoring is a per-file AST diff that scores both added and deleted nodes. There's no cross-file reconciliation, so a function **moved** from one file to another is scored once as a deletion (in the source file) and once as an addition (in the destination file) — **2× for zero net new code**. This reconciles moves PR-wide so relocated code scores once, like writing it once.

## Root Cause

`score_tree_diff` computes a per-file symmetric difference and scores both sides (`gittensor/validator/utils/tree_sitter_scoring.py`); `calculate_token_score_from_file_changes` loops per file and **sums**. The docstring's "moving code around = no change" only holds *within* a single file (where a signature appears in both old and new and cancels). Across files there's no cancellation, so a relocation is counted on both sides.

Empirically, scoring a Python function moved A→B vs. written once:

```
ADD-ONLY (genuinely new func) : 5.250
MOVE A->B (zero net new code) : 10.500   <- 2.00x
```

This inflates `token_score`, which feeds eligibility/credibility gates, the dynamic open-PR threshold, the contribution bonus, and issue-discovery base scores. It's both an anti-gaming hole ("extract module" PRs on existing code earn 2×, repeatably) and a fairness issue (honest refactors over-rewarded). Note it gets **worse under #1352** (drop density), where `base_score` becomes a direct function of `src_tok` with no line-denominator to mask the doubling.

## Fix

`calculate_token_score_from_file_changes` now runs in two passes:

1. Classify each file; for tree-diff files, parse once and accumulate a PR-global **move budget** of all added signatures.
2. When scoring each file's *deleted* nodes, suppress those whose signature was added elsewhere in the PR (a relocation), capped by the actual added count. Genuine deletions (no matching add) still score; additions are untouched.

Because a single file's `added`/`deleted` are disjoint (Counter subtraction), **single-file PRs are completely unaffected** — only genuine cross-file moves net out. Processing is deterministic by file order (consensus-safe across validators), and files are parsed once (no perf regression). `score_tree_diff` is split into `diff_signatures` + `score_signatures` and keeps identical behavior for existing callers.

After the fix:

```
MOVE A->B : 5.250   <- 1.00x (same as writing it once)
```

## Trade-off (intentional)

For a real *delete-of-X + unrelated-add-of-Y*, the shared **structural type-count** (e.g. one `function_definition`) nets PR-wide, so that single structural unit isn't scored — but all content-specific tokens of both X and Y still score. This is consistent with the existing "structural nodes are type-only / moving code is no change" design, applied PR-wide.

## Test Plan

- [x] Regression tests: `tests/validator/test_token_scoring_integration.py::TestCrossFileMoveReconciliation`
  - `test_moved_function_scored_once_not_twice` (the 2×→1× regression)
  - `test_genuine_deletion_still_scored`
  - `test_delete_plus_unrelated_add_both_count`
  - `test_single_file_in_place_change_unaffected`
- [x] Full validator + utils suite: **642 passed**
- [x] `ruff check` + `ruff format --check` clean

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -149,34 +149,27 @@ def has_inline_tests(content: str, extension: str) -> bool:
     return pattern.search(content) is not None
 
 
-def score_tree_diff(
+def diff_signatures(
     old_content: Optional[str],
     new_content: Optional[str],
     extension: str,
     weights: TokenConfig,
-) -> ScoreBreakdown:
+) -> Tuple[Counter[NodeSignature], Counter[NodeSignature]]:
+    """Symmetric difference of a file's old vs new AST node signatures.
+
+    Returns ``(added, deleted)`` signature multisets:
+    - ``added``   = signatures in new but not old (in/out of scope by type/content)
+    - ``deleted`` = signatures in old but not new
+
+    Structural nodes are identified by type only (moving code *within a file* is no
+    change); leaf nodes by type + content. ``added`` and ``deleted`` are always
+    disjoint for a single file, so cross-file move reconciliation must happen at the
+    caller (see ``calculate_token_score_from_file_changes``).
     """
-    Calculate score by comparing old and new file ASTs using symmetric difference.
-    - Structural nodes are identified by type only (moving code around = no change)
-    - Leaf nodes are identified by type + content (actual token changes)
-    - Both additions (in new but not old) and deletions (in old but not new) are scored
-
-    Args:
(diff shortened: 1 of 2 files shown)


anderdc:
This shouldn't be a concern of the top level scoring algorithm, per repository maintainers should adjust their credibility and scoring hyperparameters if they want to mitigate that, or just accept less code that purely just moves things around

dripsmvcp:
@anderdc Hmm...maybe it could be resolved by repo maintainers. It's for the fair score calculation

PR #1374, opened 2026-05-26
fix(cli): guard miner score _drain_logs against Warning-state crash

statxc:
## Summary

`gitt miner score --log-level warning` crashed at shutdown with
`statemachine.exceptions.TransitionNotAllowed: Can't disable_logging when in Warning.`
This guards the `_drain_logs` cleanup path so every value Click already accepts in
`--log-level` completes cleanly.

Closes #1373 

## Root cause

bittensor's `LoggingMachine` defines `disable_logging` transitions from
`Trace`/`Debug`/`Default`/`Disabled`/`Info` but **not** from `Warning` — the only
shutdown transition out of `Warning` is `disable_warning = Warning.to(Default)`
(`bittensor/utils/btlogging/loggingmachine.py:127-137`). Any caller that lands in
`Warning` and then invokes `bt.logging.off()` raises `TransitionNotAllowed`.

`gitt miner score` exposes `warning` via `click.Choice([...])`, calls
`_apply_log_level('warning')` → `bt.logging.set_warning()`, then unconditionally
calls `bt.logging.off()` in `_drain_logs()`. Deterministic crash, not flaky.

## Fix

Single guarded transition in `_drain_logs()`: if the LoggingMachine is in
`Warning`, step down to `Default` via `set_warning(on=False)` first, then call
`off()` (which is defined from `Default`). This preserves the helper's existing
contract — `before_disable_logging` sets all stdlib loggers to `CRITICAL` — that
a naked skip-when-Warning would silently break.

State check follows bittensor's own internal idiom
(`set_debug`/`set_trace`/`set_info`/`set_warning` all compare
`current_state_value` before transitioning) and avoids using exceptions for
control flow.

## A/B reproduction

Before:
info exit=0
debug exit=0
trace exit=0
warning exit=1   ← TransitionNotAllowed in _drain_logs

After:
info exit=0
debug exit=0
trace exit=0
warning exit=0

## Tests

New `TestDrainLogsAcrossAdvertisedLevels` in `tests/cli/test_miner_score.py`:

- **Parametrized over all four advertised levels** (`warning`/`info`/`debug`/
  `trace`) — pins the regression and prevents drift if Click's `--log-level`
  choices ever expand.
- **`Disabled.to(Disabled)` idempotency** — a second drain after a successful
  drain must not raise.
- **`Warning → Disabled` contract** — proves the fix doesn't just sidestep the
  crash; the machine actually ends in `Disabled` so logging is genuinely off.

[...]

PR #1376, opened 2026-05-27
fix(validator): SCALE-encode str args in _encode_args (#1374)

### LINKED ISSUE #1375: [Bug] `gitt issues register` cannot encode `register_issue` string arguments after #1335
## Description

After #1335, `gitt issues register` no longer goes through `ContractInstance.exec(...)`. It now routes through the new `IssueCompetitionContractClient.register_issue(...)` method, which calls `_exec_contract_raw('register_issue', ...)` → `_encode_args('register_issue', ...)`.

That new path is broken before any extrinsic is submitted.

`metadata.json` declares `register_issue` with two `str` arguments:

- `github_url`
- `repository_full_name`

But `IssueCompetitionContractClient._encode_args()` only handles `u32`, `u64`, `u128`, `AccountId`, and `array32`. Anything else hits the fallback:

```python
```

So every `register_issue` call fails while encoding `github_url`, before `compose_call` is ever invoked.

## Affected Code

- `gittensor/cli/issue_commands/mutations.py` — `issue_register()` now calls `client.register_issue(...)` (introduced in #1335).
- `gittensor/validator/issue_competitions/contract_client.py`
  - `register_issue()` → `_exec_contract_raw(method_name='register_issue', ...)`
  - `_exec_contract_raw()` → `_encode_args(...)`
  - `_encode_args()` has no `str` branch.
- `gittensor/validator/issue_competitions/metadata.json` — `register_issue` declares `github_url` and `repository_full_name` as `str`.

## Steps to Reproduce

No live chain required — the failure happens entirely inside local argument encoding.

Against current `test` at `fc503f9`:

```python
from gittensor.validator.issue_competitions.contract_client import (
    IssueCompetitionContractClient,
)

# Bypass __init__ so we don't need a Subtensor connection.
client = IssueCompetitionContractClient.__new__(IssueCompetitionContractClient)

client._encode_args(
    "register_issue",
    {
        "github_url": "https://github.com/owner/repo/issues/1",
        "issue_number": 1,
        "target_bounty": 10_000_000_000,
    },
)
```

Observed:

```text
ValueError: Unsupported type: str for arg github_url
```

The full CLI path reaches the same encoder via:

```
gitt issues register
  → mutations.issue_register()
  → IssueCompetitionContractClient.register_issue(...)
  → _exec_contract_raw(method_name="register_issue", ...)
  → _encode_args("register_issue", ...)   # raises here
```

## Expected Behavior

`gitt issues register` should encode the `register_issue` arguments and submit the contract call. Specifically, `_encode_args` should SCALE-encode `str` as compact-length-prefixed UTF-8 bytes (the SCALE encoding of `Vec<u8>`).

## Actual Behavior

`register_issue` fails during local argument encoding, before any on-chain submission. The CLI can only surface a pre-submission error similar to:

```text
register_issue error: Unsupported type: str for arg github_url
```

## Impact

#1335 was intended to remove the local `.contract` artifact requirement for issue registration ("removes the runtime requirement to have `smart-contracts/issues-v0/target/ink/issue_bounty_manager.contract` built locally before registering an issue"). The replacement path achieves that goal but cannot register any issue, because `register_issue` always carries string arguments.

This breaks the owner-only `gitt issues register` workflow on current `test`.

#1335's own description notes that "Basic things were tested, but not the chain interaction itself" — this is the gap that comment was flagging.

## Suggested Fix

Add SCALE string encoding to `_encode_args()` for `type_def == "str"`:

- UTF-8-encode the value.
- Prepend the SCALE compact length of the byte string.
- Append the bytes.

The compact-length helper should cover at least modes 0 (`n < 64`), 1 (`64 ≤ n < 16384`), and 2 (`16384 ≤ n < 2³⁰`) so that longer repo names / URLs don't fail at the >64-byte boundary.

A regression test should cover `register_issue` specifically, e.g.:

- Construct the client via `__new__` (no chain dependency).
- Call `_encode_args("register_issue", {...})` with both a short URL and a URL long enough to push the encoded length past 64 bytes.
- Assert it does not raise.
- Assert the encoded bytes begin with the correct SCALE compact-length prefix followed by the UTF-8 bytes, in order, for `github_url` then `repository_full_name`, then `u32` `issue_number` (4 bytes LE), then `u128` `target_bounty` (16 bytes LE).
- Negative case: a non-`str` value for `github_url` raises a clear error.

## Prior Art / Not a Duplicate

#987 / #988 (and the parallel #993) proposed *removing* the `str` fallback as dead code, with bodies explicitly stating:

> "`str` appears only in `register_issue` (never routed here)… `register_issue` is never called through `_exec_contract_raw` → `_encode_args` — it is called from the CLI (`mutations.py`) via a different execution path."

That premise was true at the time, and those PRs were closed without merging.

#1335 inverted that premise: `register_issue` is now routed through `_exec_contract_raw()` → `_encode_args()`, making the previously unreachable `str` path reachable in production — and now broken, because the `str` branch was never implemented.

taillred:
## Summary

Fixes #1375

After #1335, `gitt issues register` routes through `IssueCompetitionContractClient.register_issue` → `_exec_contract_raw` → `_encode_args`. `_encode_args` had no `str` branch and raised `ValueError: Unsupported type: str for arg github_url` on every `register_issue` call, before any extrinsic was submitted.

This PR adds SCALE encoding for `str`: a compact-length prefix (modes 0/1/2 covered by a new `_scale_compact_length` helper) followed by the UTF-8 bytes — matching how `String` / `Vec<u8>` are encoded in SCALE.

## Changes

- `gittensor/validator/issue_competitions/contract_client.py`
  - New module-level `_scale_compact_length(n)` covering SCALE compact integer modes 0 (`n < 2⁶`), 1 (`n < 2¹⁴`), and 2 (`n < 2³⁰`). Rejects negative input and values that would require big-integer mode 3 (not reachable for any payload sent by the client).
  - New `'str'` branch in `_encode_args`: type-checks, UTF-8 encodes, prepends compact length.
- `tests/validator/test_contract_client_transactions.py`
  - `TestScaleCompactLength` — parametrized boundary coverage for every mode (0/1/2) plus negative and oversize rejection.
  - `TestEncodeArgsStr` — byte-level assertion of the full `register_issue` encoding: short URL (mode 0), long URL ≥64 bytes (mode 1), non-`str` rejected with a clear message, UTF-8 multibyte URL.

## Testing

- Full suite: `pytest tests/` → **892 passed**.
- Targeted: `pytest tests/validator/test_contract_client_transactions.py` → **43 passed** (28 pre-existing + 15 new).
- Reproducer from #1374 verified: raises on `fc503f9`, succeeds after this change with a 70-byte payload whose layout is byte-checked in the new tests.
- `ruff check` and `ruff format --check` clean repo-wide.

## Notes

- `_raw_contract_read` has an unrelated compact-length encoding branch (also flagged in #987/#988) that would overflow for `data_len ≥ 64`. It's currently only fed 4-byte selectors so it never triggers; out of scope for this PR but a candidate for follow-up using the new `_scale_compact_length` helper.
- `register_issue`'s `gas_limit` (`proof_size=1_000_000`) was set in #1335 and is unchanged here; this PR only fixes pre-submission encoding.


anderdc:
Can you show me the tests you ran to verify it worked. Local testnet subtensor with locally deployed contract is fine.

taillred:
Hi @anderdc  how are you? Thanks for reviewing. I've tested locally against substrate-contracts-node --dev with a stub contract that has the same register_issue signature + same selector (0x5c056a24). I've run gitt issues register twice - once with a short URL, once with a long one to hit both compact-length branches the PR adds:

  gitt issues register --contract 5C9feb... --network local \
    --repo entrius/gittensor --issue 1375 --bounty 10 --wallet local-test -y
  → ✓ tx 0x91df25dd...826ff4bf (URL = 48 bytes, mode 0)

  gitt issues register --contract 5C9feb... --network local \
    --repo electron-react-boilerplate/electron-react-boilerplate \
    --issue 3729 --bounty 10 --wallet local-test -y
  → ✓ tx 0x7037c6de...96ad1c93 (URL = 84 bytes, mode 1)

Read back via cargo contract call get_last_registered → both decode byte-perfect:
  ("https://github.com/entrius/gittensor/issues/1375", "entrius/gittensor", 1375, 10000000000)
  ("https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/3729", ..., 3729, 10000000000)

I used a stub, not the real contract, because the real one needs SubtensorExtension(5001) which stock substrate-contracts-node doesn't have.
The fix is in _encode_args though, which is upstream of any chain extension - the stub still exercises gitt's encoder - Contracts::call extrinsic - pallet-contracts dispatch - ink! decode - storage.
If you want the real contract in the loop, bittensor's public `test` chain has pallet-contracts + SubtensorExtension(5001), so I can deploy the real unmodified contract there (owned by my test wallet) and redo end-to-end.

anderdc:
Thanks for confirmation, I'll need to test this myself when I have a moment to spin up our dev environment

anderdc:
Fix conflicts. Tested locally, looks good.

anderdc (approved):

PR #1390, opened 2026-05-28
perf(cli): read a single on-chain issue by ID instead of scanning all

kinjitakabe:
**Why this is necessary:** `fetch_issue_from_contract` is used to act on *one* issue by ID (e.g. `gitt issues submissions --id <N>`), but it currently reads **every** issue registered on the contract to find it:

```python
issues = read_issues_from_contract(...)          # one childstate RPC per issue, 1..next_issue_id
issue = next((i for i in issues if i.get('id') == issue_id), None)
```

That's O(N) substrate RPC round-trips to fetch O(1) data. `next_issue_id` only ever grows — completed and cancelled issues are never removed from storage — so every single-issue command gets progressively slower and pounds the node with N round-trips for data whose storage key it can compute directly. The Ink! 5 lazy mapping key for a specific `issue_id` is already computable inline, so a single read was always possible; the code just wasn't using it.

**The fix:** add `read_issue_from_contract`, which computes the one issue's lazy key and does a single `childstate_getStorage` call, and route `fetch_issue_from_contract` through it. O(N) → O(1).

The per-issue decode is factored into a shared `_read_one_issue_from_child_storage` helper, so the full-scan path (`read_issues_from_contract`, used by `issues list`) and the new single-read path share one decode implementation instead of duplicating it. Behavior is unchanged: `fetch_issue_from_contract` still raises the same `ClickException`s for not-found, non-bountied, and incomplete issues.

## Related Issues

N/A

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] Other (performance)

## Testing

- New `tests/cli/test_issue_single_read.py`: single read does exactly one RPC,
  returns `None` when absent, `fetch_issue_from_contract` uses the single read
  (asserts the full scan is **not** called), and the not-found / non-bountied
  paths still raise `ClickException`.
- `uv run pytest` — 882 passed.
- `uv run pyright` / `ruff check` / `vulture` on changed files — clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -672,6 +672,53 @@ def _read_contract_packed_storage(substrate, contract_addr: str, verbose: bool =
         return None
 
 
+_ISSUE_STATUS_NAMES = ['Registered', 'Active', 'Completed', 'Cancelled']
+
+
+def _read_one_issue_from_child_storage(
+    substrate, child_key: str, issue_id: int, verbose: bool = False
+) -> Optional[Dict[str, Any]]:
+    """Read and decode a single issue from contract child storage by ID (one RPC)."""
+    encoded_id = struct.pack('<Q', issue_id)
+    lazy_key = compute_ink5_lazy_key(ISSUES_MAPPING_ROOT_KEY, encoded_id)
+
+    val_result = substrate.rpc_request('childstate_getStorage', [child_key, lazy_key, None])
+    if not val_result.get('result'):
+        if verbose:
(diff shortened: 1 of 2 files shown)


anderdc:
fix conflicts

anderdc:
Fix conflicts. Heads up: test now distinguishes a contract read error (ClickException "Error reading from contract: …") from a genuine not-found in fetch_issue_from_contract. read_issue_from_contract here swallows exceptions and returns None, which would report connection failures as "Issue ID N not found on-chain." Preserve that distinction when you rebase.

PR #1391, opened 2026-05-28
perf(cli): reuse subtensor connection in issues pending-harvest

plind-junior:
**Why this is necessary:** `gitt issues pending-harvest` opens **two** WebSocket
connections to the same node for a single command — a `bt.Subtensor` (used for the
treasury-stake read via the contract client) and then a separate
`SubstrateInterface(url=ws_endpoint)` for the issue read:

```python
subtensor = bt.Subtensor(network=ws_endpoint)
...
treasury_stake = client.get_treasury_stake()

substrate = SubstrateInterface(url=ws_endpoint)   # second connection to the same node
issues = _read_issues_from_child_storage(substrate, contract_addr, verbose)
```

The bittensor `Subtensor` already wraps a `SubstrateInterface` at
`subtensor.substrate`, exposing the exact `rpc_request` / `query` API the issue
read uses (the contract client reads through `subtensor.substrate` already). So
the second connection is pure redundancy — extra socket setup/teardown against
the same endpoint for one command.

**The fix:** read issues through `subtensor.substrate` and drop the second
`SubstrateInterface` plus its now-unused import. One command, one connection.
Behavior is unchanged — same data, same code path, just one fewer socket.

## Related Issues

N/A

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] Other (performance)

## Testing

- New `tests/cli/test_pending_harvest_single_connection.py`: asserts
  `pending-harvest --json` succeeds, opens **no** second `SubstrateInterface`,
  and the issue read reuses `subtensor.substrate`.
- Existing `--json` exception-path regression (`pending-harvest`) still passes.
- `uv run pytest` — 882 passed.
- `uv run pyright` / `ruff check` / `vulture` on changed files — clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -268,7 +268,6 @@ def issues_pending_harvest(network: str, rpc_url: str, contract: str, verbose: b
 
     try:
         import bittensor as bt
-        from async_substrate_interface import SubstrateInterface
 
         from gittensor.validator.issue_competitions.contract_client import (
             IssueCompetitionContractClient,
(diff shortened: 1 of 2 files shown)


anderdc (approved):
Verified: subtensor.substrate exposes the same API the issue read uses (contract client already reads through it). Drops a redundant second connection; behavior unchanged.

PR #1397, opened 2026-05-29
fix(scoring): gate valid merged PRs on source token score

### LINKED ISSUE #1396: Eligibility "valid merged PR" gate counts test content the base-score gate ignores
## Description

The OSS contribution scoring path applies a single threshold constant, `min_token_score_for_base_score` (default `MIN_TOKEN_SCORE_FOR_BASE_SCORE = 5`), in two places, but each place compares it against a different quantity:

- The base-score gate in `calculate_base_score_for_pr_files` (`gittensor/validator/oss_contributions/mirror/scoring.py`) keys off the **SOURCE-only** token score. A PR below the threshold earns zero source base score.
- The eligibility "valid merged PR" gate in `check_eligibility` (`gittensor/validator/oss_contributions/credibility.py`) keys off `pr.token_score`, which is the **all-category aggregate** (`SOURCE + TEST`, because test and inline-test files are tree-diff scored and folded into `score_breakdown.total_score`).

Because the two gates measure different scores against the same constant, they disagree for the same PR. The constant's comment and the `check_eligibility` docstring both state that "valid" means "clears the base-score threshold", so the current behavior contradicts the documented intent.

## Steps to Reproduce

1. Take a merged PR whose SOURCE token score is just under the threshold (for example `4`) but whose aggregate `SOURCE + TEST` score clears it (for example `10`) thanks to a small test or inline-test file.
2. Run the PR through OSS scoring and `finalize_miner_scores`.
3. Inspect the base score and the eligibility count for that PR.

## Expected Behavior

A PR counts as a "valid merged PR" only when it clears the same SOURCE-only threshold that grants it a source base score. Test content alone should not make a PR "valid".

## Actual Behavior

The PR earns zero source base score (only the cross-category contribution bonus remains) yet still counts toward `min_valid_merged_prs`. Since `min_valid_merged_prs` is the gate that unlocks all scoring for a repository, a near-threshold source PR plus a small test file can reach eligibility without contributing meaningful source work.

## Environment

- OS: any
- Python version: 3.11
- Commit/Version: `test` branch at 357917c

## Additional Context

The existing unit test `tests/validator/oss_contributions/mirror/test_base_score_helper.py` zeroes the aggregate to isolate the SOURCE gate, so it never exercises the divergence. The issue discovery path is unaffected: its validity check uses a separate constant (`min_token_score_for_valid_issue`) compared consistently against the aggregate, so there is no internal mismatch there.

Dexterity104:
## Summary

The OSS scoring path measured one threshold constant against two different quantities. The base-score gate in `calculate_base_score_for_pr_files` keyed off the SOURCE-only token score, while the eligibility "valid merged PR" gate in `check_eligibility` keyed off `pr.token_score`, the all-category aggregate (`SOURCE + TEST`). As a result a merged PR could earn zero source base score yet still count toward `min_valid_merged_prs`, the gate that unlocks all scoring for a repository. A near-threshold source PR plus a small test file was enough to flip a PR from "not valid" to "valid".

This change makes validity match the base-score gate:

- `BaseScoreResult` now returns `source_token_score` (already computed inside the helper, so no extra work).
- `ScoredPR` stores `source_token_score` alongside the aggregate `token_score`, set during `score_pr`.
- `check_eligibility` counts a merged PR as valid only when its `source_token_score` clears the threshold.
- The constant comment and the `check_eligibility` docstring now describe the actual behavior.

The aggregate `token_score` is still kept for reporting rollups, the storage adapter, and the issue discovery cache, so both fields remain in use. The issue discovery path is intentionally untouched: it gates on a separate constant (`min_token_score_for_valid_issue`) compared consistently against the aggregate, so it has no internal mismatch.

## Related Issues

Fixes #1396

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added `test_validity_gate_keys_off_source_not_aggregate_token_score`, which confirms a repo of merged PRs that pass the aggregate threshold but fail the SOURCE threshold is ineligible. Updated the `_merged` test helper to set the gating field, extended the base-score helper test to assert `source_token_score` is surfaced, and updated the `BaseScoreResult` stubs in `test_scoring.py`. The eligibility gate logic was verified directly: SOURCE-fail rows produce an ineligible result and SOURCE-pass rows produce an eligible one.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -120,7 +120,7 @@ INLINE_TEST_PATTERNS: Dict[str, re.Pattern] = {
 # Eligibility Gate (OSS Contributions)
 # =============================================================================
 # Per-repo defaults — each repo may override these in master_repositories.json.
-MIN_VALID_MERGED_PRS = 3  # minimum "valid" merged PRs (token_score >= MIN_TOKEN_SCORE_FOR_BASE_SCORE) to receive score
+MIN_VALID_MERGED_PRS = 3  # minimum "valid" merged PRs (source_token_score >= MIN_TOKEN_SCORE_FOR_BASE_SCORE) to receive score
 MIN_CREDIBILITY = 0.80  # minimum credibility ratio to receive score
 
 # =============================================================================
--- a/gittensor/validator/oss_contributions/credibility.py
+++ b/gittensor/validator/oss_contributions/credibility.py
@@ -29,8 +29,12 @@ def check_eligibility(
 
     Gate requires:
     1. At least ``cfg.min_valid_merged_prs`` merged PRs with
-       ``token_score >= cfg.min_token_score_for_base_score``
+       ``source_token_score >= cfg.min_token_score_for_base_score``
     2. At least ``cfg.min_credibility`` credibility
+
+    Validity keys off ``source_token_score`` (SOURCE-only) so it matches the
+    base-score gate; the aggregate ``token_score`` (SOURCE + TEST) must not let
+    test content flip a PR to "valid" without earning a source base score.
 
     Returns:
         (is_eligible, credibility, reason)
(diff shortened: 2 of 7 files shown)


anderdc:
The eligibility gate this patches no longer exists on test. check_eligibility now counts merged PRs directly against min_valid_merged_prs and no longer compares token_score against min_token_score_for_base_score, so the SOURCE-vs-aggregate mismatch from #1396 is already gone and this diff conflicts against the current path. Closing as out of date.

PR #1398, opened 2026-05-29
Range-check label multipliers and fixed base score during registry load

### LINKED ISSUE #1399: Registry validation skips label_multipliers, default_label_multiplier, and fixed_base_score
## Description

`load_master_repo_weights()` runs `_validate_eligibility_configs` and `_validate_scoring_configs` to range-check almost every per-repo field, and fails fast at startup when a value is out of bounds. Three scoring fields slip past both validators:

- `label_multipliers`
- `default_label_multiplier`
- `fixed_base_score` (also never coerced to a number, stored raw from JSON)

These feed live scoring exactly like fields that are already bounded. `label_multipliers` flows through `resolve_highest_label_multiplier` and multiplies into a PR's earned score, while the structurally identical issue-bonus multipliers (`standard_issue_multiplier`, `maintainer_issue_multiplier`) are clamped to `[1, 5]`. `fixed_base_score` is assigned straight onto `ScoredPR.base_score`.

`master_repositories.json` is the most frequently edited file in the repo, and nearly every recent merge is a `chore(weights)` PR touching these exact fields. A single typo (`5.0` becomes `50.0`, or a stray `-`) would not fail at validator startup the way every other mis-set field does. It would silently distort scoring and emissions. A non-numeric `fixed_base_score` raises a `TypeError` deep inside the multiplier math that `score_pr` swallows, leaving the PR silently unscored.

The bounds already exist as live-config assertions in `tests/validator/test_load_weights.py`, but those only run against the committed file in CI. They never protect a validator that loads a config at runtime.

## Steps to Reproduce

1. Edit `gittensor/validator/weights/master_repositories.json` and set a label multiplier out of range, for example:
   ```json
   {
     "owner/repo": {
       "emission_share": 0.5,
       "label_multipliers": { "kind/feature": 50.0 }
     }
   }
   ```
2. Call `load_master_repo_weights()`.
3. Repeat with `fixed_base_score: -1.0` and with `fixed_base_score: "oops"`.

## Expected Behavior

Loading raises `RepositoryRegistryError` at startup, the same fail-fast behavior every other scoring field already has.

## Actual Behavior

The config loads successfully. The bad multiplier silently distorts scoring, and a non-numeric `fixed_base_score` surfaces later as a swallowed `TypeError` that leaves PRs unscored.

## Environment

- OS: any
- Python version: project default
- Commit/Version: `test` branch, current head

Dexterity104:
## Summary

`load_master_repo_weights()` already range-checks nearly every per-repo field through `_validate_eligibility_configs` and `_validate_scoring_configs`, failing fast at startup when a value is out of bounds. Three scoring fields were left out: `label_multipliers`, `default_label_multiplier`, and `fixed_base_score`. The last one was not even coerced to a number, so it was stored raw from JSON.

These fields drive live scoring. `label_multipliers` resolves through `resolve_highest_label_multiplier` and multiplies into a PR's earned score, while the structurally identical issue-bonus multipliers are already clamped to `[1, 5]`. `fixed_base_score` is assigned straight onto `ScoredPR.base_score`. Because `master_repositories.json` is the most edited file in the repo and almost every recent merge touches these exact fields, a single typo such as `5.0` becoming `50.0` would pass validation and quietly distort emissions, and a non-numeric `fixed_base_score` would later raise a `TypeError` that `score_pr` swallows, leaving the PR unscored.

This change adds `_validate_label_configs`, wired into the same load chain as the existing validators, so these fields fail fast like the rest of the registry. Bounds mirror the live-config assertions already present in the tests:

- `label_multipliers`: at most 10 entries, each value within `[0, 20]`
- `default_label_multiplier`: within `[0, 20]`
- `fixed_base_score`: a number (not a bool), within `[0, 100]`

The `[0, 20]` range is used rather than the issue-bonus `[1, 5]` because live configs legitimately use `0` and sub-1 multipliers, so the issue-bonus bounds would reject valid data. Every currently valid config keeps its behavior, and the `fixed_base_score` docstring now points at the load-time check instead of the test.

## Related Issues

Fixes #1399

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Parametrized rejection tests cover negative, over-cap, non-numeric, bool, and too-many-entries cases for all three fields. The live `master_repositories.json` loads cleanly through the full validation chain, confirming no valid config changes behavior.

```bash
uv run pytest tests/validator/test_load_weights.py -q
```

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/utils/load_weights.py
+++ b/gittensor/validator/utils/load_weights.py
@@ -152,8 +152,8 @@ class RepositoryConfig:
             same fnmatch wildcard syntax as ``additional_acceptable_branches``.
         default_label_multiplier: Multiplier used when no configured label
             pattern matches. Defaults to neutral scoring.
-        fixed_base_score: Override for the PR base score. Expected
-            to be within [0.0, 100.0]; range is enforced by the live-config test.
+        fixed_base_score: Override for the PR base score. Must be within
+            [0.0, 100.0]; range is enforced at load time by ``_validate_label_configs``.
         eligibility: Per-repo overrides for the eligibility / spam knobs. Unset
             fields fall back to the global default constants — see
             ``resolve_eligibility``.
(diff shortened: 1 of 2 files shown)


anderdc:
These three fields are already range-checked in CI by the live-config tests in tests/validator/test_load_weights.py (test_live_mirror_scoring_fields_have_valid_shape and the live label-multiplier tests), which assert the identical bounds against the committed master_repositories.json — the same file the validator loads. A bad value in a chore(weights) PR fails CI before merge, so the load-time check duplicates an existing gate rather than closing a real gap. Closing.

PR #1129, opened 2026-05-09
fix: gitt miner check exits with code 1 when no validators have a valid PAT (#842)

alpurkan17:
## Summary

- Return exit code 1 when valid_count == 0 in check.py
- Add exit code documentation to CLI help
## Root Cause

`check.py` iterates over configured validators and displays PAT status, but exits with code 0 even when `valid_count == 0` (all PATs are invalid). Downstream scripts rely on exit codes to detect failure, so they miss the all-PATs-invalid state.
## Impact

Automation scripts that call `gitt miner check` don't detect when all PATs are invalid. Miners remain unaware their setup has no valid tokens.
## Related Issues

Fixes #842
## Test plan

- [x] ruff check
- [x] ruff format --check
- [x] pyright
- [x] pytest tests/cli/test_check_exit_code.py -q
### Live verification

- [x] `gitt miner check --network finney; echo Exit: $?` shows 1 when no valid PATs
### How to verify

1. Set up a config with invalid PATs (or no PATs)
1. Run `gitt miner check --network finney; echo $?`
1. Verify exit code is 1 (not 0)
### Edge cases considered

- Empty validators list (no validators configured at all) — should also exit 1
- Mix of valid and invalid PATs — should exit 0 (some are valid)
### Post-merge verification

- [ ] Confirm exit code 1 when valid_count==0 before next scoring cycle
- [ ] Update any CI scripts that call gitt miner check to handle exit 1

Issue #1392, opened 2026-05-28
PAT validation fails when token has leading or trailing whitespace


techforgeworks:
## Description

Validators and the miner CLI reject GitHub PATs that contain leading or trailing whitespace (common when the token is copied from `.env` files or secret managers with a trailing newline). The Authorization header is built from the raw string, so `/user` returns 401 even though the token is otherwise valid.

## Steps to Reproduce

1. Store a valid PAT with a trailing newline, e.g. `GITTENSOR_MINER_PAT="ghp_xxxx\n"` or paste a token with spaces into `gitt miner post`.
2. Run `gitt miner post` or broadcast the PAT to a validator.
3. Observe validation failure despite the same token working after manual trimming.

## Expected Behavior

Whitespace around the PAT is ignored before GitHub API calls and before the PAT is persisted.

## Actual Behavior

The untrimmed token is sent to GitHub and validation fails.

## Environment

- OS: Any
- Python version: 3.12+
- Commit/Version: 9ac03b1

## Additional Context

Affected entry points include `get_github_identity`, `validate_github_credentials_result`, and `handle_pat_broadcast`.

PR #1407, opened 2026-05-29
fix(cli): distinguish inconclusive PAT reply from no-response in miner check

### LINKED ISSUE #1406: [Bug] `gitt miner check` reports a validator's inconclusive PAT reply (has_pat=True, pat_valid=None) as "no response"
## Description

`gitt miner check` buckets each validator's `PatCheckSynapse` reply through `_pat_check_row_category` in `gittensor/cli/miner_commands/helpers.py:175-183`:

```python
def _pat_check_row_category(row: dict[str, Any]) -> str:
    """Classify a PAT probe row; must match `miner check` table rendering order."""
    if row.get('pat_valid') is True:
        return 'valid'
    if row.get('has_pat') is False:
        return 'no_pat'
    if row.get('has_pat') is True and row.get('pat_valid') is False:
        return 'invalid_pat'
    return 'no_response'
```

There is no branch for `has_pat=True, pat_valid=None`. But that is exactly the shape the validator emits on a *transient* GitHub failure. PR #1107 (merged, `c40de7a`) added a third outcome to `handle_pat_check` in `gittensor/validator/pat_handler.py:121-129`:

```python
    synapse.has_pat = True
    # Re-validate the stored PAT
    validation = validate_github_credentials_result(uid, entry['pat'], stored_github_id=entry.get('github_id'))
    if validation.transient_failure:
        synapse.pat_valid = None
        synapse.rejection_reason = 'GitHub API temporarily unavailable; retry the check in a few minutes.'
        return synapse
```

`PatCheckSynapse.has_pat` / `pat_valid` are `Optional[bool]` (`gittensor/synapses.py:44-45`), so `None` survives serialization; `check.py:113-114` reads them with `getattr(resp, ...)`. The inconclusive reply therefore arrives as `has_pat=True, pat_valid=None` and falls through every branch above to `'no_response'` — the same bucket as a validator that never replied (`has_pat=None, pat_valid=None`).

Consequences:
- **Table mode** renders the Status column as `[yellow]— no response[/yellow]` (`_PAT_CHECK_STATUS_MARKUP`, `check.py:32-37`), which means *the validator never answered* — the opposite of what happened.
- **`--json` mode** (`check.py:130-142`): `_pat_check_aggregate_counts` (`helpers.py:186-194`) counts it under `no_response` and excludes it from `valid`, and `success` is computed as `valid_count > 0`. There is no Reason column in JSON, so machine consumers cannot tell a healthy validator (holding the PAT) from a dead one.

## Steps to Reproduce

Categorizer (pure function, no network):

```python
from gittensor.cli.miner_commands.helpers import _pat_check_row_category, _pat_check_aggregate_counts

inconclusive = {'has_pat': True, 'pat_valid': None,
                'rejection_reason': 'GitHub API temporarily unavailable; retry the check in a few minutes.'}
dead         = {'has_pat': None, 'pat_valid': None}

print(_pat_check_row_category(inconclusive))   # 'no_response'
print(_pat_check_row_category(dead))           # 'no_response'  -> indistinguishable
print(_pat_check_aggregate_counts([inconclusive]))
# {'valid': 0, 'no_pat': 0, 'invalid_pat': 0, 'no_response': 1}
```

End-to-end: with a PAT stored for the miner's UID, make GitHub's `/user` lookup fail transiently (so `get_github_identity` returns `GitHubIdentityStatus.TRANSIENT_FAILURE`). `handle_pat_check` returns `has_pat=True, pat_valid=None` — the exact shape asserted by the merged validator test `tests/validator/test_pat_handler.py:234-242` (`test_transient_identity_lookup_reports_inconclusive`). Run `gitt miner check`: that validator's row shows `— no response` and is dropped from the valid count.

## Expected Behavior

A reply where the validator *did* respond and confirmed it holds the PAT, but could not confirm validity because GitHub was momentarily unavailable, should be its own status (e.g. `inconclusive` / `unconfirmed`), distinct from `no_response`. The miner should be able to tell "the validator has my PAT, GitHub just blipped — retry later" apart from "the validator never answered / doesn't have my PAT", in both table and `--json` output.

## Actual Behavior

The inconclusive reply is bucketed as `no_response`. Table mode shows `— no response` (contradicting the truthful `rejection_reason` in the adjacent Reason column); `--json` counts it under `no_response`, drops it from `valid`, and can flip `success` to `false` — so automated consumers silently conflate a healthy validator holding the PAT with a dead one. A miner may needlessly re-broadcast their secret PAT or conclude a healthy validator is offline.

## Relationship to existing work

- **PR #1107 (MERGED, `c40de7a`)** introduced the producer-side `pat_valid=None` state. `git show --stat c40de7a` confirms it touched **only** `gittensor/validator/pat_handler.py` and its test — never the CLI consumer. This issue is the missing consumer-side half.
- The closed precursors of #1107 — **#1123 / #1143 / #1175 / #1170** (and issue **#1106**) — are all validator-side (`pat_handler.py` only); none touches `_pat_check_row_category`. (#1127 edits `helpers.py` but only `_require_validator_axons`, not the categorizer.)
- **This is not the same as #1106 "already fitting" the CLI.** #1106's body notes that `pat_valid=None` "fits the existing CLI classification: `_pat_check_row_category()` maps that state to `no_response`, not `invalid_pat`." That observation only establishes the inconclusive state is **not mislabeled as `invalid_pat`** — which is true and not what this issue is about. The defect here is that it is **indistinguishable from `no_response` (validator never answered)**: a previously-accepted stored PAT during a GitHub blip renders identically to an unreachable validator. That conflation was never considered in #1106 and is lost information, not a cosmetic relabel.
- **NOT an exit-code change.** Distinct from the closed **#1129** (`gitt miner check` exit-1 when no valid PAT) and **#542** ("no one accepted the PAT is simply the output"), both of which argue exit-code/`success` policy for a *valid zero* result and were closed as working-as-intended. Here the validator genuinely replied confirming it holds the PAT, so labeling it `no response` is *factually wrong about the response that arrived* — a categorization-accuracy fix. **This proposal does not alter exit codes or the `success = valid_count > 0` rule.**
- **Positive precedent:** PRs **#724** and **#807** (both MERGED) fixed exactly this class of PAT-bucket-accuracy defect in this same CLI — #724 introduced `_pat_check_row_category` itself (un-conflating `no_pat` vs `invalid_pat`); #807 mirrored it for `gitt miner post` (un-folding `no_response` vs `rejected`). Un-conflating `inconclusive` from `no_response` is the same species of change.

## Affected Files

- `gittensor/cli/miner_commands/helpers.py:175-194` — `_pat_check_row_category` and `_pat_check_aggregate_counts`.
- `gittensor/cli/miner_commands/check.py:32-37` (the `_PAT_CHECK_STATUS_MARKUP` map), `check.py:150-153` (the table render, which does a bare `_PAT_CHECK_STATUS_MARKUP[category]` lookup), and `check.py:130-142` (the `--json` payload, which spreads `**counts`).
- `tests/cli/test_miner_commands.py:317-330` — `TestPatCheckAggregateCounts` covers `{pat_valid:None, has_pat:None}` but never `{pat_valid:None, has_pat:True}`.

## Suggested Fix Shape

Add a fifth category for the inconclusive state and surface it everywhere the four existing ones are. All four edits below must land together:

1. In `_pat_check_row_category`, before the final `return 'no_response'` and after the `invalid_pat` branch (so ordering still mirrors the validator outcomes), add:
   ```python
   if row.get('has_pat') is True and row.get('pat_valid') is None:
       return 'inconclusive'
   ```
   The dead-validator case (`has_pat=None, pat_valid=None`) still misses this branch and correctly stays `no_response`.
2. Add the `inconclusive` key to `_pat_check_aggregate_counts`'s returned dict (ordered between `invalid_pat` and `no_response`). This automatically surfaces in `--json` via the existing `**counts` spread — no separate JSON edit needed.
3. Add an entry to `_PAT_CHECK_STATUS_MARKUP` in `check.py`, e.g. `'inconclusive': '[yellow]? inconclusive[/yellow]'`. **Required, not optional:** the table renders via `status = _PAT_CHECK_STATUS_MARKUP[category]` (`check.py:152`), a bare dict lookup — a new category without a markup entry raises `KeyError` at render time.
4. **Update** the existing `TestPatCheckAggregateCounts` assertion (it compares the returned dict with `==`, so a new key breaks it — add `'inconclusive': 0`), **and add** a unit test covering `{'has_pat': True, 'pat_valid': None}` asserting it is `'inconclusive'` and explicitly **not** `'no_response'`.

## Scope / Out of Scope

**In scope:** the CLI consumer mapping of the already-shipped `pat_valid=None` reply — `_pat_check_row_category`, `_pat_check_aggregate_counts`, the `check.py` markup map and `--json` payload, plus the test updates above.

**Out of scope:** the validator-side producer (`pat_handler.py`, already handled by merged #1107); any exit-code / `success`-rule change (cf. closed #1129 / #542); PAT-whitespace handling (#1392 / #1393); and `gitt miner post` categorization (`_pat_post_row_category`), which has a different reply shape.

taillred:
## Summary

Fixes #1406 

After #1107, `handle_pat_check` returns `has_pat=True, pat_valid=None` on a transient GitHub failure — a validator that *has* the miner's PAT but couldn't confirm validity because GitHub's `/user` lookup momentarily failed. The CLI consumer `_pat_check_row_category` had no branch for that shape, so it fell through to `no_response`, rendering a healthy validator identically to one that never replied:

- **Table mode**: shows `— no response` (contradicting the truthful "retry in a few minutes" reason in the adjacent column).
- **`--json` mode**: counted under `no_response`, excluded from `valid`, and can flip `success` to `false`.

This adds a fifth `inconclusive` category so the two states are distinguishable. It does **not** change exit codes or the `success = valid_count > 0` rule (cf. the closed #1129 / #542, which were exit-code/`success` policy for a *valid zero* result) — it only fixes the categorization of a reply that genuinely arrived. This is the same class of accuracy fix as the merged #724 / #807 in this same file.

## Relationship to existing work

- **#1107 (merged)** added the producer-side `pat_valid=None` state, touching only `validator/pat_handler.py` + its test — it never updated the CLI consumer. This PR is that missing consumer-side half.
- **#1106** noted `pat_valid=None` "fits the existing CLI classification … maps that state to `no_response`, not `invalid_pat`." That only establishes the inconclusive reply isn't mislabeled as `invalid_pat` (true). The defect fixed here is different: it was **indistinguishable from `no_response` (never answered)** — lost, operator-actionable information.

## Changes

- `gittensor/cli/miner_commands/helpers.py`
  - New `inconclusive` branch in `_pat_check_row_category`, placed after the `invalid_pat` branch and guarded on `has_pat is True` so a genuinely silent validator (`has_pat=None, pat_valid=None`) still resolves to `no_response`.
  - `inconclusive` key added to `_pat_check_aggregate_counts` (auto-surfaces in `--json` via the existing `**counts` spread).
- `gittensor/cli/miner_commands/check.py`
  - `inconclusive` entry in `_PAT_CHECK_STATUS_MARKUP`. **Required**, not cosmetic: the table renders via a bare `_PAT_CHECK_STATUS_MARKUP[category]` lookup, so a new category without a markup entry would raise `KeyError`.
- `tests/cli/test_miner_commands.py`
  - `TestPatCheckRowCategory`: every category, including `inconclusive` ≠ `no_response`, silent-validator → `no_response`, and missing-keys → `no_response`.
  - Updated the existing aggregate-counts assertion to include `inconclusive`, plus a non-collapse test.
  - A categorizer↔markup parity guard (`test_every_category_has_table_markup`) so a future category addition can't silently reintroduce the `KeyError`.
  - CLI-level `gitt miner check --json` integration tests (parity with the sibling `miner post` test): inconclusive surfaces distinctly, `success` stays `False` for an inconclusive-only result, and the table render does not `KeyError`.

## Testing

- `pytest tests/` → **892 passed**.
- `ruff check` / `ruff format --check` → clean.
- Verified the new tests fail against pre-fix code (regression locked).

## Scope

Consumer-side categorization only. Out of scope: the validator producer (`pat_handler.py`, already merged in #1107), exit-code / `success`-rule changes, PAT-whitespace handling (#1392/#1393), and `gitt miner post` categorization (different reply shape — `PatBroadcastSynapse` has no `pat_valid` field).

### DIFF
--- a/gittensor/cli/miner_commands/check.py
+++ b/gittensor/cli/miner_commands/check.py
@@ -33,6 +33,7 @@ _PAT_CHECK_STATUS_MARKUP = {
     'valid': '[green]✓ valid[/green]',
     'no_pat': '[red]✗ no PAT[/red]',
     'invalid_pat': '[red]✗ invalid[/red]',
+    'inconclusive': '[yellow]? inconclusive[/yellow]',
     'no_response': '[yellow]— no response[/yellow]',
 }
 
--- a/gittensor/cli/miner_commands/helpers.py
+++ b/gittensor/cli/miner_commands/helpers.py
@@ -180,6 +180,8 @@ def _pat_check_row_category(row: dict[str, Any]) -> str:
         return 'no_pat'
     if row.get('has_pat') is True and row.get('pat_valid') is False:
         return 'invalid_pat'
+    if row.get('has_pat') is True and row.get('pat_valid') is None:
+        return 'inconclusive'
     return 'no_response'
 
 
@@ -190,6 +192,7 @@ def _pat_check_aggregate_counts(results: list[dict[str, Any]]) -> dict[str, int]
         'valid': counts['valid'],
         'no_pat': counts['no_pat'],
         'invalid_pat': counts['invalid_pat'],
+        'inconclusive': counts['inconclusive'],
         'no_response': counts['no_response'],
     }
 
--- a/tests/cli/test_miner_commands.py
+++ b/tests/cli/test_miner_commands.py
@@ -326,6 +326,7 @@ class TestPatCheckAggregateCounts:
             'valid': 1,
             'no_pat': 1,
             'invalid_pat': 1,
+            'inconclusive': 0,
             'no_response': 1,
         }
 


anderdc (changes requested):
too many tests for a two line change, remove the tests

taillred:
Hi @anderdc, how are you. I've fixed. Thanks for the review.

anderdc (approved):

taillred:
Thanks @anderdc

PR #1414, opened 2026-05-29
[Bug]: int/str github_id mismatch breaks duplicate penalty and bounty lookup

### LINKED ISSUE #1413: [Bug]: int/str github_id mismatch breaks duplicate penalty and bounty lookup

## Summary

## Steps to reproduce

## Expected behavior

## Actual behavior

## Impact

glorysr1209-png:
## Summary

Closes #1413.

Normalize GitHub IDs to strings in credential validation, duplicate detection, and issue bounty `registered_miners` mapping.

## Changes

- `str()` coercion in `github_validation`, `inspections`, and `forward.py`.
- Regression tests for int/str duplicate penalty and bounty solution vote.

## Real Behavior Proof

- [x] `test_duplicate_detection_penalizes_int_and_str_github_id`
- [x] `test_registered_miner_with_int_github_id_receives_solution_vote`

### What I ran

```bash
python3 -m pytest tests/validator/test_duplicate_penalty_propagation.py::test_duplicate_detection_penalizes_int_and_str_github_id tests/validator/test_issue_competitions_forward.py::test_registered_miner_with_int_github_id_receives_solution_vote -q
```

### What I observed

Both int/str collision paths behave consistently after normalization.

### DIFF
--- a/gittensor/validator/issue_competitions/forward.py
+++ b/gittensor/validator/issue_competitions/forward.py
@@ -67,9 +67,9 @@ async def issue_competitions(
         # penalized in oss_contributions() before this pass and arrive here as
         # failed evaluations.
         registered_miners = {
-            eval.github_id: eval.hotkey
+            str(eval.github_id): eval.hotkey
             for eval in miner_evaluations.values()
-            if eval.github_id and eval.github_id != '0' and eval.failed_reason is None
+            if eval.github_id and str(eval.github_id) != '0' and eval.failed_reason is None
         }
         bt.logging.info(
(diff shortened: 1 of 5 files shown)


anderdc:
github_id is already coerced to a string at the source — get_github_identity returns str(user_id), and the bounty lookup already wraps the solver in str(solver_github_id). In production github_id is never an int, so these consumer-side str() wraps guard a collision that can't occur; the only way to hit it is hand-constructing a MinerEvaluation with an int, as the added tests do. Closing.

PR #1416, opened 2026-05-30
fix: normalize github_id to str, prevent false cancel votes on truncated timelines

### LINKED ISSUE #1411: [Bug]: Issue bounty forward false cancel when closure timeline truncated
# [Bug]: Issue bounty forward can vote cancel when closure timeline is truncated

## Summary

`check_github_issue_closed()` (used by `issue_competitions/forward.py`) loads only the last 100 `CLOSED_EVENT` timeline nodes. On busy issues, the current close event may fall outside that window. The helper then returns no solver with `solver_lookup_failed=False`, and the forward pass calls `vote_cancel_issue` even when a merged PR closed the issue.

## Steps to reproduce

1. Configure a validator with `GITTENSOR_VALIDATOR_PAT` and an active bounty issue that is **completed** on GitHub (closed by merged PR).
2. Use an issue whose timeline has many `CLOSED_EVENT` entries (reopen/close history, or >100 close events).
3. Ensure the authoritative current `closedAt` close event is not among the last 100 timeline nodes returned by GraphQL.
4. Run the issue bounties forward pass.

## Expected behavior

When the current close event cannot be resolved, the validator should **not** vote cancel (same as GraphQL failure: `solver_lookup_failed=True` and skip in `forward.py`).

## Actual behavior

`find_solver_from_closure_event()` returns `(None, None)` with `solver_lookup_failed=False`. `issue_competitions()` treats this as “no identifiable solver” and votes cancel, potentially cancelling a valid bounty payout.

## Impact

On-chain bounty cancellation for legitimately solved issues on high-churn GitHub threads.

## Suggested fix

- Widen the closure timeline window (e.g. `last: 100`).
- If `closedAt` is set and completed close events exist but none match the current closure, return a lookup failure sentinel so `forward.py` skips cancel votes.

### LINKED ISSUE #1413: [Bug]: int/str github_id mismatch breaks duplicate penalty and bounty lookup

zeroknowledge0x:
## Summary

Normalizes `github_id` / `solver_github_id` to `str` at all boundaries to prevent int/str dict key mismatches, and returns a lookup-failure sentinel when a closed issue's authoritative close event is missing from the (possibly truncated) timeline.

Fixes #1413  
Fixes #1411

### Bug #1413 — int/str github_id mismatch

**Root cause:** `_solver_from_closed_event()` returns `databaseId` directly from GraphQL as `int`, while `registered_miners` keys come from `get_github_identity()` → `str(user_id)`. Even though `forward.py` line 132 wraps the lookup in `str()`, the inconsistency is fragile and breaks duplicate detection in `inspections.py`.

**Fix:**
- `github_api_tools._solver_from_closed_event` — normalize `databaseId` to `str` at the source; update return type from `tuple[Optional[int], ...]` to `tuple[Optional[str], ...]`
- `forward.py` — `str()` normalize `registered_miners` dict keys  
- `inspections.py` — `str()` normalize `github_id_to_uids` dict keys

### Bug #1411 — false cancel votes on truncated timelines

**Root cause:** `_ISSUE_CLOSURE_QUERY` fetches only the last 20 `CLOSED_EVENT` nodes. On high-churn issues the current close event can fall outside that window. `_select_current_close_event` then returns `None`, `find_solver_from_closure_event` returns `(None, None)` (not a failure sentinel), and `forward.py` votes cancel on a legitimately solved bounty.

**Fix:** In `find_solver_from_closure_event`, when `close_event is None` but `closedAt` is set and completed close events exist in the timeline (meaning the authoritative event was truncated), return `None` (lookup failure) instead of `(None, None)`. This causes `forward.py` to skip the issue rather than cancel it.

## Testing

- All 34 tests in `tests/utils/test_github_api_tools.py` pass (updated 6 assertions from int to str)
- All 4 tests in `tests/validator/test_issue_competitions_forward.py` pass
- All 561 tests in `tests/validator/` pass

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -426,7 +426,7 @@ def _select_current_close_event(issue_data: Dict[str, Any]) -> Optional[Dict[str
     return None
 
 
-def _solver_from_closed_event(repo: str, event: Dict[str, Any]) -> tuple[Optional[int], Optional[int]]:
+def _solver_from_closed_event(repo: str, event: Dict[str, Any]) -> tuple[Optional[str], Optional[int]]:
     target_repo = repo.lower()
     closer = event.get('closer') or {}
     if closer.get('__typename') != 'PullRequest':
@@ -445,12 +445,17 @@ def _solver_from_closed_event(repo: str, event: Dict[str, Any]) -> tuple[Optiona
         return None, None
 
     author = closer.get('author') or {}
-    return author.get('databaseId'), closer.get('number')
+    raw_id = author.get('databaseId')
+    # Normalize to str so dict lookups against registered_miners (whose keys
+    # come from get_github_identity -> str) always match regardless of the
+    # JSON int/str ambiguity.  See entrius/gittensor#1413.
+    github_id = str(raw_id) if raw_id is not None else None
+    return github_id, closer.get('number')
 
 
 def find_solver_from_closure_event(
     repo: str, issue_number: int, token: str
-) -> Optional[tuple[Optional[int], Optional[int]]]:
+) -> Optional[tuple[Optional[str], Optional[int]]]:
     """Resolve the issue solver from GitHub's authoritative current close event.
 
     Cross-reference and ``closingIssuesReferences`` entries are declarations made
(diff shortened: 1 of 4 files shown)


anderdc:
github_id is already coerced to a string at the source — get_github_identity returns str(user_id), and the bounty lookup already wraps the solver in str(solver_github_id). The int/str collision this targets can't occur on the real code path, so the normalization is redundant. Closing.

PR #1417, opened 2026-05-30
Normalize GitHub IDs for miner matching

### LINKED ISSUE #1413: [Bug]: int/str github_id mismatch breaks duplicate penalty and bounty lookup

hoangtrung1108:
## Summary
- Normalize miner GitHub IDs to strings when building duplicate-detection buckets.
- Normalize issue-bounty registered miner lookup keys so int and string IDs match consistently.
- Add regressions for mixed int/string duplicate detection and bounty solver lookup.

Fixes #1413

## Tests
- PYTHONPATH=C:\\Users\\49159\\money-engine\\test_stubs;C:\\Users\\49159\\money-engine\\bounties\\gittensor .\\.venv\\Scripts\\python -m pytest tests\\validator\\test_github_id_normalization.py tests\\validator\\test_issue_competitions_forward.py::test_int_github_id_miner_receives_solution_vote -q
- .\\.venv\\Scripts\\python -m ruff check gittensor\\validator\\oss_contributions\\inspections.py gittensor\\validator\\issue_competitions\\forward.py tests\\validator\\test_github_id_normalization.py tests\\validator\\test_issue_competitions_forward.py

Note: full dependency install on this Windows host failed because bittensor native dependencies require the MSVC linker (link.exe), so the focused tests were run with a minimal import stub for external bittensor modules.

### DIFF
--- a/gittensor/validator/issue_competitions/forward.py
+++ b/gittensor/validator/issue_competitions/forward.py
@@ -67,7 +67,7 @@ async def issue_competitions(
         # penalized in oss_contributions() before this pass and arrive here as
         # failed evaluations.
         registered_miners = {
-            eval.github_id: eval.hotkey
+            str(eval.github_id): eval.hotkey
             for eval in miner_evaluations.values()
             if eval.github_id and eval.github_id != '0' and eval.failed_reason is None
         }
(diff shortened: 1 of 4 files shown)


anderdc:
github_id is already coerced to a string at the source — get_github_identity returns str(user_id), so MinerEvaluation.github_id is never an int in production and the duplicate-detection / registered_miners dicts key consistently. The int/str collision this targets only arises from hand-constructing an evaluation with an int, not from the real code path. Closing.

PR #1432, opened 2026-06-02
fix: ignore commented inline-test markers

### LINKED ISSUE #1431: [Bug] Inline-test detection treats block comments and multiline strings as real tests
## Summary

`has_inline_tests()` in `gittensor/validator/utils/tree_sitter_scoring.py` can classify production Rust/D source files as test files when inline-test markers only appear inside comments or string literals.

The current implementation uses regex scanning over raw source text. After PR #1360, the regex no longer has the newline-heavy indentation issue, but it can still match markers that are not real syntax nodes.

## Reproduction

```python
from gittensor.validator.utils.tree_sitter_scoring import has_inline_tests

assert has_inline_tests(
    "fn prod() {}\n/*\n#[test]\nfn example() {}\n*/\n",
    "rs",
) is True

assert has_inline_tests(
    'fn prod() {}\nlet s = r#"\n#[test]\nfn example() {}\n"#;\n',
    "rs",
) is True

assert has_inline_tests(
    "int prod() { return 1; }\n/*\nunittest { assert(true); }\n*/\n",
    "d",
) is True
```

## Expected

Markers inside comments or string literals should not trigger inline-test detection.

## Actual

The examples above return `True`, so production files can be classified as test files.

## Impact

`calculate_token_score_from_file_changes()` demotes the whole file to `ScoringCategory.TEST` when `has_inline_tests()` returns `True`.

That applies `TEST_FILE_CONTRIBUTION_WEIGHT = 0.05`, causing valid production edits to be downweighted 20x and removed from source scoring.

In a minimal Rust example:

- `#[test]` only inside a block comment: score `0.006`, category `test`
- Same edit without that marker: score `0.12`, category `source`

## Suggested Fix

Keep the regex as a fast prefilter, but confirm matches using tree-sitter AST nodes before applying file-level inline-test demotion.

Possible AST checks:

- Rust: real `attribute_item` / `inner_attribute_item`
- Zig: real `TestDecl`
- D: real `unittest_declaration`

statxc:
## Summary

close #1431  

Fix `has_inline_tests()` so Rust/Zig/D inline-test detection no longer treats markers inside comments or string literals as real tests.

The existing regex remains as a fast prefilter, but matches are now verified against tree-sitter AST nodes:

- Rust: `attribute_item` / `inner_attribute_item`
- Zig: `TestDecl`
- D: `unittest_declaration`

This preserves the current file-level inline-test demotion behavior while avoiding false positives from docs, examples, comments, and raw strings.

## Problem

`has_inline_tests()` regex-scanned raw source text. A production Rust file containing `#[test]` only inside a block comment or raw multiline string was classified as an inline-test file.

That caused `calculate_token_score_from_file_changes()` to demote the whole file to `ScoringCategory.TEST` and apply `TEST_FILE_CONTRIBUTION_WEIGHT = 0.05`.

## Validation

- `uv run pytest -q tests/validator/test_inline_test_detection.py`
- `uv run pytest -q tests/validator/test_token_scoring_integration.py tests/validator/test_inline_test_detection.py`
- `uv run pytest tests/ -v`
- `uv run pre-commit run --all-files --show-diff-on-failure`
- `SKIP=pytest uv run pre-commit run --all-files --hook-stage pre-push`

Result: full test suite passed, 914 tests.

### DIFF
--- a/gittensor/validator/utils/tree_sitter_scoring.py
+++ b/gittensor/validator/utils/tree_sitter_scoring.py
@@ -1,7 +1,7 @@
 # The MIT License (MIT)
 # Copyright © 2025 Entrius
 from collections import Counter
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, Iterator, List, Optional, Tuple, Union
 
 import bittensor as bt
 from tree_sitter import Node, Parser, Tree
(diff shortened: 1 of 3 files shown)


statxc:
@anderdc I hope you are doing well. 
Could you explain about why this was closed.

PR #1435, opened 2026-06-02
fix(validator): pin scoring reference time to round start to prevent validator drift

### LINKED ISSUE #1427: [Bug]: Time-decay scoring uses wall clock causing validator drift
# [Bug]: Time-decay scoring uses wall clock causing validator drift

## Description

`calculate_time_decay` calls `datetime.now()` internally. Validators scoring the same PR in the same round at different times can get different multipliers and divergent weights.

## Steps to reproduce

1. Two validators score identical mirror data in one round near sigmoid midpoint.
2. Compare time-decay multipliers when runs are hours apart within the same step window.

## Expected behavior

All scoring in one forward step uses a single reference instant.

## Actual behavior

Each call uses wall clock at evaluation time.

nickmopen:
## Summary

`calculate_time_decay` called `datetime.now(timezone.utc)` on every invocation. Validators scoring the same PR in the same round but at slightly different wall-clock times (e.g. when a round spans several seconds) could compute different time-decay multipliers, causing weight divergence across the network.

This pins a single reference instant at the start of each forward step. Every `calculate_time_decay` call within that round uses the same timestamp, so all validators scoring identical mirror data produce identical multipliers.

**Changes:**

- `gittensor/validator/utils/datetime_utils.py`: Added `_scoring_reference_time` module variable and `set_scoring_reference_time()`. `calculate_time_decay` uses the pinned time when set, falls back to `datetime.now()` otherwise (CLI, tests, anything outside a scoring round — zero behaviour change).
- `gittensor/validator/forward.py`: Pins the reference time once before the scoring step with `try/finally` so it is always cleared even if scoring raises. Extracted scoring body to `_run_scoring_round()` to keep `forward()` readable.
- `tests/validator/utils/test_datetime_utils.py`: 8 new tests covering grace period, multiplier range, min-multiplier floor, determinism under a pinned anchor, and fallback to wall clock when cleared.

## Related Issues

Closes #1427

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -2,6 +2,7 @@
 # Copyright © 2025 Entrius
 
 import asyncio
+from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Dict, Optional, Set, Tuple
 
 import bittensor as bt
@@ -17,6 +18,7 @@ from gittensor.validator.utils.config import (
     VALIDATOR_STEPS_INTERVAL,
     VALIDATOR_WAIT,
 )
+from gittensor.validator.utils.datetime_utils import set_scoring_reference_time
 from gittensor.validator.utils.load_weights import (
     RepositoryConfig,
     load_master_repo_weights,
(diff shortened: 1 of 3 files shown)


anderdc:
Within-round wall-clock drift here is below the rounding threshold — time-decay is a 2-decimal sigmoid over days, and the per-repo lookback cutoff only shifts by the seconds between consecutive UIDs. Threading a reference-time param through six scoring functions to pin a difference that rounds away isn't worth the added surface in consensus-critical code. Closing.

PR #1439, opened 2026-06-04
perf(cli): read a single on-chain issue by ID instead of scanning all

plind-junior:
**Why this is necessary:** `fetch_issue_from_contract` is used to act on *one* issue by ID (e.g. `gitt issues submissions --id <N>`), but it reads **every** issue registered on the contract to find it:

```python
issues = read_issues_from_contract(...)          # one childstate RPC per issue, 1..next_issue_id
issue = next((i for i in issues if i.get('id') == issue_id), None)
```

That's O(N) substrate RPC round-trips to fetch O(1) data. `next_issue_id` only ever grows — completed and cancelled issues are never removed from storage — so every single-issue command gets progressively slower and pounds the node with N round-trips for data whose storage key it can compute directly. The Ink! 5 lazy mapping key for a specific `issue_id` is already computable inline, so a single read was always possible; the code just wasn't using it.

**The fix:** add `read_issue_from_contract`, which computes the one issue's lazy key and does a single `childstate_getStorage` call, and route `fetch_issue_from_contract` through it. O(N) → O(1).

The per-issue decode is factored into a shared `_read_one_issue_from_child_storage` helper, so the full-scan path (`read_issues_from_contract`, used by `issues list`) and the new single-read path share one decode implementation instead of duplicating it.

**Re-submission of #1390 (closed), addressing @anderdc's review:** rebased onto current `test`, and the new `read_issue_from_contract` now **propagates** RPC / connection failures instead of swallowing them with `return None`. `fetch_issue_from_contract` keeps the same try/except wrapper #1358 introduced for `read_issues_from_contract`, so a connection error surfaces as `ClickException('Error reading from contract: …')` — distinct from `ClickException('Issue ID N not found on-chain.')`. Two regression tests guard the distinction.

## Related Issues

Re-submission of #1390. Mirrors the read-failure distinction established by #1358 for `gitt issues list`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [x] Other (performance)

## Testing

- New `tests/cli/test_issue_single_read.py` (7 tests):
  - `_read_one_issue_from_child_storage` does exactly one RPC and returns the decoded dict.
  - Returns `None` when the storage entry is absent.
  - `fetch_issue_from_contract` uses the single read and **does not** call `read_issues_from_contract`.
  - Not-found path raises `ClickException('Issue ID N not found on-chain.')`.
  - Non-bountied status (`Completed`) raises.
  - **Read failure surfaces as `ClickException('Error reading from contract: …')`, not as "not found"** — guards the @anderdc distinction.
  - `ClickException` from the read path propagates verbatim (not re-wrapped).
- `uv run pytest` — **920 passed**.
- `uv run pyright` — 0 errors, 0 warnings.
- `uv run ruff check` / `ruff format --check` — clean.
- `uv run vulture` — clean.

## Checklist

- [x] Self-review completed

🤖Powered By Vouch!


anderdc:
This alters a gitt cli command, can you post screenshots of the change/improvement and that it works

plind-junior:
@anderdc — demo on finney (`next_issue_id = 34`), same contract, same issue ID:

| Branch | Wall clock | RPCs |
|---|---:|---:|
| upstream/test (OLD) | **26.130s** | 33 (full scan) |
| this PR (NEW) | **9.048s** | 1 (computed lazy key) |

**~2.9× speedup at N=33; gain scales linearly with `next_issue_id`**, which only grows — completed/cancelled issues are never removed from contract storage.

### Single-RPC fetch works (existing issue)
1-new-single-rpc

### Old code scanned every registered issue
2-old-full-scan

### Not-found path keeps the distinct error message
3-new-not-found

### Read-failure path stays distinct from not-found (your #1390 review concern)
4-new-read-error

Both error-distinction paths are guarded by the two regression tests added in this PR (`test_fetch_issue_surfaces_read_failure_distinctly_from_not_found`, `test_fetch_issue_passes_clickexception_through_unchanged`).

anderdc (approved):

PR #1442, opened 2026-06-04
fix(validator): pin scoring reference time for decay and PR lookback

claytonlin1110:
## Summary

Validator scoring used wall-clock time at each call site for (1) sigmoid time-decay multipliers and (2) per-repo PR lookback cutoffs sent to the mirror. Miners evaluated later in the same round could see a stricter lookback window and different decay than earlier UIDs; issue discovery already anchored one `now` per phase but OSS fetch did not.

This change captures a single `scoring_reference_time` at the start of each forward scoring step and threads it through `get_rewards`, `load_miner_prs`, `score_miner_prs`, `calculate_time_decay`, and `run_issue_discovery`. Optional parameters default to `None` so CLI/dev paths keep prior behavior when no anchor is passed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

Added:
- `tests/validator/utils/test_datetime_utils.py` — same reference → identical decay; advancing reference changes decay
- `tests/validator/oss_contributions/mirror/test_load.py` — two loads with the same anchor send identical `since_by_repo`

```bash
uv run pytest tests/validator/utils/test_datetime_utils.py \
  tests/validator/oss_contributions/mirror/test_load.py::TestLookbackWindow::test_scoring_reference_time_pins_since_by_repo -q

### DIFF
--- a/gittensor/validator/forward.py
+++ b/gittensor/validator/forward.py
@@ -2,6 +2,7 @@
 # Copyright © 2025 Entrius
 
 import asyncio
+from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Dict, Optional, Set, Tuple
 
 import bittensor as bt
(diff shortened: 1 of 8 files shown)


anderdc:
Within-round wall-clock drift here is below the rounding threshold — time-decay is a 2-decimal sigmoid over days, and the per-repo lookback cutoff only shifts by the seconds between consecutive UIDs. Threading a reference-time param through six scoring functions to pin a difference that rounds away isn't worth the added surface in consensus-critical code. Closing.

PR #1445, opened 2026-06-04
fix(validator): key GitHub identity pin to hotkey, not UID slot

### LINKED ISSUE #1444: [Bug] GitHub identity pin is keyed by reusable UID slot, not the hotkey — the same hotkey can switch GitHub accounts and bypass identity pinning
## Description

The validator's GitHub identity-pinning guarantee is documented as **permanent and per-hotkey**:

> This identity is permanently locked to the hotkey — the same hotkey cannot switch to a different GitHub account. […] To change GitHub accounts, the miner must deregister their UID and re-register with a **new hotkey**.

In practice the pin is **keyed on the UID slot, not on the hotkey**. The lock only survives while the hotkey continuously occupies its *original* UID. Because UID slots are recycled across (de)registrations, the pin is silently destroyed the moment any other hotkey takes that slot, after which the **same hotkey** can re-register and bind a **different** GitHub account — exactly what the spec says is impossible without a new hotkey. There is also no global `github_id` uniqueness, so two hotkeys can pin the same identity.

The pin is both **stored** and **looked up** by `uid`:

`gittensor/validator/pat_storage.py`:
```python
def save_pat(uid: int, hotkey: str, pat: str, github_id: str) -> None:
    """Upsert a PAT entry by UID. Creates the file if needed."""
    with _lock:
        entries = _read_file()
        entry = {'uid': uid, 'hotkey': hotkey, 'pat': pat,
                 'github_id': github_id, 'stored_at': datetime.now(timezone.utc).isoformat()}
        for i, existing in enumerate(entries):
            if existing.get('uid') == uid:      # <-- upsert keyed on UID only
                entries[i] = entry              #     overwrites the slot, destroying
                break                           #     the previous hotkey's pin
        else:
            entries.append(entry)
        _write_file(entries)

def get_pat_by_uid(uid: int) -> Optional[dict]:   # <-- only a by-UID index exists
    ...
```

`gittensor/validator/pat_handler.py:38`:
```python
def _github_identity_pin_error(uid: int, hotkey: str, github_id: Optional[str]) -> Optional[str]:
    existing = pat_storage.get_pat_by_uid(uid)               # <-- lookup keyed on UID slot
    if existing and existing.get('hotkey') == hotkey and existing.get('github_id'):
        if existing['github_id'] != github_id:
            return 'GitHub identity is locked for this hotkey. Deregister and re-register to change GitHub accounts.'
    return None
```

The check only fires when the **current occupant** of that exact UID slot is the *same* hotkey with a *different* `github_id`. There is no by-hotkey index and no `github_id`-uniqueness check anywhere in the repo, so once the slot is reused the pin is gone.

The downstream scoring path does not re-enforce the pin either: `inspections.py` sets `miner_eval.github_id = validation.github_id` from the **live** PAT and never compares it to the stored pin (the stored id is consulted only as a transient-failure fallback in `github_validation.py`), and the evaluation cache matches on the live `github_id` too.

## Steps to Reproduce

(Mirrors the storage/handler semantics exercised by `tests/validator/test_pat_handler.py`.)

1. Hotkey `H` is registered at UID `5` and broadcasts a PAT for GitHub id `"AAA"`.
   Storage → `{uid: 5, hotkey: H, github_id: "AAA"}` (pinned).
2. `H` deregisters. A throwaway hotkey `T` takes UID `5` and broadcasts GitHub `"CCC"`.
   `save_pat(5, T, …, "CCC")` upserts the slot → `{uid: 5, hotkey: T, github_id: "CCC"}`.
   The pin for `H` is now **erased** (this exact slot-overwrite is what `test_new_miner_on_uid_can_use_any_github` already asserts is allowed).
3. `H` re-registers (same hotkey) and lands on a fresh empty UID `8`, broadcasting a PAT for a **different** account `"BBB"`.
   `_github_identity_pin_error(8, H, "BBB")` → `get_pat_by_uid(8)` is `None` → returns `None` (no error). **Broadcast accepted.**

## Expected Behavior

Step 3 is rejected: `H` is permanently locked to `"AAA"` regardless of which UID slot it currently occupies, because the lock is documented as bound to the **hotkey**. The only way to use a different GitHub account is to register a **new hotkey** (paying for a distinct identity), per the spec. Additionally, a `github_id` already pinned to a live hotkey should not be silently re-pinnable by a different hotkey.

## Actual Behavior

Step 3 is accepted. The same hotkey `H` is now pinned to and scored under `"BBB"`, having switched GitHub accounts without a new hotkey — defeating the per-hotkey identity pin and weakening the Sybil/anti-gaming guarantee. The existing tests don't catch this because they only cover the case where the pin entry is still present at the same UID (`test_github_identity_change_rejected`) or where a *new* hotkey reuses a slot (`test_new_miner_on_uid_can_use_any_github`); none cover the original hotkey returning at a *different* UID after its old slot was overwritten.

## Environment

- OS: Linux
- Python version: 3.11
- Commit/Version: `test` branch @ `5ff21d3` (`chore(weights): tune awesome-claude reward labels (#1415)`)

### Suggested fix

Anchor the pin to the stable **hotkey** rather than the ephemeral UID slot.

1. Add a by-hotkey lookup in `pat_storage` that returns the most recent pin for a hotkey, and stop discarding a hotkey's historical pin when its UID slot is reused (keep the `(hotkey → github_id)` binding even after the UID is taken over):

```python
def get_pat_by_hotkey(hotkey: str) -> Optional[dict]:
    """Most recent stored entry for a hotkey, regardless of its current UID slot."""
    with _lock:
        latest = None
        for entry in _read_file():
            if entry.get('hotkey') == hotkey and entry.get('github_id'):
                if latest is None or entry.get('stored_at', '') >= latest.get('stored_at', ''):
                    latest = entry
        return latest
```

2. Use it in the pin check:

```python
def _github_identity_pin_error(uid: int, hotkey: str, github_id: Optional[str]) -> Optional[str]:
    existing = pat_storage.get_pat_by_hotkey(hotkey)
    if existing and existing.get('github_id') and existing['github_id'] != github_id:
        return 'GitHub identity is locked for this hotkey. Deregister and re-register to change GitHub accounts.'
    return None
```

3. (Completeness) Add a global `github_id`-uniqueness guard so a different hotkey cannot claim a `github_id` already pinned to a live hotkey, and have the scoring path (`inspections.py`) reject when `validation.github_id != stored_github_id` instead of silently adopting the live id.

### Token-score note

This is a structural keying fix, not a one-line tweak: it adds a by-hotkey storage index, changes the broadcast pin check, adjusts `save_pat`'s retention semantics, and adds regression tests for the "same hotkey re-registers on a new UID" path — a meaningful, multi-function change in well-weighted Python source.

### Severity

Anti-Sybil / identity integrity. The documented invariant *"permanently locked to the hotkey"* is not actually enforced; the pin degrades to "locked only while the hotkey keeps its original UID," which churn and deregistration routinely break. Filed publicly (matching the project's existing public anti-gaming bug reports, e.g. #1420 stale `is_admin`); happy to move it to a private Security Advisory if maintainers prefer.

philluiz2323:
## Summary

Fixes #1444. The validator's GitHub identity-pinning guarantee is documented as **permanent and per-hotkey**, but the pin was keyed on the **reusable UID slot** rather than the stable hotkey. The lock only held while a hotkey kept its original UID: once that slot was taken over by another hotkey (normal deregistration/churn), `save_pat`'s upsert-by-UID erased the displaced hotkey's record, and on re-registration onto a fresh UID the **same hotkey** could bind a **different** GitHub account — exactly what the spec says requires a new hotkey.

## Changes

- **`pat_storage.save_pat`** now upserts **by hotkey** (one record per hotkey). When a hotkey takes over a UID, the previous occupant's UID is **released** (`uid → None`) but its record — and therefore its identity pin — is **retained**.
- **`pat_storage.get_pat_by_hotkey`** (new) is the source of truth for pinning: the binding follows the stable hotkey, so a hotkey cannot shed its pin by cycling deregistration → re-registration.
- **`pat_storage.get_pat_by_uid`** still returns the current slot occupant (released records carry `uid=None`), so the scoring snapshot (`reward.py`) and `handle_pat_check` are unaffected.
- **`pat_handler._github_identity_pin_error`** now looks the pin up by hotkey (and ignores the `'0'` sentinel).

After the fix, a hotkey re-registering on a new UID stays locked to its original GitHub id; legitimate flows (PAT rotation with the same identity, a genuinely new hotkey on a recycled UID) still pass.

## Tests

- `test_pat_storage.py`: rewrote `test_save_upsert_replaces_hotkey_on_uid` (it encoded the buggy overwrite) into `test_save_releases_uid_but_retains_displaced_pin`; added a `TestGetPatByHotkey` class incl. pin-survives-UID-reuse-and-reregistration.
- `test_pat_handler.py`: added `test_identity_pin_survives_uid_slot_reuse_and_reregistration` (regression). Existing pin/rotation/new-miner/concurrent-recheck tests still hold.

## Notes

Scoped intentionally to the per-hotkey pin (the clear spec violation). A broadcast-time global `github_id`-uniqueness lock was deliberately left out: duplicate-account detection already zeroes miners sharing a GitHub account at scoring time, and such a lock would risk permanently locking out a user who lost access to an old hotkey.

### DIFF
--- a/gittensor/validator/pat_handler.py
+++ b/gittensor/validator/pat_handler.py
@@ -35,11 +35,14 @@ def _get_hotkey(synapse: bt.Synapse) -> str:
     return synapse.dendrite.hotkey
 
 
-def _github_identity_pin_error(uid: int, hotkey: str, github_id: Optional[str]) -> Optional[str]:
-    existing = pat_storage.get_pat_by_uid(uid)
-    if existing and existing.get('hotkey') == hotkey and existing.get('github_id'):
-        if existing['github_id'] != github_id:
-            return 'GitHub identity is locked for this hotkey. Deregister and re-register to change GitHub accounts.'
+def _github_identity_pin_error(hotkey: str, github_id: Optional[str]) -> Optional[str]:
+    # Pin is keyed on the stable hotkey, not the reusable UID slot, so a hotkey
+    # stays locked to its original GitHub account even after its old UID is
+    # taken over and it re-registers onto a different UID.
+    existing = pat_storage.get_pat_by_hotkey(hotkey)
+    pinned_github_id = existing.get('github_id') if existing else None
+    if pinned_github_id and pinned_github_id != '0' and pinned_github_id != github_id:
+        return 'GitHub identity is locked for this hotkey. Deregister and re-register to change GitHub accounts.'
     return None
 
 
@@ -71,7 +74,7 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
         return _reject(error)
 
     # 3. Enforce GitHub identity pinning — same hotkey cannot switch GitHub accounts
-    identity_error = _github_identity_pin_error(uid, hotkey, github_id)
+    identity_error = _github_identity_pin_error(hotkey, github_id)
     if identity_error:
         return _reject(identity_error)
 
@@ -80,7 +83,7 @@ async def handle_pat_broadcast(validator: 'Validator', synapse: PatBroadcastSyna
     if test_error:
         return _reject(f'PAT test query failed: {test_error}')
 
-    identity_error = _github_identity_pin_error(uid, hotkey, github_id)
+    identity_error = _github_identity_pin_error(hotkey, github_id)
     if identity_error:
         return _reject(identity_error)
 
--- a/gittensor/validator/pat_storage.py
+++ b/gittensor/validator/pat_storage.py
@@ -34,7 +34,14 @@ def load_all_pats() -> list[dict]:
 
 
 def save_pat(uid: int, hotkey: str, pat: str, github_id: str) -> None:
-    """Upsert a PAT entry by UID. Creates the file if needed."""
+    """Upsert a PAT entry, keyed by HOTKEY (one record per hotkey).
+
+    The record is keyed by the stable hotkey rather than the reusable UID slot
+    so a hotkey's GitHub identity pin survives UID churn. When this hotkey takes
+    over a UID that another hotkey used to occupy, the previous occupant's UID is
+    released (set to None) but its record — and therefore its identity pin — is
+    retained, so a displaced hotkey stays locked to its original GitHub account.
+    """
     with _lock:
         entries = _read_file()
 
@@ -46,8 +53,14 @@ def save_pat(uid: int, hotkey: str, pat: str, github_id: str) -> None:
             'stored_at': datetime.now(timezone.utc).isoformat(),
         }
 
+        # The UID slot now belongs to `hotkey`; release it from any other hotkey
+        # that previously occupied it, but keep that hotkey's identity pin.
+        for existing in entries:
+            if existing.get('uid') == uid and existing.get('hotkey') != hotkey:
+                existing['uid'] = None
+
         for i, existing in enumerate(entries):
-            if existing.get('uid') == uid:
+            if existing.get('hotkey') == hotkey:
                 entries[i] = entry
                 break
         else:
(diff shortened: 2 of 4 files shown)


anderdc:
Triggering this bypass requires deregistering the hotkey and re-registering it — a full fresh registration, which is the same cost as the spec's sanctioned "new hotkey" path. Credibility and scoring are derived from the GitHub account's real merged-PR history, not the hotkey, so reusing the same hotkey carries no reputation forward; the swapped-in account is scored fresh either way. Net attacker yield is ~zero, while retaining released (uid=None) records grows the PAT store unboundedly under churn. The documented "per-hotkey" wording is violated only in letter, not in the anti-Sybil economics. Closing.

PR #1436, opened 2026-06-02
refactor: hoist gittensor.constants import out of get_contract_address

ebios-star:
## Summary

The `CONTRACT_ADDRESS` import in [`gittensor/utils/utils.py`](gittensor/utils/utils.py) was function-local inside `get_contract_address`, which is the only use of the name in this module. Moving it to module-level:

- matches the **imports-at-top** convention used everywhere else in `gittensor/`
- removes per-call import lookup overhead (`get_contract_address` is called on every validator forward pass for the issue-bounty contract)

Verified **no circular-import risk**: `gittensor.constants` only imports stdlib (`re`, `typing`) — there is no path back into `gittensor.utils.utils`.

```diff
 import os

+from gittensor.constants import CONTRACT_ADDRESS
+

 def backoff_seconds(attempt: int, base: int = 5, cap: int = 30) -> int:
     return min(base * (2**attempt), cap)


 def get_contract_address() -> str:
     """..."""
-    from gittensor.constants import CONTRACT_ADDRESS
-
     return os.environ.get('CONTRACT_ADDRESS') or CONTRACT_ADDRESS
```

Net: **+2 / -2 lines**, single file. No behaviour change.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- `pytest tests/` — all 913 tests pass.

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

PR #1457, opened 2026-06-05
refactor: hoist 'from math import prod' out of calculate_open_pr_collateral_score

ebios-star:
## Summary

The `math.prod` import in [`gittensor/validator/oss_contributions/scoring.py`](gittensor/validator/oss_contributions/scoring.py#L250) is function-local inside `calculate_open_pr_collateral_score`, which is the only use of the name in this module.

[`gittensor/classes.py:6`](gittensor/classes.py#L6) already imports `from math import prod` at module-level for the same purpose (per-PR score multiplier composition), so this brings `scoring.py` in line with the established in-codebase pattern.

`calculate_open_pr_collateral_score` runs for **every open PR in every scoring round** — the per-call import lookup is on a validator hot path. Hoisting removes that overhead and matches the imports-at-top convention used everywhere else in the module.

Verified **no circular-import risk**: `math` is stdlib.

```diff
+from math import prod
 from typing import TYPE_CHECKING, Dict, List, Optional, Set

 import bittensor as bt
 ...

 def calculate_open_pr_collateral_score(pr: 'ScoredPR', scoring: ResolvedScoring) -> float:
     """..."""
-    from math import prod
-
     multipliers = {
```

Net: **+1 / -2 lines**, single file. No behaviour change.

Same shape as just-merged #1436 (hoist `gittensor.constants` import out of `get_contract_address`).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- `pytest tests/` — all 913 tests pass.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/oss_contributions/scoring.py
+++ b/gittensor/validator/oss_contributions/scoring.py
@@ -1,6 +1,7 @@
 # The MIT License (MIT)
 # Copyright © 2025 Entrius
 
+from math import prod
 from typing import TYPE_CHECKING, Dict, List, Optional, Set
 
 import bittensor as bt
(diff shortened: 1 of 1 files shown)


ebios-star:
@anderdc whenever you have a moment, this hoists a function-local math.prod import to module-level on the validator collateral-scoring hot path. Happy to revise scope if you'd prefer. Thanks!

PR #1472, opened 2026-06-12
fix(cli): report vote list validator read failures as read_failed

### LINKED ISSUE #1471: [Bug] `gitt vote list --json` reports an empty validator whitelist when the contract read fails
## Summary

`gitt vote list --json` can report a successful empty validator whitelist when the underlying `get_validators` contract read fails.

`IssueCompetitionContractClient.get_validators()` collapses **both** failure paths into an empty list — a failed raw contract read (`None`) and any exception raised while reading or decoding:

```python
try:
    response = self._raw_contract_read('get_validators')
    if response is None:
        return []

    return self._decode_validator_list(response)
except Exception as e:
    bt.logging.error(f'Error fetching validators: {e}')
    return []
```

`gitt vote list --json` then emits:

```json
{
  "success": true,
  "validators": [],
  "count": 0,
  "consensus_threshold": 0
}
```

This makes a failed contract/RPC read indistinguishable from a genuinely empty validator whitelist.

## Motivation

JSON consumers use `gitt vote list --json` for monitoring validator whitelist state. A transient contract read failure should not look like a valid “no validators whitelisted” result.

This is the same correctness pattern fixed for `gitt issues list` in #1357 / #1358 (merged): read failures should be distinguishable from valid empty results.

`vote list` also already has accepted JSON-consistency fixes, including #753 / #828 (merged) for the empty-whitelist threshold. This issue preserves that real empty-whitelist behavior, but separates it from read failure.

Related in-flight work: #1449 / #1450 apply the same `read_failed` pattern to `pending-harvest`. This issue covers the remaining `vote list` read path and does not overlap that PR.

## Steps to Reproduce

Real CLI reproduction (no mocks). Point `vote list` at a contract that is deployed but is not the issues contract — the contract-existence check passes, but the `get_validators` read itself fails (reverts), and the failure is swallowed:

```bash
gitt vote list --json \
  --network finney \
  --contract 5DU266Ac3kVL6b3gGKNVqQ467WAq4JpucRAwxU1z9B98Bkav
```

Observed output (finney, 2026-06-12), exit code `0`:

```json
{
  "success": true,
  "validators": [],
  "count": 0,
  "consensus_threshold": 0
}
```

`5DU266…` is an unrelated contract deployed on finney (any deployed non-issues contract behaves the same — e.g. a stale configured address left over from a previous deployment). Transient RPC failures during the read hit the same swallow path.

Control — the same command against the real issues contract reads correctly:

```bash
gitt vote list --json --network finney \
  --contract 5FWNdk8YNtNcHKrAx2krqenFrFAZG7vmsd2XN2isJSew3MrD
# exit 0 — {"success": true, "validators": ["5Gjr7Vu..."], "count": 1, "consensus_threshold": 1}
```

Note the inconsistency with existing guards: a plain non-contract address is already rejected hard (`No contract found at …`, exit 1), but a failed read against a deployed contract silently reports an empty whitelist with `success: true`.

Deterministic variant for regression tests: mock `IssueCompetitionContractClient._raw_contract_read("get_validators")` to return `None` (or to raise) — both failure paths currently collapse to `[]` and produce the same false-success payload.

A real empty whitelist should still succeed, but a failed read should not.

## Expected Behavior

When the validator whitelist read fails:

- `--json` mode should emit `success: false`
- `error.type` should be `read_failed`
- the command should exit non-zero
- it should not emit `validators: []` as if the whitelist was successfully read

A genuine empty whitelist should continue to emit:

```json
{
  "success": true,
  "validators": [],
  "count": 0,
  "consensus_threshold": 0
}
```

## Suggested Fix

Make `get_validators()` distinguish failed reads from successful empty Vecs:

- If `_raw_contract_read("get_validators")` returns `None`, raise a distinct read error (e.g. `ContractReadError`) instead of returning `[]`.
- Stop swallowing exceptions in the broad `except Exception` block — re-raise (or wrap in the same read error) instead of logging and returning `[]`, so RPC/decode failures are not silently converted into an empty whitelist either.
- Keep returning `[]` when the contract successfully returns an encoded empty validator Vec.
- In `vote_list_validators`, catch the read error before the generic exception handler and route it through `handle_exception(..., error_type="read_failed")`, exiting non-zero.

In the current tree, the only observed caller of `get_validators()` is `vote_list_validators` (`gittensor/cli/issue_commands/vote.py`), so raising does not change validator-runtime behavior. If keeping the soft `[]` return for potential external callers is preferred, an opt-in strict mode (e.g. `get_validators(strict=True)` used by the CLI) achieves the same outcome.

## Suggested Tests

Add regression coverage for:

- `_raw_contract_read()` returns `None`
  - `gitt vote list --json` exits non-zero
  - payload has `success: false`
  - payload has `error.type == "read_failed"`
  - payload does not contain `validators: []`
- `_raw_contract_read()` raises (e.g. connection refused)
  - same `read_failed` envelope, non-zero exit
- successful encoded empty validator Vec
  - still exits `0`
  - still emits `validators: []`, `count: 0`, `consensus_threshold: 0`

carlos4s:
## Summary

Fixes #1471 

Fix `gitt vote list --json` so validator whitelist read failures no longer look like a successful empty whitelist.

Changes:
- Add `ContractReadError` for trustworthy contract-read failure reporting.
- Make `IssueCompetitionContractClient.get_validators()` raise on failed raw reads, read exceptions, and malformed validator Vec payloads.
- Preserve successful empty whitelist behavior for a valid encoded empty Vec.
- Map validator whitelist read failures in `vote list` to `error.type == "read_failed"`.
- Add CLI regression tests for raw read `None`, read exception, malformed payload, and real empty whitelist.

## Verification (live `finney` — real data, no mocks)

Run against the live `finney` network using the exact contract addresses from #1471:

| Contract | Meaning |
|---|---|
| `5DU266Ac3kVL6b3gGKNVqQ467WAq4JpucRAwxU1z9B98Bkav` | deployed non-issues contract — the `get_validators` read fails |
| `5FWNdk8YNtNcHKrAx2krqenFrFAZG7vmsd2XN2isJSew3MrD` | real issues contract — returns the live validator whitelist |

### Before / After — failed read

Same failed-read contract, same command. Before this PR a failed read was reported as a successful empty whitelist (`success: true`, exit 0); after it is correctly surfaced as `read_failed` (exit 1).

```bash
gitt vote list --json --network finney \
  --contract 5DU266Ac3kVL6b3gGKNVqQ467WAq4JpucRAwxU1z9B98Bkav
```

before/after — failed read now returns read_failed

### Regression check — real whitelist still reads correctly

A genuine read against the real issues contract is unchanged: it still succeeds and returns the live validator (`success: true`, exit 0).

```bash
gitt vote list --json --network finney \
  --contract 5FWNdk8YNtNcHKrAx2krqenFrFAZG7vmsd2XN2isJSew3MrD
```

real whitelist still succeeds

## Testing

- `uv run --extra dev pytest`
- `uv run --extra dev ruff check .`
- `uv run --extra dev ruff format --check .`
- `uv run --extra dev pyright`


anderdc:
please provide screenshots of the cli command working

carlos4s:
@anderdc I've attached cli command working screenshots in PR description. Please review again.

anderdc (approved):

PR #1474, opened 2026-06-12
feat: add --status filter to gitt issues list

### LINKED ISSUE #1473: [Feature] gitt issues list can filter by repo but not by lifecycle state — add a `--status` filter (registered/active/completed/cancelled)
## Summary

`gitt issues list` supports filtering by repository via `--repo` (#892, shipped in #910), but there is no way to filter by issue lifecycle state. Add a client-side `--status` filter:

​```bash
gitt issues list --status active
gitt issues list --status registered
gitt issues list --status completed
gitt issues list --status cancelled
​```

Supported values should match the four statuses the CLI already renders in the Status column: `Registered`, `Active`, `Completed`, `Cancelled`.

## Motivation

The most common read-side workflow is finding actionable issues — in practice, `Active` bounties. Issues are never removed from contract storage, so completed and cancelled entries accumulate permanently and the signal-to-noise of an unfiltered listing only degrades as the subnet grows.

Today the workarounds are the same shell gymnastics #892 described for repos:

​```bash
# eyeball the colored Status column across every row
$ gitt issues list

# rebuild the filter in jq, losing the table rendering
$ gitt issues list --json | jq '.issues[] | select(.status == "Active")'
​```

## Proposed behavior

- `--status` as a case-insensitive `click.Choice` over the four statuses, matching the existing `--network` option style.
- Applies identically to the table output and `--json` output.
- Composes with `--repo`; `--id` takes precedence (same semantics the `--repo` filter already has).
- No flag provided → behavior unchanged.

A client-side filter applied after the contract read is sufficient — the same approach as the existing `--repo` filter in `issues_list` (`gittensor/cli/issue_commands/view.py`).

taillred:
## Summary

Adds a `--status` filter to `gitt issues list` so users can list only issues in a specific lifecycle state:

​```bash
gitt issues list --status active
gitt issues list --status registered
gitt issues list --status completed
gitt issues list --status cancelled
​```

Closes #1473

## Changes

- New `STATUS_CHOICE` click constant in `issue_commands/helpers.py`, alongside the existing `NETWORK_CHOICE`, covering the four on-chain statuses.
- New `--status` option on `issues list`, mirroring the existing `--repo` filter (#910): a client-side, case-insensitive filter applied identically to both the table view and `--json` output, with `--id` taking precedence.
- Behavior is unchanged when the flag is omitted; unknown values are rejected at parse time by `click.Choice`.
- Two tests in `tests/cli/test_issues_list_json.py`: JSON filtering with case-insensitive input, and parse-time rejection of invalid values (no contract read).

## Testing

- `uv run pytest tests/` — 919 passed
- `uv run pre-commit run --all-files` (ruff lint/format) and pre-push hooks (pyright, vulture) — all green
- Manually exercised table + JSON paths, `--status` combined with `--repo`, `--id` precedence, and no-match output via `CliRunner`

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -60,6 +60,7 @@ err_console = Console(stderr=True)
 
 CommandFunc = TypeVar('CommandFunc', bound=Callable[..., Any])
 NETWORK_CHOICE = click.Choice(['finney', 'test', 'local'], case_sensitive=False)
+STATUS_CHOICE = click.Choice(['registered', 'active', 'completed', 'cancelled'], case_sensitive=False)
 
 
 def apply_click_options(*decorators: Callable[[CommandFunc], CommandFunc]) -> Callable[[CommandFunc], CommandFunc]:
--- a/gittensor/cli/issue_commands/view.py
+++ b/gittensor/cli/issue_commands/view.py
@@ -21,6 +21,7 @@ from gittensor.cli.json_output import emit_json
 
 from .help import StyledCommand
 from .helpers import (
+    STATUS_CHOICE,
     _read_contract_packed_storage,
     _read_issues_from_child_storage,
     _resolve_contract_and_network,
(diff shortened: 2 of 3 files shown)


anderdc:
Not really looking for more PRs related to issue bounties

PR #1475, opened 2026-06-14
fix(validator): close GitHub API sessions to stop connection-pool leaks

jaso0n0818:
## Summary

`get_github_identity`, `execute_graphql_query`, and `check_github_issue_closed` in `gittensor/utils/github_api_tools.py` each open a `requests.Session` via `get_session()` but never close it. Every PAT validation, GraphQL query, and issue-closure check therefore leaks the session's underlying `urllib3` connection pool (open sockets) for the life of the process. On a validator these run continuously, once per miner, so the leaked pools accumulate over time.

This wraps each session in `try/finally` (matching the `try/except` that already surrounds `check_github_issue_closed`) so the session is closed on every return and exception path. No control flow or return values change — only session lifetime.

## Related Issues

No issue filed; this is a self-contained resource-leak fix in the validator's GitHub API helpers, in the same vein as the per-PR content cleanup in #1456.

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing

- Added a `close()` shim to the `_ForwardingSession` proxy in `tests/conftest.py` so it mirrors `requests.Session` now that callers release the session. (Not a test-only change — it accompanies the code fix.)
- Full gate run locally with `uv` (Python 3.12):

```
uv run pre-commit run --all-files                        # ruff lint + format: Passed
SKIP=pytest uv run pre-commit run --all-files --hook-stage pre-push  # pyright + vulture: Passed
uv run pytest tests/ -q                                  # 917 passed
```

No CLI output is affected (change is in `gittensor/utils/**`, not `gittensor/cli/**`), so no terminal evidence applies.

## Checklist

- [x] Code follows repository style (ruff lint + format clean)
- [x] Self-reviewed the diff (the large line count is re-indentation from the `try/finally` wrap, not logic changes)
- [x] Tests pass (`uv run pytest tests/` — 917 passed) and pyright/vulture clean
- [x] No unrelated changes

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -108,56 +108,58 @@ def get_github_identity(token: str) -> GitHubIdentityResult:
         return GitHubIdentityResult(None, GitHubIdentityStatus.INVALID_AUTH)
 
     session = get_session(token)
+    try:
+        # Retry logic for timeout issues
+        for attempt in range(6):
+            try:
+                response = session.get(f'{BASE_GITHUB_API_URL}/user', timeout=GITHUB_HTTP_TIMEOUT_SECONDS)
+                if response.status_code == 200:
+                    try:
+                        user_data: Dict[str, Any] = response.json()
+                    except Exception as e:
+                        bt.logging.warning(f'Failed to parse GitHub /user JSON response: {e}')
+                        if attempt < 5:
+                            time.sleep(2)
+                            continue
+                        return GitHubIdentityResult(None, GitHubIdentityStatus.TRANSIENT_FAILURE)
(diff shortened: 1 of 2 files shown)


anderdc:
These sessions are function-local with no long-lived references — they go out of scope at return, GC reclaims them, and urllib3's connection-pool finalizers close the sockets at that point. There's no unbounded accumulation over the process lifetime: each helper makes one request and the session is collected within a GC cycle, so the only effect of the explicit close is cleanup determinism and quieter ResourceWarning logs, not an actual socket leak. Closing.

PR #1482, opened 2026-06-16
feat(cli): add `gitt miner ensure` and retry post broadcasts so PAT coverage self-heals

### LINKED ISSUE #1481: [Bug] `gitt miner post` coverage silently and permanently erodes after validator restarts — valid miners are de-scored with no signal and no recovery
> Verified against `origin/test` @ `1c813f5` (2026-06-15). The two reproductions below run against the real code; the live evidence is publicly verifiable in the validator telemetry at `wandb.ai/entrius-gittensor/gittensor-validators`.

## Summary

The documented miner workflow is "run `gitt miner post` once; validators store your PAT and score your merged PRs." In practice a miner's PAT coverage **monotonically decays**: each time a validator loses its on-disk PAT store, that validator scores the miner **0** every round thereafter, and **nothing restores it** — `gitt miner post` is a one-shot best-effort broadcast, and miners run no axon, so there is no re-broadcast and no way for a validator to re-request. The miner gets **no error, no log, no signal**; coverage is only re-established by manually re-running `gitt miner post`.

Net effect: a valid, registered, actively-contributing miner is silently and indefinitely de-scored by a growing subset of validators. This is live and systemic (receipts below), not theoretical — it has dropped my own miner (UID 29) out of eligibility twice on a multi-week cadence.

## Mechanism (verbatim @ `origin/test`)

**A validator loses its store** (`data/miner_pats.json`, `gittensor/validator/pat_storage.py:18`) on any of: a restart without a persistent `./data` (`ensure_pats_file()` then starts empty, `neurons/validator.py:62`); a crash/restart loop; or a single failed/corrupt read — `_read_file()` (`pat_storage.py:68`) catches `(json.JSONDecodeError, OSError)` → returns `[]`, and the next `save_pat()` upserts one entry into that `[]` and atomically overwrites, erasing all others. (On the standard `docker-compose.vali.yml` — `./data:/app/data`, atomic writes — the store *does* survive a clean restart, so this is **not** about per-operator volume config.)

**Downstream**, `reward.py:100` snapshots `load_all_pats()` per round; any miner without a stored PAT is logged `No stored PAT for miner {uid} — miner must run gitt miner post` (`oss_contributions/inspections.py:134`) and scored 0.

**Nothing recovers it:** `gitt miner post` issues a single `dendrite(..., timeout=30.0)` broadcast with no retry/resend (`cli/miner_commands/post.py:133-142`) then exits; miners serve no axon (README: "No miner neuron required"), so a validator cannot re-request; and no miner-facing signal exists (the guide implies one post suffices).

## Proof 1 — `gitt miner post` silently reports success and never retries a missed validator

Runs the real `gitt miner post` (Click `CliRunner`); only the network/identity boundaries are stubbed — the broadcast-result handling, success flag, exit code, and retry behavior are the real code. Validator UID 30 does not respond:

```text
REPRODUCTION: `gitt miner post` — silent partial success + no retry
  3 validators; validator UID 30 did NOT respond (status 408).
  command exit code        : 0
  reported success         : True
  accepted / total         : 2 / 3
  no_response (silent miss): 1
  broadcast attempts made  : 1
PROVEN: exited 0 and reported success=True while UID 30 got nothing, with exactly
ONE broadcast (no retry). A transient failure to any validator is a silent, un-
retried, un-surfaced coverage gap — the miner is never told they are uncovered.
```

So a brief blip during the one post (or a validator that is mid-restart) permanently drops the miner from that validator until a future manual re-post — and `post` reports success anyway.

## Proof 2 — a single failed read of `miner_pats.json` erases every stored PAT

Runs the real `pat_storage.save_pat`/`_read_file`; only `PATS_FILE` is redirected to a scratch path:

```text
[setup]   3 miners stored normally   -> file UIDs = [10, 20, 30]
[control] 4th miner, file readable   -> file UIDs = [10, 20, 30, 40]   (correct)
[trigger] file unreadable (corrupt / I/O error) before the next broadcast
[result]  one miner (UID 99) broadcasts -> file UIDs = [99]
PROVEN: PATs for UIDs [10,20,30,40] silently and permanently erased by one unrelated
broadcast after a single failed read. Those miners are scored 0 until each re-posts.
```

(Both scripts are self-contained and deterministic; available on request.)

## Proof 3 — it is happening live (publicly verifiable)

From `wandb.ai/entrius-gittensor/gittensor-validators` (anonymously readable; console telemetry; display names are `vali-{uid}-{version}`):

- **`gittensor-181`** (run `bbd93w0w`) restarted **at the release** — new run created `2026-06-15T23:51Z`, ~4 min after the `release-20260615-234738` tag — and **kept** its PATs (`PAT check result — UID: 29: valid`). Confirms the store persists on a well-configured validator across an upgrade-restart.
- **`vali-116`** (run `u1ropz51`, running since `2026-06-13`) was **missing** that same miner's PAT (4/7 coverage on my `gitt miner check`) despite no restart at this release — i.e. lost on an earlier restart, never recovered. In one ~110-min window it logged a *second* miner cycling through loss → re-post (`UID: 32: no PAT stored` → `PAT broadcast accepted — UID: 32`), so this is systemic across miners.
- **`rt21-64`** (UID 64) is in a restart/crash loop (10+ runs over Jun 8–12) and is unreachable for re-broadcast.

## Impact

A transient/operational event (a validator restart) produces a **lasting** 0-score for a valid miner across every validator that lost its PAT; those validators set weights from the zeros, so it reduces real emission. Coverage only erodes (no mechanism ever restores it), the miner is never told, and it recurs as validators restart (upgrades/crashes).

## Relationship to prior reports

- **#781 / #1079** (corrupt `miner_pats.json` → empty → overwrite) and their fixes **#829 / #1081** addressed only loss-vector 3 (the read-wipe). This issue is scoped to the **missing recovery path**, which is independent of *how* the store was lost and has not been filed.
- **#931 / #932 / #1107** (merged) established that a *transient* condition must not cause a *permanent wrong outcome* — there, a transient GitHub `/user` failure is treated as **inconclusive** (cache fallback) rather than a hard failure. A validator restart is the same shape of event; today it yields a permanent silent de-score. The fix applies the same principle to PAT coverage.

## Proposed fix

Recovery must be miner-initiated (validators cannot pull — no miner axon). Make the reference tooling keep coverage *asserted* instead of one-shot:

- Add a re-assert mode to `gitt miner` (e.g. `gitt miner post --watch` / `gitt miner ensure`) that, on an interval, runs the existing **check** and re-broadcasts **only** to validators reporting `has_pat=False` / `pat_valid != True`. Cheap (check transmits no PAT), non-spammy (posts only where missing), and makes the existing "miner must re-post" remedy reliable instead of manual-and-silent.
- Surgical first step if a daemon is out of scope: have `gitt miner post` **retry** validators that return no response, and surface partial coverage as a non-zero exit / explicit warning so it is not silent (Proof 1).
- Document that PAT coverage can be lost on validator restarts and should be re-asserted.

Separate, optional hardening for loss-vector 3 (Proof 2): make `save_pat()` fail closed when the prior read could not be parsed instead of overwriting (as #829 / #1081 proposed).

## Suggested tests

- `gitt miner post` retries a non-responding validator and reports full coverage only when actually achieved; persistent partial coverage → non-zero exit / warning (Proof 1 becomes a regression test).
- The re-assert path posts to exactly the validators whose check returns missing/invalid and skips the valid ones.
- `save_pat()` preserves the existing entries (raises) when `_read_file()` cannot parse the file (Proof 2 becomes a regression test).

## Footer — verification provenance

- Live on `origin/test` @ `1c813f5` (2026-06-15). Code: `pat_storage.py:18/23/36/68/78`; `neurons/validator.py:62`; `oss_contributions/reward.py:100`, `inspections.py:134`; `cli/miner_commands/post.py:133-142`; README miner section. Deploy: `docker-compose.vali.yml` (`./data:/app/data`).
- Telemetry (public, verifiable): `wandb.ai/entrius-gittensor/gittensor-validators` — `gittensor-181`/`bbd93w0w` (restarted at release, kept PATs), `vali-116`/`u1ropz51` (lost a PAT, no recovery; UID 32 systemic), `rt21-64` (crash loop).
- Both reproductions run against current `origin/test`. Fix is Python-side (CLI / validator); no contract changes.

JSONbored:
Fixes #1481

> Targets `test`. Verified against `origin/test` @ `1c813f5` (2026-06-15); full suite green (`uv run pytest tests/` → 945 passed), ruff lint + format clean.

## Summary

A miner's PAT coverage silently and permanently erodes. `gitt miner post` is a one-shot, best-effort broadcast: a validator briefly unreachable during it is silently dropped — no retry, and the command still reports success. And once any validator loses its stored PAT (e.g. it restarts without persistent `./data`), there is **no mechanism to restore coverage** — the miner is scored 0 by that validator every round until they remember to manually re-post. (Root cause, live evidence, and two runnable reproductions in #1481.)

This makes coverage **reliable** and **self-healing**, miner-side only:

- **`gitt miner post` now retries** validators that don't respond (`--retries`, default 2), so a transient blip during the broadcast isn't a silent, permanent coverage gap. An explicit accept/reject is final; only no-responses are retried.
- **New `gitt miner ensure`** — probes current coverage, then re-broadcasts the PAT **only** to validators that are missing a valid one (**no PAT is sent to validators that already have it**). It's cheap and non-spammy, safe to run on a schedule (cron), and exits non-zero if any reachable validator is still uncovered. Running it after a validator restart restores coverage automatically instead of silently de-scoring the miner. `--watch SECONDS` runs it as a self-healing loop (re-syncing the metagraph each round to pick up validators that (re)join), so coverage recovers without external cron.
- Shared `_probe_pat` / `_broadcast_pat_with_retry` helpers; `post` is refactored onto the latter (net −39 lines there).

This aligns with the existing "miner must re-post" remedy (surfaced in `inspections.py` when a validator has no stored PAT) by making it reliable and automatable, and with the transient-failure-resilience already adopted in #931 / #932 / #1107 (a transient/operational event must not cause a permanent wrong outcome).

## Behavior

`gitt miner ensure --json` (validator UID 1 was missing; UIDs 0 and 2 already had a valid PAT):

```json
{ "success": true, "total_validators": 3, "already_valid": 2,
  "reposted": 1, "now_valid": 3, "still_missing": [], "results": [ ... ], "skipped": [] }
```
Only the missing validator is re-sent the PAT; the two that already had it are untouched. Exits non-zero (for cron/alerting) if any reachable validator stays uncovered.

## Tests

`tests/cli/test_miner_commands.py`:
- `_broadcast_pat_with_retry`: retries **only** no-response validators; an explicit rejection is final (not retried); a persistent no-response is surfaced in the result, not silently dropped.
- `gitt miner ensure`: re-broadcasts to exactly the validators missing a PAT and never to the ones that have it; exits non-zero when a validator stays uncovered.
- `--watch`: loops the coverage cycle, re-syncs the metagraph between rounds, and exits cleanly on `KeyboardInterrupt`.
- Updated the existing `post` JSON-envelope test so its dendrite mock is axon-aware (a no-response validator stays no-response across retries — previously the mock returned fixed responses regardless of axons).

```
uv run pytest tests/      -> 945 passed
ruff check / ruff format  -> clean   (vulture clean on changed files)
```

## Notes

- Targets `test`. Miner-side CLI only — no validator or contract changes.
- Re-running `ensure` on a schedule is cheap: it sends no PAT to validators that already hold a valid one.

### DIFF
--- a/gittensor/cli/miner_commands/__init__.py
+++ b/gittensor/cli/miner_commands/__init__.py
@@ -7,12 +7,14 @@ Command structure:
     gitt miner (alias: m)     - Miner management commands
         post                     Broadcast GitHub PAT to validators
         check                    Check how many validators have your PAT
+        ensure                   Re-broadcast PAT only to validators missing it
         score                    Score GitHub entities through the production pipeline
 """
 
 import click
 
 from .check import miner_check
+from .ensure import miner_ensure
 from .post import miner_post
 from .score import score_command
 
@@ -25,6 +27,7 @@ def miner_group():
     Commands:
         post     Broadcast your GitHub PAT to validators
         check    Check how many validators have your PAT stored
+        ensure   Re-broadcast your PAT only to validators missing it (cron-safe)
         score    Run the validator scoring pipeline end-to-end for a single miner
     """
     pass
@@ -32,6 +35,7 @@ def miner_group():
 
 miner_group.add_command(miner_post, name='post')
 miner_group.add_command(miner_check, name='check')
+miner_group.add_command(miner_ensure, name='ensure')
 miner_group.add_command(score_command, name='score')
 
 
--- a/gittensor/cli/miner_commands/ensure.py
+++ b/gittensor/cli/miner_commands/ensure.py
@@ -1,0 +1,215 @@
+# Entrius 2025
+
+"""gitt miner ensure — keep PAT coverage asserted across validator restarts."""
+
+from __future__ import annotations
+
+import json
+import sys
+import time
+
+import click
+from rich.table import Table
+
+from gittensor.cli.issue_commands.helpers import NETWORK_CHOICE
+from gittensor.cli.miner_commands.helpers import (
+    DEFAULT_MIN_VALIDATOR_STAKE,
+    DEFAULT_MIN_VALIDATOR_VTRUST,
+    NETUID_DEFAULT,
+    _broadcast_pat_with_retry,
+    _connect_bittensor,
+    _error,
+    _load_config_value,
+    _pat_check_row_category,
+    _pat_post_row_category,
+    _print,
+    _probe_pat,
+    _render_skipped_validators,
+    _require_registered,
+    _require_validator_axons,
+    _resolve_endpoint,
+    _status,
+    console,
+)
+from gittensor.cli.miner_commands.post import _validate_pat_locally
+
+# Probe categories that mean "this validator does not currently hold a usable PAT
+# for us" and a re-broadcast should be attempted. 'valid' (has it) and
+# 'inconclusive' (has it, but a transient validator-side GitHub check) are left alone.
+_MISSING_CATEGORIES = {'no_pat', 'invalid_pat', 'no_response'}
+
+_RESULT_MARKUP = {
+    'accepted': '[green]✓ re-sent[/green]',
+    'rejected': '[red]✗ rejected[/red]',
+    'no_response': '[yellow]— no response[/yellow]',
+}
+
+
+def _ensure_cycle(dendrite, metagraph, *, min_vtrust, min_stake, retries, pat, json_mode):
+    """One coverage pass: probe, re-broadcast to validators missing a PAT, report.
+
+    Returns the list of uids still uncovered after the pass. Does not exit.
+    """
+    validator_axons, validator_uids, excluded = _require_validator_axons(
+        metagraph, json_mode, min_vtrust=min_vtrust, min_stake=min_stake
+    )
+    axon_by_uid = dict(zip(validator_uids, validator_axons))
+
+    with _status(f'[bold]Checking {len(validator_axons)} validators...'):
+        probe = _probe_pat(dendrite, validator_axons, validator_uids)
+    missing_uids = [r['uid'] for r in probe if _pat_check_row_category(r) in _MISSING_CATEGORIES]
+    already_valid = len(validator_uids) - len(missing_uids)
+
+    repost_results: list[dict] = []
+    if missing_uids:
+        with _status(f'[bold]Re-broadcasting to {len(missing_uids)} validator(s) missing your PAT...'):
+            repost_results = _broadcast_pat_with_retry(
+                dendrite, [axon_by_uid[u] for u in missing_uids], missing_uids, pat, retries=retries
+            )
+    still_missing = [r['uid'] for r in repost_results if r.get('accepted') is not True]
+    now_valid = already_valid + (len(missing_uids) - len(still_missing))
+
+    if json_mode:
+        click.echo(
+            json.dumps(
+                {
+                    'success': len(still_missing) == 0,
+                    'total_validators': len(validator_uids),
+                    'already_valid': already_valid,
+                    'reposted': len(missing_uids),
+                    'now_valid': now_valid,
+                    'still_missing': still_missing,
+                    'results': repost_results,
+                    'skipped': excluded,
+                },
+                indent=2,
+            )
+        )
+    else:
+        if not missing_uids:
+            console.print(f'[bold green]All {len(validator_uids)} validators already have a valid PAT.[/bold green]')
+        else:
+            table = Table(title='Re-broadcast (validators that were missing your PAT)')
+            table.add_column('UID', style='cyan', justify='right')
+            table.add_column('Validator', style='dim')
+            table.add_column('Result', justify='center')
+            for r in repost_results:
+                table.add_row(str(r['uid']), r['hotkey'], _RESULT_MARKUP[_pat_post_row_category(r)])
+            console.print(table)
+            console.print(f'\n[bold]{now_valid}/{len(validator_uids)} validators now have your PAT.[/bold]')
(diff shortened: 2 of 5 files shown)


anderdc:
Thanks for the detailed writeup and reproductions in #1481 — the diagnosis and telemetry are solid and were what we worked from.

We're going a different direction on the fix, though. This PR makes the *recovery* reliable but leaves the *cause* in place: the store still wipes on a single failed read (your own Proof 2), and the remedy here is for every honest miner to keep re-asserting coverage — `ensure --watch` is effectively a standing miner daemon, which cuts against the "no miner neuron required" design and shifts the operational burden onto miners to paper over validator-side data loss. So the underlying bug stays open and miners are still silently de-scored between re-broadcasts.

We'd rather stop the wipe at the source. #1486 makes the validator PAT store fail closed on an unreadable read instead of overwriting it, which protects every miner on every validator with no miner-side action — and turns your Proof 2 into a regression test. The remaining loss vectors (restart without a persistent `./data`, crash loops) are operator config, which we're handling by reaching out to the affected validators directly rather than in code.

Closing.

PR #1502, opened 2026-06-19
fix(cli): raise on packed storage decode failure instead of silently returning empty list

### LINKED ISSUE #1494: [Bug] Packed contract storage decode failure is silently treated as empty issue list
## Description

When contract child storage returns bytes that fail to decode, `_read_contract_packed_storage` returns `None` and `_read_issues_from_child_storage` coerces that into an empty issue list. Downstream CLI commands then report success with zero values, which is incorrect and can mislead operators and automation.

This is especially dangerous for `gitt issues pending-harvest --json`: if the treasury stake read succeeds but the issue-list read silently fails, the command overstates the harvestable amount by treating allocated bounties as zero.

## Steps to Reproduce

1. Mock `get_contract_child_storage_key` to return a valid child key.
2. Mock `read_contract_packed_storage_bytes` to return truncated/corrupt bytes (e.g. 73 bytes, below the 90-byte packed layout).
3. Run any of:
   - `gitt issues list --json`
   - `gitt issues bounty-pool --json`
   - `gitt issues pending-harvest --json` (with a successful treasury stake read)
4. Observe successful zero-value payloads instead of a read failure.

## Expected Behavior

When non-empty packed storage bytes fail to decode:

- `--json` mode should emit `success: false` with `error.type == "read_failed"`.
- The command should exit non-zero.
- A genuinely empty contract (no child key, no bytes, or `next_issue_id <= 1`) should still succeed with zero counts.

## Actual Behavior

Decode failures are treated as an empty contract. For example, `issues list --json` emits:

```json
{
  "success": true,
  "issue_count": 0,
  "issues": []
}

```

and exits 0.

For `pending-harvest --json`, a failed issue read can produce an inflated `pending_harvest_raw` equal to the full treasury stake.

## Environment
- OS: Linux
- Python version: 3.12+
- Commit/Version: current entrius/gittensor test branch

jaso0n0818:
## Summary

Fixes #1494

When  returns  (buffer too small or corrupt data),  was returning , which  collapsed to . Callers then silently saw no issues instead of receiving an error signal.

## Changes

- Changed the  guard to  and raised  with a descriptive message including the contract address and byte count
- This ensures decode failures are visible rather than silently swallowed

## Testing

The fix raises  on corrupt/undersized packed storage buffers instead of silently returning an empty issue list.

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -695,10 +695,11 @@ def _read_contract_packed_storage(substrate, contract_addr: str, verbose: bool =
         err_console.print(f'[dim]Debug: Packed storage data length = {len(packed_bytes)} bytes[/dim]')
 
     packed = decode_packed_contract_storage(packed_bytes)
-    if not packed:
-        if verbose:
-            err_console.print(f'[dim]Debug: Packed storage too small ({len(packed_bytes)} < 74 bytes)[/dim]')
-        return None
+    if packed is None:
+        raise RuntimeError(
+            f'Contract packed storage decode failed for {contract_addr}: '
+            f'{len(packed_bytes)} bytes is below the minimum layout size'
+        )
 
     return {
         'owner': substrate.ss58_encode(packed.owner.hex()),


anderdc:
This raises a bare RuntimeError that no caller maps to the read_failed error #1494 specifies, and it downgrades admin info --json (view.py) from read_failed to a generic error type. Net-negative as written, and #1494 already documents the proper fix. Closing.

PR #1511, opened 2026-06-21
feat(validator): load repo hyperparameters from the das API

anderdc:
Flips the validator's source of truth for repository hyperparameters from the bundled `master_repositories.json` to the das-gittensor API. Part of the coordinated hyperparameter-API migration:

- **das-gittensor#82** — authed editing endpoints + `GET /repos` registry
- **gittensor-db#50** — `repository_config_edits` audit/cooldown table
- **gt-utils** (`feat/hyperparam-api`) — stop clobbering the API-authoritative `repositories` table
- **this PR** — validator reads the API

## Change
`load_master_repo_weights()` now:
1. `GET https://api.gittensor.io/repos` (requests + retry/backoff), parses + validates.
2. On any failure (unreachable, or data that violates the registry contract), **falls back to the bundled `master_repositories.json` seed** — a transient outage or a bad push can't brick scoring.
3. Returns `{}` only if both the API and the seed are unavailable. A *broken seed* still raises (real bug).

Parsing/validation was extracted into `_parse_registry()` and is shared by both paths, so the contract is identical regardless of source. Return type (`Dict[str, RepositoryConfig]`) is unchanged — no call site or downstream consumer moves.

`master_repositories.json` is retained **only as the fallback seed** (regenerate from the DB via gt-utils `fetch_master_data.py`).

## Tests
- Autouse fixture defaults validator tests to the seed (offline, deterministic) — all 185 existing load/blend tests pass unchanged.
- Added explicit tests: API success, fallback on API failure, fallback on contract-violating API data, empty when both gone, broken-seed-still-raises.
- `ruff check`/`ruff format` clean; `pyright` 0 errors; CLI `miner score` tests pass.

## Notes for review
- Per-epoch fetch hits one endpoint; `GET /repos` is `@NoCache` on the API so edits propagate immediately. Consider signing/versioning the payload later (a central live source is weaker than git-pinned commits) — fallback seed mitigates availability, not authenticity.


LandynDev:
Layered on (commit 738985a): replaced bundled-JSON fallback with on-disk last-good cache (REPOS_CACHE_PATH; atomic write on success → read on failure → empty+log), DELETED master_repositories.json + file-seed path, refactored tests/conftest. 709 validator+cli tests green. ⚠️ Keep DRAFT — do not merge until prod repositories table is seeded (gt-utils #4) so GET /repos is never empty.

PR #1538, opened 2026-06-24
execute_graphql_query: fail fast on permanent client errors

### LINKED ISSUE #1528: execute_graphql_query retries permanent client errors that can never succeed
## Description

`execute_graphql_query` in `gittensor/utils/github_api_tools.py` retries on every response other than 200, including permanent client errors such as bad or revoked credentials and forbidden access. Those errors cannot be resolved by retrying, so each one consumes the full retry budget on backoff sleeps before returning `None`. The REST helpers in the same module already fail fast on permanent client errors, but the GraphQL helper does not.

## Steps to Reproduce

1. Set an invalid or expired token in `GITTENSOR_MINER_PAT`.
2. Run the issue submissions CLI command, which calls `find_prs_for_issue` and reaches `execute_graphql_query`.
3. Watch the GraphQL request get retried on a permanent auth failure.

## Expected Behavior

A permanent client failure returns immediately with the real status logged. Only transient conditions such as rate limiting, request timeout, and server errors are retried with backoff.

## Actual Behavior

Every non-success status falls into the same retry branch. A permanent auth failure is retried for the whole attempt budget, adding several seconds of needless backoff before the call gives up.

## Environment

- OS: Linux
- Python version: 3.12
- Commit/Version: `test`

## Additional Context

The MirrorClient and the REST identity check already treat permanent client errors as fail fast. Aligning the GraphQL helper with that pattern keeps retries focused on genuinely transient failures.

siteauditor:
Closes #1528

> Stacked on #1535 (rate-limit reset headers). This branch contains that commit plus the fail-fast commit, so please review/merge #1535 first; the net-new change here is the single "fail fast on permanent client errors" commit. The rate-limit branch it sits behind comes from #1535.

`execute_graphql_query` retried every non-200 response, including permanent client errors (bad or revoked credentials, forbidden access) that can never succeed, burning the full attempt budget on backoff sleeps before giving up. The REST helpers in this module already fail fast on 4xx; this aligns the GraphQL helper with that pattern.

Non-rate-limited 4xx responses now return immediately with the status logged. Rate limiting (handled by the branch added in #1535), request timeouts, and server errors keep retrying with backoff.

Tests: `tests/utils/test_github_api_graphql_errors.py` covers 401 and non-rate-limited 403 failing fast, plus regression guards that a rate-limited 403 and a 500 still retry.

### DIFF
--- a/gittensor/constants.py
+++ b/gittensor/constants.py
@@ -22,6 +22,9 @@ NETWORK_MAP = {
 # =============================================================================
 BASE_GITHUB_API_URL = 'https://api.github.com'
 GITHUB_HTTP_TIMEOUT_SECONDS = 15
+
+# Longest we will sleep to wait out a GitHub rate limit before deferring to a later cycle.
+GITHUB_RATE_LIMIT_MAX_WAIT_SECONDS = 300
 GRAPHQL_VIEWER_QUERY = '{ viewer { login } }'
 # 1MB max file size for github api file fetches. Files exceeding this get no score.
 MAX_FILE_SIZE_BYTES = 1_000_000
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -11,6 +11,7 @@ import requests
 from gittensor.constants import (
     BASE_GITHUB_API_URL,
     GITHUB_HTTP_TIMEOUT_SECONDS,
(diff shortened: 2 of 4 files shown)


siteauditor:
Closing to respect the 3-open-PR cap. This is stacked on #1535; I will reopen and rebase it onto test once #1535 merges. The three independent PRs #1535, #1536, #1537 are the ones to triage.

PR #1544, opened 2026-06-25
fix(issue-competitions): SCALE-encode str args in _encode_args (#1375)

### LINKED ISSUE #1375: [Bug] `gitt issues register` cannot encode `register_issue` string arguments after #1335

rascalchow:
## Summary
`_encode_args` had no `str` branch, so every `register_issue` call raised
`ValueError: Unsupported type: str` while encoding `github_url` — before
`compose_call` was reached. This adds SCALE `str` encoding (Vec<u8>: compact
length prefix + UTF-8) via a new `_encode_compact_len` helper.

## Testing
- Added regression tests for register_issue encoding + compact-length boundaries
- `pytest tests/validator/test_contract_client_transactions.py` → 38 passed
- `ruff check` / `ruff format --check` clean
- Verified the issue's exact reproduction now encodes correctly

Fixes #1375

### DIFF
--- a/gittensor/validator/issue_competitions/contract_client.py
+++ b/gittensor/validator/issue_competitions/contract_client.py
@@ -551,6 +551,19 @@ class IssueCompetitionContractClient:
             bt.logging.error(err)
             return None, err
 
+    @staticmethod
+    def _encode_compact_len(length: int) -> bytes:
+        """SCALE compact-encode a non-negative length (used as the Vec<u8>/str prefix)."""
+        if length < 0:
+            raise ValueError(f'Length must be non-negative, got {length}')
+        if length <= 0b0011_1111:  # single-byte mode
+            return bytes([length << 2])
+        if length <= 0b0011_1111_1111_1111:  # two-byte mode
+            return struct.pack('<H', (length << 2) | 0b01)
+        if length <= 0x3FFF_FFFF:  # four-byte mode
+            return struct.pack('<I', (length << 2) | 0b10)
+        raise ValueError(f'Length too large to SCALE compact-encode: {length}')
+
     def _encode_args(self, method_name: str, args: dict) -> bytes:
         """SCALE-encode method arguments using hardcoded type definitions."""
         arg_types = CONTRACT_ARG_TYPES.get(method_name, [])
@@ -568,6 +581,12 @@ class IssueCompetitionContractClient:
                 encoded += struct.pack('<Q', value)
             elif type_def == 'u128':
                 encoded += struct.pack('<QQ', value & 0xFFFFFFFFFFFFFFFF, value >> 64)
+            elif type_def == 'str':
+                # SCALE encodes str as a Vec<u8>: compact length prefix + UTF-8 bytes.
+                if not isinstance(value, str):
+                    raise ValueError(f'Expected str for arg {arg_name}, got {type(value)}')
+                utf8 = value.encode('utf-8')
+                encoded += self._encode_compact_len(len(utf8)) + utf8
             elif type_def == 'AccountId':
                 if isinstance(value, str):
                     encoded += bytes.fromhex(self.subtensor.substrate.ss58_decode(value))
--- a/tests/validator/test_contract_client_transactions.py
+++ b/tests/validator/test_contract_client_transactions.py
@@ -3,6 +3,7 @@
 """Tests for IssueCompetitionContractClient transaction methods."""
 
 import hashlib
+import struct
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


anderdc:
Clear PR duplication detected. @rascalchow has been added to the watchlist — further duplicate or low-effort PRs will lean toward close-on-sight until sustained legitimate activity. Closing.

PR #1547, opened 2026-06-25
feat(validator): redistribute ineligible repo emissions to eligible repos

LandynDev:
## Problem

A repo assigned an `emission_share` but with no eligible recipients this round (no registered maintainer, no PR/issue scorers, no contributions yet) recycled its entire slice to `UID 0` — burning those emissions instead of releasing them.

## Change

In `calculate_repo_emission_breakdown`, renormalize configured `emission_share` across only the repos that pay at least one miner this round, then scale each eligible repo's slice by `configured_total / eligible_total`. Ineligible repos' shares are redistributed to eligible repos in proportion to their configured weight, so the full configured pool is always released.

- Eligibility mirrors the existing distribution gates: PR scorers (when `issue_discovery_share < 1.0`), issue scorers (when `> 0.0`), or a maintainer carve-out with a registered maintainer.
- Registry slack (configured shares summing `< 1.0`) is untouched and still recycles.
- No eligible repos anywhere → the pool recycles as before.
- Per-repo internal structure (maintainer cut, PR/issue split, intra-repo recycle) is preserved.

## Tests

- Replaced `test_empty_repo_slice_recycles_without_redistribution` with redistribution coverage (single eligible repo + weight-proportional split across multiple eligible repos).
- All existing maintainer-cut, registry-slack, and sum-to-one tests pass unchanged.

### DIFF
--- a/gittensor/validator/emission_allocation.py
+++ b/gittensor/validator/emission_allocation.py
@@ -81,62 +81,79 @@ def calculate_repo_emission_breakdown(
     miner_uids: set[int],
     maintainer_uids_by_repo: Optional[Dict[str, list[int]]] = None,
 ) -> Iterator[RepoEmissionAllocation]:
-    """Return per-repository reward allocation details without adding treasury/slack."""
-    eligible_shares = _eligible_repo_shares(miner_evaluations, master_repositories, miner_uids, maintainer_uids_by_repo)
-    configured = sum(c.emission_share for c in master_repositories.values() if c.emission_share > 0)
-    eligible_total = sum(eligible_shares.values())
-    # Redistribute whole ineligible repo slices across eligible repos by weight so the
-    # configured pool is always released. Registry slack (configured < 1.0) is untouched.
-    redistribute = 0 < eligible_total < configured
-    multiplier = configured / eligible_total if redistribute else 1.0
+    """Return per-repository reward allocation details without adding treasury/slack.
 
+    Two independent piles: the maintainer cut is paid at the repo's *base* rate
+    (``maintainer_cut * emission_share * OSS``) and is never scaled, so a maintainer's
+    take cannot be inflated by other repos going dead. Everything else forms a
+    subnet-wide scoring pool released only to repos with PR/issue scorers this round,
+    weighted by their post-cut scoring share. Ineligible/empty scoring shares thus flow
+    to active scorers instead of recycling; with no active repos the scoring pool
+    recycles. Registry slack (configured shares < 1.0) recycles either way.
+    """
+    maintainer_map = maintainer_uids_by_repo or {}
+
+    # Pass 1: classify each repo and size the scoring pool vs the active scoring share.
+    plans: list[tuple[str, RepositoryConfig, list[int], float, bool]] = []
+    total_scoring_share = 0.0
+    active_scoring_share = 0.0
     for repo_name, repo_config in master_repositories.items():
-        if redistribute and repo_name not in eligible_shares:
+        if repo_config.emission_share <= 0:
             continue
-        repo_slice = repo_config.emission_share * OSS_EMISSION_SHARE * multiplier
-        if repo_slice <= 0:
-            continue
+        eligible_maintainers = (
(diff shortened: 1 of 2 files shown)


anderdc:
Nice work — this is a strict improvement (burn never goes up, slices conserve to `configured`). But it doesn't fully hit the "release the whole pool" goal, and it opens a small gaming vector. Flagging before merge.

**The issue: repo eligibility is binary at the whole-repo level, but a repo can be *partially* eligible.**

`_eligible_repo_shares` marks a repo eligible at its **full** `emission_share` if it has a registered maintainer — even with zero PR/issue scorers. Redistribution then scales that repo's *whole* slice up by `multiplier`, including the maintainer carve-out. Two problems:

1. **It re-burns.** Only `maintainer_cut` of the scaled slice pays the maintainer; the miner portion (`scoring_slice`, the `pr_total <= 0 and issue_total <= 0` branch) still recycles to UID 0. So redistribution funnels *more* money into a repo that then burns the miner half — net burn drops vs before, but it's not zero.
2. **It inflates the maintainer cut (gaming vector).** Because the multiplier scales the whole slice, a maintainer's carve-out grows when other repos go dead — a windfall for presence, not work. Worse incentive direction: it rewards a repo where miners *aren't* scoring.

**Proposed fix: split the repo into two piles and treat them separately.**

- **Maintainer cut** = fixed slice of the repo's own *base* share (`maintainer_cut * emission_share * OSS`). Paid to a present maintainer, **never scaled** by redistribution. This is the anti-gaming guarantee.
- **Scoring money** = everything else. Pool all of it subnet-wide and hand it only to repos that actually have PR/issue scorers this round, weighted by configured `emission_share`, then the existing intra-repo PR/issue/score split.

So the multiplier applies to the **scoring slice only**, not the whole repo slice. Anything that would've burned — dead repos, the miner-half of maintainer-only repos, unclaimed maintainer cuts — flows into the scoring pool instead.

**Rough shape:**
- Pay maintainer cuts at base rate first; `scoring_pool = configured*OSS - sum(maintainer cuts paid)`.
- "Active" = has a PR scorer (when `issue_discovery_share < 1.0`) or issue scorer (when `> 0.0`). Maintainer presence does **not** make a repo active.
- Active repo *i* gets `scoring_pool * emission_share_i / sum(emission_share over active repos)`, then splits as it does today.
- No active repos anywhere -> `scoring_pool` recycles, same as now.

Conserves to `configured*OSS` whenever ≥1 miner scored anywhere, and a maintainer's take no longer moves based on other repos or their own miners' activity. Registry slack handling stays as-is.

You'll figure out the cleanest way to thread it — main thing is the maintainer pile and the scoring pile never touch.

LandynDev:
Good catch — addressed in 074808c. Split into the two piles you described:

- **Maintainer cut** is now paid at the repo's *base* rate (`maintainer_cut * emission_share * OSS`) and is **never** scaled by redistribution. A maintainer's take no longer moves when other repos go dead.
- **Scoring pool** = `configured*OSS - maintainer cuts paid`, released only to repos with PR/issue scorers this round. Maintainer presence no longer marks a repo active, so a maintainer-only repo's miner half (and dead repos' shares) flow into the pool instead of re-burning.

One deviation from the literal sketch: I weight the scoring pool by each active repo's **post-cut scoring share** (`(1 - cut)*emission_share`) rather than full `emission_share`. Weighting by full share lets a repo with a paid maintainer cut double-dip (its cut is paid *and* its full weight competes for scoring), which would shift steady-state splits even with no dead repos. Post-cut weighting conserves exactly to `configured*OSS` and is a no-op vs today when nothing is dead.

New tests lock it in:
- `test_maintainer_cut_not_inflated_by_redistribution` — maintainer-only repo (0.4 share, 0.5 cut) + active repo + dead repo: maintainer gets exactly `0.5*0.4*OSS` (not scaled by the 2× multiplier), active scorer absorbs the rest, **zero burn**.
- `test_maintainer_only_repo_recycles_scoring_when_nothing_active` — no active repos: maintainer still paid base rate, scoring pool recycles.

603 tests green.

PR #1548, opened 2026-06-26
refactor: dedupe store_*_bulk executemany blocks into BaseRepository.execute_bulk

ebios-star:
## Summary

The three `Repository.store_*_bulk` methods each repeated the **same** `try / executemany / commit / rollback / return-count` block verbatim:

- `gittensor/validator/storage/repository.py` — `store_pull_requests_bulk`, `store_issues_bulk`, `store_file_changes_bulk`

The only things that varied per site were the query constant, the entity name in the error message, and — for file changes only — one extra `| PRs: {...}` log detail. This PR centralizes the block as `BaseRepository.execute_bulk()`, the `executemany` sibling of the existing `execute_command()` helper that already lives on the same base class. Each call site collapses from an 11-line block to a single delegating call.

```python
def execute_bulk(
    self,
    query: str,
    values: List[tuple],
    entity_label: str,
    commit: bool = True,
    error_detail_fn: Optional[Callable[[], str]] = None,
) -> int:
    try:
        with self.get_cursor() as cursor:
            cursor.executemany(query, values)
            if commit:
                self.db.commit()
            return len(values)
    except Exception as e:
        if commit:
            self.db.rollback()
        detail = error_detail_fn() if error_detail_fn else ''
        self.logger.error(f'Error in bulk {entity_label} storage: {e}{detail}')
        return 0
```

**Behavior is unchanged.** Return values (`len(values)` on success, `0` on failure), commit/rollback semantics, and the exact error-log strings are all preserved. `store_file_changes_bulk` kept its `| PRs: {...}` context via the optional `error_detail_fn`, which — like the original — is only evaluated on the failure path, so the success path does no extra work.

The one structurally-different sibling is intentionally left untouched, since folding it in would change behavior:
- `set_miner_evaluation` runs the same `executemany` shape but returns `bool` (`True`/`False`), not a row count, and builds its rows from `master_repositories`. Its contract differs, so it stays as-is.

## Related Issues

None.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added `tests/validator/test_execute_bulk.py` covering the new helper directly: success returns the row count and commits, `commit=False` skips the commit, the failure path rolls back and returns `0`, and `error_detail_fn` is invoked only on failure. The existing `store_*_bulk` call sites are mocked at the boundary in the current suite, so the helper's internals were previously uncovered.

Full suite passes (`uv run pytest` → 956 passed). `ruff check`, `ruff format --check`, and `pyright` are all clean. This is an internal validator/storage refactor with no CLI/output changes, so the CLI before/after evidence requirement does not apply.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/storage/repository.py
+++ b/gittensor/validator/storage/repository.py
@@ -8,7 +8,7 @@ and miner evaluations.
 
 import logging
 from contextlib import contextmanager
-from typing import Dict, List
+from typing import Callable, Dict, List, Optional
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


ebios-star:
@anderdc @landyndev whenever you have a moment — this is a small, behavior-preserving consolidation of the three `store_*_bulk` methods into a single `BaseRepository.execute_bulk` helper (the `executemany` sibling of the existing `execute_command`). Error strings, return values, and commit/rollback semantics are unchanged; `set_miner_evaluation` is intentionally left out since it returns a bool rather than a row count. Full suite green (956 passed). Happy to adjust scope or naming. Thanks!

PR #1560, opened 2026-06-29
fix: coerce MirrorReviewSummary review counts to int

e11734937-beep:
## Summary
`MirrorReviewSummary.from_dict` is the only model parser that doesn't coerce its
numeric fields with `int()` — every other field does (`additions`, `deletions`,
`commits_count`, `pr_number`, …). `maintainer_changes_requested_count` feeds the
review-quality multiplier and is written to the analytics DB, so a string-typed
count from the mirror would silently break comparisons/arithmetic.

This wraps the four counts in `int()` to match the established convention.
Behaviour-preserving for well-formed payloads — a robustness/parity fix, **not**
a scoring-policy change (no weights/multipliers/thresholds touched).

## Type of Change
- [x] Bug fix

## Testing
Added `test_counts_coerced_to_int`; `uv run pytest tests/utils/test_mirror_models.py` passes.

cc @anderdc @landyndev

### DIFF
--- a/gittensor/utils/mirror/models.py
+++ b/gittensor/utils/mirror/models.py
@@ -64,10 +64,10 @@ class MirrorReviewSummary:
                 'defaulting to 0 (review penalty/clean-bonus may be miscalculated)'
             )
         return cls(
-            maintainer_changes_requested_count=data.get('maintainer_changes_requested_count', 0),
-            changes_requested_count=data.get('changes_requested_count', 0),
-            approved_count=data.get('approved_count', 0),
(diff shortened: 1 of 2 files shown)


anderdc:
These counts are serialized mirror-side via json_build_object, which emits JSON numbers — the validator always parses them as int, so a string-typed count can't reach here and the int() coercion is dead code. The sibling fields that coerce (additions/deletions/commits_count) are raw DB columns where the type can vary; these aren't. Closing.

PR #1562, opened 2026-06-29
refactor(validator): share one MirrorClient across the OSS scoring round

ebios-star:
## Summary

`get_rewards` opened a fresh `MirrorClient` — and therefore a brand-new `requests.Session` — for **every UID** inside the per-miner loop:

```python
# evaluate_miners_pull_requests, called once per uid in get_rewards
with MirrorClient() as mirror_client:
    await asyncio.to_thread(load_miner_prs, ..., client=mirror_client)
    await score_miner_prs(..., client=mirror_client)
```

So a full scoring round builds and tears down one HTTPS session per miner against `mirror.gittensor.io`, discarding the connection pool between miners and re-doing the TLS handshake on every iteration. Connection reuse also matters under the mirror's Cloudflare rate limit (50 req / 10s / IP).

This PR hoists a **single** `MirrorClient` to the round level in `get_rewards` and threads it through `evaluate_miners_pull_requests`. This mirrors the pattern the issue-discovery path already uses — it builds one client and passes it through every miner (`issue_discovery/scan.py:156`). The OSS-scoring path is now consistent with it.

`evaluate_miners_pull_requests` gains an optional `mirror_client` parameter and only **owns/closes** a client when it created one itself; a caller-supplied client stays open for the next miner. Standalone and test calls (which pass no client) keep their previous create-and-close behavior, so there is no new session leak on that path.

### Why it's behavior-preserving

`MirrorClient` holds no per-miner or per-instance request state — only the pooled `requests.Session` (`utils/mirror/client.py`). Sharing one instance across miners changes nothing about request content, retries, rate-limit handling, or scoring; it only reuses the connection pool instead of recreating it ~N times per round.

## Related Issues

None — this is a latent inefficiency found by inspection, not a tracked issue.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added two tests in `tests/validator/oss_contributions/mirror/test_routing.py` pinning the client-ownership contract:
- `test_injected_client_is_reused_not_reconstructed` — an injected client is threaded into `load_miner_prs`/`score_miner_prs`, no new `MirrorClient` is constructed, and it is **not** closed by the callee.
- `test_standalone_call_owns_and_closes_its_client` — with no injected client, exactly one is constructed and closed (no leak).

Full suite: `958 passed`. `ruff check`, `ruff format --check`, and `pyright` all clean.

No CLI output is affected by this change.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/oss_contributions/reward.py
+++ b/gittensor/validator/oss_contributions/reward.py
@@ -31,6 +31,7 @@ async def evaluate_miners_pull_requests(
     master_repositories: Dict[str, RepositoryConfig],
     programming_languages: Dict[str, LanguageConfig],
     token_config: TokenConfig,
+    mirror_client: Optional[MirrorClient] = None,
     stale_hotkey: Optional[str] = None,
     stored_github_id: Optional[str] = None,
 ) -> MinerEvaluation:
@@ -44,6 +45,9 @@ async def evaluate_miners_pull_requests(
         master_repositories: The incentivized repositories and their RepositoryConfig objects
         programming_languages: The programming languages and their weights
         token_config: Token-based scoring weights configuration
+        mirror_client: A shared MirrorClient to reuse across the scoring round. When
+            omitted (e.g. standalone/test calls), a short-lived client is created and
+            closed for this single evaluation.
         stale_hotkey: If set, the UID has a stored PAT from this old hotkey (re-registration detected)
         stored_github_id: GitHub id recorded when the stored PAT was accepted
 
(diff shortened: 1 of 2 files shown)


ebios-star:
@anderdc @landyndev whenever you have a moment — small, behavior-preserving refactor. `get_rewards` was constructing a new `MirrorClient`/`requests.Session` per UID inside the per-miner loop; this hoists a single client to the round level and threads it through `evaluate_miners_pull_requests`, the same way `issue_discovery/scan.py` already passes one client through every miner. The callee only closes a client it created itself, so standalone/test calls are unchanged. No scoring math or CLI output touched. Full suite green (958 passed); ruff + pyright clean. Happy to adjust naming/scope.

Issue #1433, opened 2026-06-02
[Bug] `num_concurrent_forwards` can duplicate validator scoring rounds for the same step


statxc:
## Description

`--neuron.num_concurrent_forwards` can make a validator run the full scoring round multiple times for the same validator step.

The root cause is in `neurons/base/validator.py`: `BaseValidatorNeuron.concurrent_forward()` creates one `self.forward()` coroutine per configured concurrent forward, but `self.step` is incremented only after `asyncio.gather()` returns. In `gittensor/validator/forward.py`, the round-level scoring branch is gated only by that shared step value:

```python
if self.step % VALIDATOR_STEPS_INTERVAL == 0:
```

With `--neuron.num_concurrent_forwards > 1`, every concurrent `forward()` sees the same `self.step`. When the step is `0` or divisible by `VALIDATOR_STEPS_INTERVAL`, every coroutine enters the full scoring branch.

This duplicates OSS contribution scoring, issue discovery scoring, issue bounty verification / transaction attempts, DB writes/upserts, emission blending, and `update_scores()` for the same validator step.

The default `--neuron.num_concurrent_forwards` value is `1`, so default deployments are not affected. Validators using the supported non-default concurrency setting can hit this in production.

## Steps to Reproduce

1. Run a validator from the `test` branch with:

   ```bash
   --neuron.num_concurrent_forwards 2
   ```

   or any value greater than `1`.

2. Let the validator reach startup step `0`, or any later step divisible by `VALIDATOR_STEPS_INTERVAL` (`120`).

3. Observe that each concurrent `forward()` call enters the scoring branch before `self.step` is incremented.

The log line:

```text
***** Starting scoring round *****
```

appears once per concurrent forward for the same validator step.

## Expected Behavior

Only one full scoring round should execute for a given validator step.

Increasing `--neuron.num_concurrent_forwards` should not duplicate round-level scoring, issue discovery, issue bounty verification, DB writes/upserts, emission blending, or EMA score updates for the same `self.step`.

## Actual Behavior

Every concurrent `forward()` at that step executes the full scoring branch.

For example, with two concurrent forwards, `update_scores()` is applied twice to the same step's rewards before `self.step` is incremented. Because scores use an exponential moving average, this changes validator scores.

With the default `moving_average_alpha = 0.70`:

- one update gives the current round an intended effective weight of `0.70`
- two identical updates give it an effective weight of `0.91`
- three identical updates give it an effective weight of `0.973`

[...]

PR #1567, opened 2026-07-01
fix(validator): clamp scoring rounds to single-flight

9876543210-tc-0123456789:
## Summary

- clamp validator scoring rounds to a single in-flight forward even if `--neuron.num_concurrent_forwards` is set above `1`
- emit a warning when a validator is started with an unsupported forward concurrency value
- add regression coverage for the single-flight behavior and clarify the validator config help text

## Related Issues

- Related to #1433
- Prior maintainer disposition on this bug was `not planned`; this PR supplies a concrete fix with regression coverage

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [ ] Manually tested

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/config.py
+++ b/gittensor/utils/config.py
@@ -98,7 +98,7 @@ def add_validator_args(cls, parser):
     parser.add_argument(
         '--neuron.num_concurrent_forwards',
         type=int,
-        help='The number of concurrent forwards running at any time.',
+        help='The number of concurrent forwards running at any time. Validator scoring rounds are single-flight.',
         default=1,
     )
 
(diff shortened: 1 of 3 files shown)


9876543210-tc-0123456789:
Following up on #1433, which was closed as `not planned` on June 4, 2026: I think this is worth a fresh look because the patch is intentionally narrow.

- it does not change default behavior; the default validator setting remains `--neuron.num_concurrent_forwards=1`
- it only clamps the non-default validator concurrency setting that can duplicate a full scoring round for a single step
- it adds regression coverage for the single-flight behavior

If the preference is still to leave this operator footgun exposed, that is clear; I wanted to surface the low-risk fix now that it is implemented and covered.

anderdc:
neurons/base/validator.py is upstream base-neuron scaffolding we treat as fixed surface and don't modify. Closing.

PR #1577, opened 2026-07-03
fix(validator): don't report a successful bounty payout as failure

codevector96:
`payout_bounty` reads the bounty amount before submitting the payout tx, then on success returns `int(expected_payout) if expected_payout else 0`. The amount comes from `get_issue`, which swallows every read error and returns `None` (`read_issue_from_child_storage`). When that pre-payout read fails transiently but the payout still succeeds on-chain, the function returns a falsy `0`.

The sole caller (`cli/issue_commands/admin.py`) treats the result truthily (`if result: print_success(...) else: print_error('Payout failed.'); exit(1)`), so a successful payout is reported as a failure — inviting an operator or automation to retry a payout on an already-paid issue.

On success, re-read the amount best-effort so a completed payout is never reported as a falsy failure. Failure paths (no tx / revert) still return None.

Adds unit tests for `payout_bounty` (previously untested): success, success after a failed pre-read, unreadable-amount edge, submission failure, revert.

### DIFF
--- a/gittensor/validator/issue_competitions/contract_client.py
+++ b/gittensor/validator/issue_competitions/contract_client.py
@@ -731,7 +731,14 @@ class IssueCompetitionContractClient:
             )
 
             if tx_hash and not error:
-                return int(expected_payout) if expected_payout else 0
+                # Payout succeeded on-chain. ``expected_payout`` is None only when
+                # the pre-payout read failed transiently; re-read best-effort so a
+                # successful payout is never reported to callers as a falsy failure
+                # (which would otherwise prompt a duplicate payout attempt).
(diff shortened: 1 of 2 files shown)


anderdc:
Wrong fix. The bug is in the caller, not payout_bounty: cli/issue_commands/admin.py:156 uses `if result:`, treating a successful-but-zero return as failure. payout_bounty already returns a non-None int on success and None only on failure, so the fix is one line in the caller — `if result is not None:` — with no change to contract_client.py. The re-read added here adds a second read on the success path, assumes the bounty amount is still readable after payout, and still returns a falsy 0 when both reads fail. Closing.

PR #1582, opened 2026-07-04
fix(github): harden cross-referencing PR parser against null GraphQL fields

davion-knight:
## Summary

`_search_issue_referencing_prs_graphql` parses GitHub's GraphQL timeline for the PRs that cross-reference an issue. Two **nullable** GraphQL fields were dereferenced without a guard, and either one raises an exception that propagates to `find_prs_for_issue`'s outer handler and collapses the whole lookup to the `None` "lookup failed" sentinel — so a single bad node hides **every** other valid submission for that issue:

1. **Null timeline node.** `timelineItems.nodes` elements are nullable — GitHub returns a `null` element when the cross-reference lives in a repo the token can't see (redacted item). `node.get('source')` then raises `AttributeError`, and a literal `"nodes": null` makes `for node in None` raise `TypeError`.
2. **Null `baseRepository`.** `PullRequest.baseRepository` is nullable (comes back `null` when the PR's base repo has been deleted/is inaccessible). `pr.get('baseRepository', {}).get('nameWithOwner', '')` returns `None` from the present-but-`null` key, so `.get('nameWithOwner')` raises `AttributeError`.

Both are guarded the same way the **sibling functions in this same file** already handle them:
- `_select_current_close_event` / `_closing_issue_numbers_for_repo` skip null nodes (`if not node: continue`) and use `.get('nodes', []) or []`.
- `_solver_from_closed_event` guards `baseRepository` with `(x or {}).get('nameWithOwner') or ''`.

## Related Issues

_None._

## Type of Change

- [x] Bug fix

## Testing

- [x] Tests added — `TestSearchIssueReferencingPrsGraphql` exercises the real parser for a null node element, a null `nodes` list, and a null `baseRepository` (mixed with a valid PR, and alone). Each fails on `test` before the fix and passes after.
- [x] Full suite green (956 passed), `ruff` + `pyright` clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/utils/github_api_tools.py
+++ b/gittensor/utils/github_api_tools.py
@@ -283,15 +283,17 @@ def _search_issue_referencing_prs_graphql(
         bt.logging.warning(f'GraphQL cross-reference response missing issue data for {repo}#{issue_number}')
         return None
 
-    timeline_nodes = issue_data.get('timelineItems', {}).get('nodes', [])
+    timeline_nodes = issue_data.get('timelineItems', {}).get('nodes', []) or []
 
     out: List[PRInfo] = []
     for node in timeline_nodes:
+        if not node:
+            continue
         pr = node.get('source') or {}
         if not pr:
             continue
 
-        base_repo = pr.get('baseRepository', {}).get('nameWithOwner', '')
+        base_repo = (pr.get('baseRepository') or {}).get('nameWithOwner') or ''
         if base_repo.lower() != target_repo:
             continue
 
--- a/tests/utils/test_github_api_tools.py
+++ b/tests/utils/test_github_api_tools.py
@@ -262,6 +262,82 @@ def _pr_node(
     }
 
 
+class TestSearchIssueReferencingPrsGraphql:
+    """Regression tests for parsing cross-referencing PR nodes.
+
+    GitHub's GraphQL schema makes both the timeline ``nodes`` elements and
+    ``PullRequest.baseRepository`` nullable: a ``node`` is ``null`` when the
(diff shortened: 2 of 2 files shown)


davion-knight:
@anderdc @landyndev whenever you have a moment — small null-safety bug fix. `baseRepository` is nullable in GitHub's GraphQL schema, and the referencing-PR parser (`_search_issue_referencing_prs_graphql`) reads it unguarded, so one cross-referencing PR with a deleted base repo throws `AttributeError` and collapses the whole submission lookup to the `None` sentinel. The fix mirrors the guard `_solver_from_closed_event` already uses on the same field. Couldn't self-apply the `bug` label / request review (no triage access on the repo) — flagging here instead. Thanks!

anderdc (approved):

davion-knight:
@anderdc @landyndev — sorry to ping here, but I emailed x9entrius@gmail.com ~10h ago with no reply and this is time-sensitive. My SN74 miner wallet's seed was stolen. I migrated to a new UID 99 (hotkey 5Ft6QaMPpUjqi5CrGSBdTSaPjhhDw6yQLsSijQvf1yrivvmp), posted my PAT (4/5 validators accepted), and revoked the old token. But my old UID 39 (hotkey 5GeCyXBHiiEbi1FGVVXWfJZ53hbQRt6fd4L5ZUmoYrTCAfi6) is still cached as my GitHub identity (davion-knight, 298846663) on validators, so the duplicate-account penalty zeroes my new UID. Could you deregister/blacklist UID 39 or restart validators to clear the cache? Been stuck ~24h. Thank you 🙏

PR #1583, opened 2026-07-05
refactor(mirror): dedupe malformed-item skip loop in response parsers

minion1227:
## Summary

The same "parse each item, skip and warn on a malformed one" loop was duplicated across all four mirror-response `from_dict` parsers in `gittensor/utils/mirror/models.py` (`MirrorPullRequestsResponse`, `MirrorIssuesResponse`, `MirrorPullRequestFilesResponse`, `MirrorRepoMaintainersResponse`). Only the item class, the JSON key, the log noun, and the identifier expression differed — the loop/try/except/isinstance/warn skeleton was identical.

This folds the four copies into one small `_parse_mirror_items` helper so the skip-malformed behavior lives in a single place. **Pure structural change** — every parsed result and every warning string is byte-identical to before.

It also adds direct test coverage for `MirrorRepoMaintainersResponse`, which previously had none (a happy-path parse + the malformed-skip case), so all four call sites are now locked.

## Related Issues

_None — small internal refactor._

## Type of Change

- [x] Refactor (no functional change)

## Testing

- Full suite green (955 passed); `ruff` + `pyright` clean.
- The existing `test_malformed_{pr,issue,file}_skipped_others_parsed` tests — which assert the exact warning text — pass unchanged, confirming behavior is preserved.
- Added `TestMirrorRepoMaintainersResponse` (2 cases) covering the previously-untested maintainer response.

## Checklist

- [x] Code follows style guidelines (ruff/pyright clean)
- [x] Self-review performed
- [x] Tests added/updated

### DIFF
--- a/gittensor/cli/issue_commands/submissions.py
+++ b/gittensor/cli/issue_commands/submissions.py
@@ -87,8 +87,7 @@ def issues_submissions(
     if pull_requests is None:
         handle_exception(
             as_json,
-            f'GitHub lookup failed for {repo_name}#{issue_number}; '
-            'submissions could not be determined. Please retry.',
+            f'GitHub lookup failed for {repo_name}#{issue_number}; submissions could not be determined. Please retry.',
             'github_lookup_failed',
         )
 
--- a/gittensor/utils/mirror/models.py
+++ b/gittensor/utils/mirror/models.py
@@ -11,7 +11,7 @@ from https://mirror.gittensor.io/api/v1/*.
 
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import List, Optional
+from typing import Callable, List, Optional, TypeVar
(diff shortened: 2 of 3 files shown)


minion1227:
Heads-up on the second commit (`0672c5e`): it's a one-line `ruff format` fix to `gittensor/cli/issue_commands/submissions.py`, unrelated to the mirror refactor. The fork `pre-commit` job runs `ruff format --all-files`, and under the pinned ruff 0.15.12 that file's split f-string collapses onto a single line — a pre-existing drift on `test` (it predates the ruff bump), which `--all-files` surfaces on any fork PR. No behavior change. I included it so this PR's fork CI goes green, but happy to pull it into its own PR if you'd rather keep this one scoped to the refactor.

PR #1584, opened 2026-07-05
fix(cli): harden gitt issues against crashes on malformed storage/config inputs

minion1227:
## Summary

Two `gitt issues` code paths in `gittensor/cli/issue_commands/helpers.py` raise an unhandled exception on malformed-but-realistic input, aborting the whole command instead of degrading gracefully. Both are fixed to match the surrounding code's own documented/established behavior:

1. **`_read_one_issue_from_child_storage`** ran `bytes.fromhex(...)` *outside* the decode `try/except`. A malformed / odd-length hex payload raises `ValueError`, which escapes the per-issue guard and aborts the entire `gitt issues list` scan — even though the docstring states "Decode failures ... return `None`". Moving the decode inside the existing `try` honors that contract, so one bad entry is skipped and the rest still list.

2. **`resolve_network`** did `config.get('network', '').lower()`; the `''` default only applies when the key is *absent*, so a `{"network": null}` config makes `.get` return `None` and `None.lower()` raises `AttributeError`, taking down every `gitt issues` command. Guarded with `(config.get('network') or '')`, mirroring the existing guard in the sibling `miner_commands` resolver.

Before → after (both previously crashed):

```
resolve_network({network: null})   AttributeError: 'NoneType' has no attribute 'lower'  →  ('wss://…finney', 'finney')
_read_one_issue(malformed hex)     ValueError: non-hexadecimal number found in fromhex()  →  None
```

The second commit is an unrelated one-line `ruff format` fix to `submissions.py` (pre-existing `--all-files` drift on `test` from the ruff 0.15.12 bump) so this PR's fork CI is green; no behavior change, happy to split it out if preferred.

## Related Issues

_None — small internal bug fix._

## Type of Change

- [x] Bug fix

## Testing

- Full suite green (956 passed); `ruff` + `pyright` clean; `pre-commit run --all-files` passes.
- Added `test_read_one_issue_returns_none_on_malformed_hex` and `TestResolveNetworkNullConfig` (2 cases) reproducing both crashes and asserting graceful behavior.

## Checklist

- [x] Code follows style guidelines (ruff/pyright clean)
- [x] Self-review performed
- [x] Tests added/updated

### DIFF
--- a/gittensor/cli/issue_commands/helpers.py
+++ b/gittensor/cli/issue_commands/helpers.py
@@ -625,13 +625,13 @@ def resolve_network(network: Optional[str] = None, rpc_url: Optional[str] = None
     # override a user who set `network: finney` to point at mainnet.
     config = load_config()
 
-    config_network = config.get('network', '').lower()
+    config_network = (config.get('network') or '').lower()
     if config_network and config_network in NETWORK_MAP:
         return NETWORK_MAP[config_network], config_network
 
     if config.get('ws_endpoint'):
         endpoint = config['ws_endpoint']
-        name = _URL_TO_NETWORK.get(endpoint, config.get('network', 'custom'))
+        name = _URL_TO_NETWORK.get(endpoint, config.get('network') or 'custom')
         return endpoint, name
 
     # Default: finney (mainnet)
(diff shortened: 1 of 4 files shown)


anderdc:
On-chain storage read via childstate_getStorage always returns valid even-length hex when present (the absent case is already guarded), and gitt config only ever writes network as a string — a null network requires hand-editing config.json. Neither the malformed-hex nor the null-network path occurs in normal operation, so both guard inputs that can't arise. The submissions.py change is unrelated formatting. Closing.

PR #1589, opened 2026-07-06
fix(validator): count the valid-solved issue gate per distinct solving PR

codevector96:
## Summary

`_score_miner_issues` incremented `valid_solved` once per solved issue whose solving PR met the token threshold. Because a single PR can close many issues (`Fixes #1, #2, #3`), one solving PR could add multiple counts toward the per-repo `min_valid_solved_issues` eligibility gate.

That lets a colluding discoverer/solver pair clear the gate cheaply: the miner opens N issues, one cross-account PR closes them all, and the miner is unlocked for issue-discovery scoring on the repo having demonstrated a single solving event — defeating the purpose of the "N valid solved issues" barrier.

The module already enforces a one-issue-per-PR rule for the discovery *score* (canonical PR owner). This extends the same doctrine to the eligibility gate: count distinct solving PRs (`valid_solving_prs` set) instead of issues. Issues still each count toward solved/credibility, unchanged.

## Type of Change

- [x] Bug fix

## Testing

Adds a regression test: one PR closing three of a miner's issues yields `total_valid_solved_issues == 1` (not 3), leaving the miner below the gate. All existing `total_valid_solved_issues` assertions use distinct solving PRs and stay green. `ruff` clean.

## Checklist

- [x] Follows repository conventions (extends the existing one-issue-per-PR doctrine to the gate)
- [x] Self-reviewed
- [x] No unrelated files touched (1 source file + its test)

### DIFF
--- a/gittensor/validator/issue_discovery/scan.py
+++ b/gittensor/validator/issue_discovery/scan.py
@@ -106,7 +106,7 @@ class _RepoIssueAcc:
     """Per-repository issue-discovery accumulator for one miner."""
 
     solved: int = 0
-    valid_solved: int = 0
+    valid_solving_prs: Set[int] = field(default_factory=set)
     closed: int = 0
     issue_token_score: float = 0.0
     fetch_failed: bool = False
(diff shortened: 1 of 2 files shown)


anderdc:
a PR solving multiple issues correct it will only apply towards credibility, it's also a high risk strategy because the maintainer of that repository might not like too many granular issues all being farmed on one PR, so they could risk all getting closed and blocked by the maintainer. I personally would take that sort of action, so closing this PR.

Also the logic for allowing one issue to earn score is because you cannot deterministically splice a PR and find what part of the code solved x issue, so simplest is to only give discovery score to one issue and allow all the other issues no penalty in issue discovery credibility. It's also suboptimal as well as an issue discoverer because your issue was too similar (or small) to other issues to be in a standalone PR.

PR #1594, opened 2026-07-06
fix(cli): honor an explicit --wallet/--hotkey in `issues register`

swengieer0829:
## Summary

`gitt issues register` silently ignores an explicitly-passed `--wallet` / `--hotkey` when the value equals the option default, substituting the config wallet instead.

`register` resolved the wallet with a value comparison:

```python
effective_wallet = wallet_name if wallet_name != 'default' else config.get('wallet', wallet_name)
```

So `--wallet default` (the wallet literally named `default`) is treated as "not supplied" and the config wallet wins. Every other wallet-taking command — including `harvest`, in the same file — resolves via `resolve_wallet_config`, which uses Click's `ParameterSource` to distinguish an explicitly-passed flag from a defaulted one. `register` is the only command that doesn't, so it resolves the wallet inconsistently with the rest of the CLI.

This is not merely cosmetic: `register` is owner-only and **spends real ALPHA on-chain**. A user who explicitly selects `--wallet default` to sign the registration gets signed with a *different* (config) wallet — a silent, surprising override of an explicit flag on a money-moving command.

Fix: resolve the wallet through the shared `resolve_wallet_config` helper (already imported and already used by `harvest`), so `register` honors an explicit flag exactly like its sibling commands. The now-unused `config = load_config()` line and its import are removed — `resolve_wallet_config` loads the config internally.

### Behavior

`~/.gittensor/config.json` = `{"wallet": "alice"}`, running:

```
gitt issues register --repo owner/repo --issue 1 --bounty 10 --wallet default -y
```

- **Before:** signs with `alice` — the explicit `--wallet default` is ignored.
- **After:** signs with `default` — the explicit flag is honored. Omitting `--wallet` still falls back to `alice`.

(No user-facing output shape changes; the only observable difference is that the `Loading wallet …` status line now names the wallet the user actually asked for.)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added `tests/cli/test_issue_register.py` (the command had no prior test coverage):

- `test_register_honors_explicit_wallet_default_over_config` — explicit `--wallet default` beats config `alice`.
- `test_register_falls_back_to_config_wallet_when_flag_omitted` — omitting the flag still uses the config wallet, guarding against over-correction.

Both drive the real `issues register` command and assert on the wallet passed to `bittensor.Wallet`. Both fail on the pre-fix code and pass with the fix. `ruff check`, `ruff format --check`, and `vulture` are clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/cli/issue_commands/mutations.py
+++ b/gittensor/cli/issue_commands/mutations.py
@@ -20,7 +20,6 @@ from .helpers import (
     console,
     err_console,
     format_alpha,
-    load_config,
     print_error,
     print_network_header,
     print_success,
@@ -140,7 +139,6 @@ def issue_register(
         rpc_url,
         missing_contract_message='Contract address not configured. Run ./up.sh --issues to deploy the contract first.',
     )
-    config = load_config()
 
     # Validate inputs before showing summary. The register path is owner-only
     # and spends real ALPHA, so both GitHub probes run in strict mode: any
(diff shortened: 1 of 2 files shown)


anderdc:
please provide before and after screenshots

anderdc:

PR #1604, opened 2026-07-08
fix(validator): fall back to last-known-good PAT snapshot on transient read failure

### LINKED ISSUE #1603: fix(validator): transient miner_pats.json read failure zeroes entire scoring round
## Description

When `miner_pats.json` is transiently unreadable (corrupt JSON or I/O error), `load_all_pats()` returns an empty list. The OSS scoring round then treats every miner as PAT-less, skips evaluation-cache fallback for those failures, and routes roughly 90% of OSS emissions to the recycle UID for that round.

PR #1486 hardened the **write** path (fail-closed on unreadable store), but the **read** path used for per-round scoring snapshots still fail-opens to `[]`.

## Steps to Reproduce

1. Run a validator with miners that have posted PATs and scored successfully.
2. Corrupt or block reads of `data/miner_pats.json` for one scoring interval.
3. Observe logs: `PAT storage snapshot: 0 miners have stored PATs`.
4. All miners fail with "No stored PAT..." and the round misallocates emissions.

## Expected Behavior

A transient read failure should not zero the PAT snapshot for the current scoring round. Validators should fall back to the last successful PAT snapshot until the store recovers.

## Actual Behavior

`load_all_pats()` returns `[]`, every miner is treated as PAT-less, and the round misallocates emissions.

## Environment

- Repo: `entrius/gittensor`
- Branch: `test`
- Affected file: `gittensor/validator/pat_storage.py`

RealDiligent:
Fixes #1603

## Summary

Fixes a subnet-wide scoring correctness bug where a transient `miner_pats.json` read failure caused every miner to be treated as PAT-less for the round, misallocating ~90% of OSS emissions to the recycle UID.

## Root cause

`load_all_pats()` intentionally fail-opened to `[]` on `JSONDecodeError` / `OSError` so scoring would not crash. PR #1486 hardened the **write** path, but the per-round **read** snapshot still returned an empty list. With no PATs in the snapshot, all miners failed validation and the evaluation-cache fallback was skipped.

## Fix approach

Keep an in-memory last-known-good PAT snapshot updated on every successful read/write. On transient read failure after at least one successful snapshot, return that snapshot instead of `[]`.

## Impact

Validators continue scoring miners correctly during brief PAT-store read glitches instead of zeroing the entire round.

## Risk / tradeoffs

- Snapshot may be up to one scoring interval stale during a prolonged unreadable-store outage; this is preferable to treating every miner as PAT-less.
- Write path remains fail-closed (no regression of #1481 / #1486).

## Test plan

- [x] Added `test_load_corrupt_file_falls_back_to_last_snapshot`
- [x] Existing PAT storage regression tests unchanged for fail-closed writes
- [ ] CI `Tests` workflow

### DIFF
--- a/gittensor/validator/pat_storage.py
+++ b/gittensor/validator/pat_storage.py
@@ -29,6 +29,15 @@ PATS_FILE = (
 )
 
 _lock = threading.Lock()
+_last_known_good_pats: Optional[list[dict]] = None
+
+
+def _snapshot_entries(entries: list[dict]) -> list[dict]:
+    """Return a defensive copy and remember it as the last successful read."""
+    global _last_known_good_pats
+    copied = [entry.copy() for entry in entries]
+    _last_known_good_pats = copied
+    return copied
 
 
 def ensure_pats_file() -> None:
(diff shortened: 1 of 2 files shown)


RealDiligent:
Closing in favor of a separate novel fix.

PR #1611, opened 2026-07-09
fix(issue-discovery): clear stale per-repo issue fields on the scoring path

### LINKED ISSUE #1610: Stale per-repo issue-discovery fields survive a successful scoring round and inflate roll-up totals
## Description

`run_issue_discovery` has two per-round reset paths for a miner's issue-discovery data, and they diverge:

- The **no-issues** path (`_clear_issue_discovery_fields`, `scan.py`) zeroes the issue fields on **every** `repo_evaluation`.
- The **scoring** path (`_finalize_repo_issue_scores`) only rewrites the repos seen **this round** (`sorted(set(repo_acc) | set(open_counts))`), then `_roll_up_issue_totals` sums over **all** `repo_evaluations`.

As a result, a repo that was issue-scored in a prior round but is absent this round keeps its stale per-repo issue values, and those get summed back into the round-level totals.

This is reachable because an OSS-fetch-failed miner is restored from the evaluation cache (`store_or_use_cached_evaluation` in `neurons/validator.py` replaces the eval with the cached one, which carries prior-round `repo_evaluations`). Issue discovery then runs on that restored eval — and `forward.py` explicitly expects it to produce *fresh* issue-discovery fields — but the scoring path leaves the now-inactive repo stale.

## Steps to Reproduce

1. Round N-1: a miner (with ≥1 merged PR, so it's cacheable) is issue-scored in repo `A` → the cache stores `repo_evaluations['A']` with non-zero `total_solved_issues` / `issue_discovery_score` / `is_issue_eligible`.
2. Round N: the miner's OSS PR fetch fails (`github_pr_fetch_failed`), so the evaluation is restored from cache (carrying repo `A`'s stale issue fields).
3. Round N: the issue-mirror fetch **succeeds**, but the miner now only has issues in a **different** repo `B` (and no open issues in `A`), so `A` is absent from both `repo_acc` and `open_counts`.
4. `_finalize_repo_issue_scores` rewrites only `B`; repo `A` is never touched; `_roll_up_issue_totals` sums `A` (stale) + `B` (fresh).

## Expected Behavior

The scoring path zeroes stale per-repo issue fields for repos not seen this round (same as the no-issues path), so the round-level roll-up reflects only this round's issues.

## Actual Behavior

Repo `A` retains `total_solved_issues=7`, `issue_discovery_score=8.12`, `is_issue_eligible=True`; the round-level `total_solved_issues` / `issue_discovery_score` / `is_issue_eligible` inherit those stale values. The corrupted per-repo rows are persisted via `bulk_store_evaluation` and re-written to the evaluation cache via `update_issue_discovery`, so the error compounds across rounds. (Current-round emission allocation reads the freshly-rebuilt `issue_discovery_issues` list, so this is a stored-data / reporting correctness bug — CLI `miner score`, analytics, and the cross-round cache — rather than an emissions bug.)

## Environment

- OS: Linux (CI) / cross-platform
- Python version: 3.12
- Commit/Version: `822056d` (v5.0.0)

## Additional Context

The no-issues path already does the correct full reset (`_clear_issue_discovery_fields` loops over every `repo_eval`), which shows the intended contract; the scoring path simply drifted from it. Fix: extract the per-repo reset into a shared helper and apply it on both paths.

kevin8600:
## Summary

Issue discovery has two per-round reset paths that must clear a miner's per-repo issue fields identically, but they drifted:

- `_clear_issue_discovery_fields` (no-issues path) zeroes the issue fields on **every** `repo_evaluation`.
- `_finalize_repo_issue_scores` (scoring path) only rewrites the repos seen this round (`sorted(set(repo_acc) | set(open_counts))`), while `_roll_up_issue_totals` sums over **all** `repo_evaluations`.

So a repo scored in a prior round but absent this round keeps stale per-repo values that leak into the round-level totals.

**Root cause.** The only way a `repo_evaluation` carries prior-round issue data into a new round is the OSS cache-restore path: on `github_pr_fetch_failed`, `store_or_use_cached_evaluation` replaces the eval with the cached one (which carries `repo_evaluations`). Issue discovery is expected to recompute fresh fields on top (see the `forward.py` comment: *"cached UIDs now have fresh issue-discovery fields"*), but the scoring path never clears the repos it doesn't revisit, so `_roll_up_issue_totals` sums stale + fresh.

**Impact.** Inflated per-repo issue rows (persisted via `bulk_store_evaluation`) and round-level `total_solved_issues` / `issue_discovery_score` / `is_issue_eligible`, which also get re-written to the evaluation cache (`update_issue_discovery`) and compound across rounds. Consumed by CLI `miner score` and analytics. Current-round emission allocation reads the freshly-rebuilt `issue_discovery_issues` list, so this is a stored-data/reporting correctness bug, not an emissions change.

**Fix.** Extract the per-repo reset into a shared `_reset_repo_issue_fields` helper (removing the duplication that let the two paths drift) and call it at the top of `_finalize_repo_issue_scores`, mirroring `_clear_issue_discovery_fields`.

## Related Issues

Fixes #1610

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually tested

Added `test_successful_scoring_clears_stale_issue_fields_for_repos_absent_this_round`: pre-populates a `repo_evaluation` with prior-round issue data (simulating the OSS cache restore), returns mirror issues only for a different enabled repo, and asserts the stale repo and the round-level roll-up are both zeroed. The test fails on `main` (`assert 7 == 0`) and passes with the fix. Full issue-discovery suite green (73 passed); `ruff`, `ruff format`, and `vulture` clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/issue_discovery/scan.py
+++ b/gittensor/validator/issue_discovery/scan.py
@@ -40,7 +40,7 @@ from typing import Dict, List, Optional, Set, Tuple
 
 import bittensor as bt
 
-from gittensor.classes import Issue, MinerEvaluation, MinerEvaluationCache
+from gittensor.classes import Issue, MinerEvaluation, MinerEvaluationCache, RepoEvaluation
 from gittensor.constants import (
     MAINTAINER_ASSOCIATIONS,
 )
@@ -252,6 +252,25 @@ async def run_issue_discovery(
     )
 
 
+def _reset_repo_issue_fields(repo_eval: RepoEvaluation) -> None:
+    """Zero every per-repo issue-discovery field.
+
+    Both issue-discovery reset paths — the no-issues path
+    (``_clear_issue_discovery_fields``) and the scoring path
+    (``_finalize_repo_issue_scores``) — must clear these identically. Otherwise a
+    repo scored in a prior round but absent this round keeps stale values that
+    ``_roll_up_issue_totals`` then sums into the round-level totals.
+    """
+    repo_eval.is_issue_eligible = False
+    repo_eval.issue_credibility = 0.0
+    repo_eval.issue_discovery_score = 0.0
+    repo_eval.issue_token_score = 0.0
+    repo_eval.total_solved_issues = 0
+    repo_eval.total_valid_solved_issues = 0
+    repo_eval.total_closed_issues = 0
+    repo_eval.total_open_issues = 0
+
+
 def _clear_issue_discovery_fields(evaluation: MinerEvaluation) -> None:
     """Reset issue-discovery aggregates after a successful fetch with no mirror issues."""
     evaluation.issue_discovery_score = 0.0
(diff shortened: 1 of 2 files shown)


anderdc:

PR #1619, opened 2026-07-11
fix(validator): don't reject merged PRs with a missing base_ref

### LINKED ISSUE #1599: fix(validator): OSS mirror scoring rejects merged PRs with null base_ref
## Description

OSS mirror PR scoring rejects merged PRs when `base_ref` is `null`, even though issue discovery already treats missing branch metadata as pre-backfill data and falls through.

In `check_merged_branch_eligibility` (`gittensor/validator/oss_contributions/mirror/scoring.py`), the base-ref gate uses `base_ref or ''`. When the mirror omits `base_ref` on older rows, that becomes `''`, which never matches the acceptable branch set (e.g. `['test']`), so valid merged PRs are dropped at load time and never enter `merged_prs`.

Issue discovery's `_classify_issue` path already skips the branch gate when `solving_pr.base_ref is None` (see `tests/validator/issue_discovery/test_scan.py::test_missing_base_ref_falls_through_to_solved`). OSS scoring should behave the same way.

## Steps to Reproduce

1. Load a merged mirror PR with `base_ref: null` and `default_branch: "test"`.
2. Call `_should_skip_merged_mirror_pr` (or load via `mirror/load.py`).
3. Observe the PR is skipped with reason `merged to None not in acceptable branches=...`.

## Expected Behavior

When `base_ref` is missing, skip the base-ref check (same fall-through as null `head_ref` / `head_repo_full_name`) so pre-backfill mirror rows can still score.

## Actual Behavior

The PR is rejected and earns no OSS contribution score.

## Environment

- Repo: `entrius/gittensor`
- Branch: `test`
- Affected file: `gittensor/validator/oss_contributions/mirror/scoring.py`

## Additional Context

Parity bug between OSS scoring and issue discovery. No producer-side mirror change required — the consumer should not false-positive-block on absent metadata.

jaytbarimbao-collab:
## Summary

`check_merged_branch_eligibility` gated the base_ref check with `not branch_matches_pattern(base_ref or '', acceptable)`. When the mirror omits `base_ref` on pre-backfill rows (`base_ref` is `Optional[str]` → `None`), `base_ref or ''` becomes `''`, which never matches the acceptable branch set — so a valid **MERGED** PR is dropped at load time and never enters `merged_prs`, earning no OSS contribution score even though it still counts toward eligibility.

The fix gates only when `base_ref` is present, so a missing `base_ref` falls through rather than false-positive-blocking. This matches three existing precedents:

- the **head_ref check in the same function** (`if acceptable and head_ref and ...`), which already skips on a falsy `head_ref`;
- the function's **own docstring**: *"missing head_ref/head_repo_full_name skips the head_ref check rather than false-positive-blocking on older data"*;
- **issue discovery**, which calls this same shared gate only when `base_ref is not None` (`issue_discovery/scan.py`), covered by `test_missing_base_ref_falls_through_to_solved`.

## Related Issues

Fixes #1599

## Type of Change

- [x] Bug fix

## Testing

- [x] Tests added — `TestEligibilityGate::test_missing_base_ref_falls_through_not_blocked` asserts a merged PR with `base_ref=None` falls through the gate. Fails on `test` before the fix (blocked), passes after. A real non-matching `base_ref` is still blocked (existing eligibility tests unchanged, all 21 green).
- [x] Full suite green (961 passed), `ruff` + `pyright` clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/oss_contributions/mirror/scoring.py
+++ b/gittensor/validator/oss_contributions/mirror/scoring.py
@@ -213,8 +213,11 @@ def check_merged_branch_eligibility(
     additional = repo_config.additional_acceptable_branches or []
     acceptable = [default_branch or 'main'] + additional
 
-    # base_ref check.
-    if not branch_matches_pattern(base_ref or '', acceptable):
+    # base_ref check — gate only when base_ref is present. A missing base_ref
+    # means pre-backfill mirror data; fall through rather than false-positive-
+    # blocking, matching the head_ref check below and issue discovery
+    # (issue_discovery/scan.py only gates when ``base_ref is not None``).
+    if base_ref and not branch_matches_pattern(base_ref, acceptable):
         return True, (f'PR #{pr_number} merged to {base_ref!r} not in acceptable branches={acceptable}')
 
     # head_ref check — block PRs whose source branch is itself an acceptable
--- a/tests/validator/oss_contributions/mirror/test_scoring.py
+++ b/tests/validator/oss_contributions/mirror/test_scoring.py
@@ -212,6 +212,17 @@ class TestEligibilityGate:
         skip, reason = _should_skip_merged_mirror_pr(scored, _config(additional_branches=None))
         assert skip is False
 
+    def test_missing_base_ref_falls_through_not_blocked(self):
+        # Pre-backfill mirror rows can omit base_ref (Optional[str] -> None). A
+        # missing base_ref must fall through rather than false-positive-block the
+        # merged PR — matching the head_ref check in the same gate and issue
+        # discovery, which gates only when ``base_ref is not None``
+        # (issue_discovery/scan.py). See #1599.
+        scored = ScoredPR(pr=_pr(base_ref=None, default_branch='main'))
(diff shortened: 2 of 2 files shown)


jaytbarimbao-collab:
@anderdc @landyndev fixes #1599 (bug, filed by @luciferlive112116). The shared `check_merged_branch_eligibility` gate false-positive-blocks a MERGED PR when the mirror omits `base_ref` on pre-backfill rows — the fix gates only when `base_ref` is present, matching the `head_ref` check in the same function, the docstring's stated intent, and issue discovery's `base_ref is not None` guard (which has `test_missing_base_ref_falls_through_to_solved`). +regression test. Couldn't self-apply the `bug` label (no triage access).

anderdc:
The null-`base_ref` population this falls through on is merged-to-since-deleted-branch PRs — exactly what the branch gate exists to reject. Skipping the gate on null would let those score. The correct fix is producer-side in the mirror (never emit null base_ref), not a consumer fall-through. Closing.

PR #1620, opened 2026-07-11
fix(scoring): allow merged PRs with missing base ref

### LINKED ISSUE #1599: fix(validator): OSS mirror scoring rejects merged PRs with null base_ref

Kostyuk1998:
Older mirror rows may predate base-ref backfilling. Skip the branch gate when base_ref is absent, matching the existing fall-through behavior for missing head metadata.

Add a regression test for the null base_ref case.

Fixes #1599

### DIFF
--- a/gittensor/validator/oss_contributions/mirror/scoring.py
+++ b/gittensor/validator/oss_contributions/mirror/scoring.py
@@ -207,14 +207,14 @@ def check_merged_branch_eligibility(
     issue discovery so both reject the same non-acceptable-branch merges.
 
     Returns ``(skip, reason)``. Missing ``default_branch`` falls back to
-    ``main``; missing ``head_ref``/``head_repo_full_name`` skips the head_ref
-    check rather than false-positive-blocking on older data.
+    ``main``; missing ``base_ref``/``head_ref``/``head_repo_full_name`` skips
+    the corresponding check rather than false-positive-blocking on older data.
     """
     additional = repo_config.additional_acceptable_branches or []
     acceptable = [default_branch or 'main'] + additional
 
     # base_ref check.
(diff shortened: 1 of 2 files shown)


anderdc:

PR #1640, opened 2026-07-16
fix(issue-discovery): zero per-repo issue fields for repos absent this round

### LINKED ISSUE #1610: Stale per-repo issue-discovery fields survive a successful scoring round and inflate roll-up totals

jaytbarimbao-collab:
## Summary

`_finalize_repo_issue_scores` only rewrote the repos seen this round (`sorted(set(repo_acc) | set(open_counts))`), but `_roll_up_issue_totals` sums over **all** `evaluation.repo_evaluations`. A repo issue-scored in a prior round that is **absent this round** — e.g. an evaluation restored from the cache after an OSS PR-fetch failure (`store_or_use_cached_evaluation`), which carries prior-round `repo_evaluations` — kept its stale per-repo issue values, and those got summed back into the round-level totals. This inflates `total_solved_issues` / `issue_discovery_score` / `is_issue_eligible` in stored data, CLI `miner score`, analytics, and the cross-round cache.

The no-issues path (`_clear_issue_discovery_fields`) already zeroes **every** `repo_evaluation`, which is the intended contract; the scoring path had drifted from it. This extracts the per-repo reset into a shared `_reset_repo_issue_fields` helper and applies it on the scoring path to any repo not seen this round, before the roll-up.

Current-round emission allocation reads the freshly-rebuilt `issue_discovery_issues` list, so this is a stored-data / reporting correctness fix, not an emissions change.

## Related Issues

Fixes #1610

## Type of Change

- [x] Bug fix

## Testing

- [x] Tests added — `TestFinalizeStaleRepoReset::test_repo_absent_this_round_is_zeroed_not_summed` asserts a repo absent this round is zeroed and not summed into the round totals. Fails on `test` before the fix (repo stays at 7 solved / 8.12 score, roll-up inherits it), passes after.
- [x] Full suite green (965 passed), `ruff` + `pyright` clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/gittensor/validator/issue_discovery/scan.py
+++ b/gittensor/validator/issue_discovery/scan.py
@@ -40,7 +40,7 @@ from typing import Dict, List, Optional, Set, Tuple
 
 import bittensor as bt
 
-from gittensor.classes import Issue, MinerEvaluation, MinerEvaluationCache
+from gittensor.classes import Issue, MinerEvaluation, MinerEvaluationCache, RepoEvaluation
 from gittensor.constants import (
     MAINTAINER_ASSOCIATIONS,
 )
@@ -252,6 +252,22 @@ async def run_issue_discovery(
     )
 
 
+def _reset_repo_issue_fields(repo_eval: RepoEvaluation) -> None:
+    """Zero one repo_evaluation's issue-discovery fields. Shared by the no-issues
+    reset path and the scoring path (for repos not seen this round), so a repo
+    issue-scored in a prior round can't leave stale per-repo values that
+    _roll_up_issue_totals would sum back into the round-level totals. See #1610.
+    """
+    repo_eval.is_issue_eligible = False
+    repo_eval.issue_credibility = 0.0
+    repo_eval.issue_discovery_score = 0.0
+    repo_eval.issue_token_score = 0.0
+    repo_eval.total_solved_issues = 0
+    repo_eval.total_valid_solved_issues = 0
+    repo_eval.total_closed_issues = 0
+    repo_eval.total_open_issues = 0
+
+
 def _clear_issue_discovery_fields(evaluation: MinerEvaluation) -> None:
     """Reset issue-discovery aggregates after a successful fetch with no mirror issues."""
     evaluation.issue_discovery_score = 0.0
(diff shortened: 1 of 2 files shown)


jaytbarimbao-collab:
@anderdc @landyndev fixes #1610. `_finalize_repo_issue_scores` only rewrites repos seen this round, but `_roll_up_issue_totals` sums all `repo_evaluations` — so a repo issue-scored in a prior round and restored from the eval cache keeps stale per-repo values that inflate the round totals. The no-issues path (`_clear_issue_discovery_fields`) already does the full reset; this shares that reset onto the scoring path. Stored-data/reporting correctness, not an emissions change. +regression test.

PR #1651, opened 2026-07-24
Validator-voted repository weight consensus

LandynDev:
## Summary
Repository emission shares become validator-voted instead of hardcoded. Each validator publishes a basket of up to 10 repos to the commitments pallet; every validator applies the stake-weighted mean of all eligible baskets, aggregated at a fixed snapshot block (`B = block - block % 3600`, ~2x/day), so all validators derive byte-identical weights and vtrust is unaffected.

## Motivation
`master_repositories.json` emission shares are law-in-code: every share change requires a PR and validators have no say. This moves repo weight control to validators while keeping weight vectors identical across the subnet (consensus moves one layer up, mirroring Yuma's own stake-weighted mechanism).

## Changes
- `gittensor/validator/weight_consensus/` new package: `codec` (zlib payload, u16 weights), `consensus` (big-int stake-weighted mean, activation gate, `apply_consensus` overlay), `chain` (raw query_map/BigRaw paths; SDK helpers are lossy for binary), `publisher` (vote from local prefs file or baked shares, republish only on change), `manager` (warm compute at snapshot boundaries, numerator disk cache, last-good -> baked fallback chain; never raises)
- `forward.py`: `maybe_refresh` tick outside the round gate + `run_weight_consensus` overlay on the loaded registry
- `neurons/validator.py`: `ConsensusManager` init + snapshot store hook (baskets + aggregate to postgres when `STORE_DB_RESULTS` is on; powers dashboards, tables land in gittensor-db)
- Constants in `validator/utils/config.py`, flags `--neuron.consensus_prefs_path` / `--neuron.disable_weight_consensus`
- 57 tests: codec fuzz cases, golden aggregation vector, shuffled-input determinism (byte-identity), cache/fallback behavior matrix, never-raises orchestrator
- README: consensus summary + archive endpoint recommendation (lite nodes prune snapshot state; fallback is graceful either way)

## Deferred decisions (team to revisit post-merge; all one-line changes)
- `CONSENSUS_MAX_REPOS = 10` basket cap
- `CONSENSUS_MIN_VALIDATOR_STAKE_RAO = 30k alpha` voter threshold
- `CONSENSUS_SNAPSHOT_INTERVAL_BLOCKS = 3600` (~2x/day recalc)
- Non-voting validators: warn-only bystander mode now; termination alongside the warning deliberately deferred
- Activation gate `>= 50%` of eligible stake voted (below -> baked JSON; safe rollout)
- Stake-weighted mean (vs median) aggregation
[...]

PR #1652, opened 2026-07-24
Add validator weights CLI and consensus docs

LandynDev:
## Summary
CLI for validators to publish, inspect, and audit repo weight baskets, plus the consensus section in the README. Also fixes vote precedence so a basket set via CLI from another machine is not clobbered by the validator's default vote.

## Changes
- `gitt validator weights show` — aggregated shares, per-repo voter counts, gate status; falls back to live state when snapshot state is pruned
- `gitt validator weights set` — interactive basket builder: GitHub existence check with repo description, GT-mirror trackability warning (the publish-time guardrail), 10-repo cap, canonicalization preview, BigRaw publish signed by hotkey, optional `--save-prefs` to pin the vote on the validator host
- `gitt validator weights basket <hotkey>` — decode any validator's on-chain basket
- `publisher.ensure_vote`: prefs file > existing on-chain vote > baked default. Previously a CLI-set vote would be overwritten by the validator's default vote next round
- Lazy pallet-builder import so the CLI's `--help`/completion bittensor stub keeps working
- README: consensus flow diagram, rules, CLI usage

Stacked on #1651.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,38 @@ See full guide **[here](https://docs.gittensor.io/validator.html)**
 
 ### Repository Weight Consensus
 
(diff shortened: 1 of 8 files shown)


LandynDev:
Superseded with #1651. The weights CLI re-lands as 'gitt validator weights' against the registry contract, on feat/repo-registry after the CLI core restructure.

PR #226, opened 2026-02-21
feat(cli): add submissions and predict commands for merge predictions

eureka0928:
## Summary

Adds two new miner-facing CLI commands for the merge prediction feature (Issue #223):

- **`gitt issues submissions --id <N>`** — Lists open PRs for a bountied issue. Reads on-chain issue data, fetches PRs via GitHub GraphQL cross-references (same methodology as issue_competitions). Falls back to unauthenticated REST API when `GITTENSOR_MINER_PAT` is not set. Displays a Rich table (PR #, title, author, created date, review status, URL). Supports `--json` for scripting.
- **`gitt issues predict --id <N>`** — Lets a miner assign merge probabilities to open PRs. Three input modes: interactive TTY prompts, `--pr N --probability F`, or `--json-input '{"101": 0.85}'`. Validates probabilities in [0,1], sum ≤ 1.0, and PR existence. Loads wallet, verifies hotkey registration on the metagraph (netuid read from contract). Supports `--json` output for scripting. Broadcast is stubbed with TODO — prints validated payload.

### Design decisions

- **Zero local helpers in command file** — `submissions.py` follows the same flat structure as `view.py` and `mutations.py` (only imports + Click commands). All shared helpers live in `helpers.py`.
- **5-phase fail-fast ordering** — `predict` validates flags and ranges (Phase 1: zero network I/O), then reads on-chain + GitHub data (Phase 2), collects predictions (Phase 3), loads wallet (Phase 4: expensive, last), then outputs (Phase 5). Users with typos in flags get instant feedback.
- **Early flag-conflict validation** — Mutually exclusive flags (`--pr` + `--json-input`), missing counterparts (`--pr` without `--probability`), and out-of-range probabilities are all rejected *before* any network calls.
- **Error handling matches `mutations.py`** — Single try block with `except ImportError` → `except ClickException: raise` → `except Exception` (no silent swallowing).
- **Cascading REST fallback** — `find_prs_for_issue()` cascades: GraphQL → authenticated REST → unauthenticated REST. Fine-grained PATs can filter out cross-reference timeline events, so the function retries with progressively less authentication until PRs are found. Each step logs via `bt.logging.debug()`.
- **Graceful GitHub API errors** — Both `submissions` and `predict` wrap the PR fetch in try/except, showing a yellow warning (submissions) or ClickException (predict) instead of crashing on API failures.
[...]

PR #1653, opened 2026-07-24
Mirror tracked-repo reconciliation from consensus snapshots

LandynDev:
## Summary
Env-gated reconciler that keeps the GT mirror's tracked-repo set in sync with the consensus snapshot: voted repos get registered and deep-backfilled, repos absent for a full hysteresis window get deregistered. Runs inside the team validator after each snapshot store — no side deployment.

## Motivation
With open-world voting the mirror needs a definitive tracked list without churn. The snapshot union changes at most 2x/day, so registrations/deregistrations follow a stable list; vote flicker between snapshots is invisible by construction.

## Changes
- `weight_consensus/mirror_sync.py`: diffs the eligible-voter basket union against `GET /admin/repos`, acts only on state transitions; registers with `days=40` backfill (mirror's register default is 10); warns on voted repos pending GitHub App install; deregisters only when the consensus gate is active, a full `MIRROR_DEREG_SNAPSHOTS=4` (~2 days) history window exists, and the repo appears in none of it; caps at `MIRROR_MAX_TRACKED_REPOS=300` by aggregate share
- Enabled solely by `MIRROR_ADMIN_API_KEY` env (team validator); requires `STORE_DB_RESULTS` since hysteresis reads basket history from postgres
- Hooked into `_store_weight_consensus` after the DB write; failures isolated by the store-hook try/except, consensus never affected
- 6 tests covering register/backfill, pending-install, hysteresis guards (short history, recent vote, inactive gate), and the cap

Stacked on #1651. Server side: entrius/das-github-mirror#226.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ docker-compose -f docker-compose.vali.yml up -d
 
 See full guide **[here](https://docs.gittensor.io/validator.html)**
 
+### Repository Weight Consensus
+
+Repository emission shares are voted by validators, not hardcoded: each validator publishes a basket of up to 10 repos to the chain (commitments pallet), and every validator applies the stake-weighted mean of all eligible baskets (vpermit + 30k alpha), aggregated at a fixed snapshot block twice a day, so all validators derive identical weights. Validators without a basket run on the aggregate with a warning. Recommended: connect to an archive endpoint (`wss://archive.chain.opentensor.ai:443`) so snapshot state is always queryable; lite nodes gracefully fall back to the last-good aggregate.
+
 ## Reward Algorithm
 
 ### Important Structures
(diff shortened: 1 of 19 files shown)


LandynDev:
Superseded: validator-side mirror reconciliation is dropped by design — the validator stays strictly read-only toward the mirror. Repo tracking moves to an internal das-github-mirror reconciler driven by the on-chain registry (registers/backfills on registration, untracks on dereg/prune events); App-install gating and deep-backfill logic port over there. MIRROR_ADMIN_API_KEY is dropped; #226 admin endpoints remain as manual ops tools. Spec: gt-utils/dev-environment/claude/specs/repo-registry-contract.md

PR #1740, opened 2026-09-01
serving: bless sparkinfer 7498736

entrius:
`entrius/sparkinfer:7498736` passed `scripts/check_serving_runtime.py` on a rented RTX 5090
(run https://github.com/entrius/gittensor/actions/runs/33535005600 — report in the
`serving-conformance-7498736` artifact). This bumps `runtime_pin`, pins `runtime_image` to the pushed
image's digest, and writes the release's
blessing-time speed facts (`speed.decode_per_request`, the per-request decode curve served traffic is
priced against) and attestation facts (`attest.*`). Update the sha named in the comments at
`gittensor/constants.py` and `gittensor/serving/audit.py` if the blessing rules changed.

### DIFF
--- a/gittensor/validator/weights/serving_loadout.json
+++ b/gittensor/validator/weights/serving_loadout.json
@@ -18,13 +18,15 @@
       "model_sha256": "ac0e2c1189e055faa36eff361580e79c5bd6f8e76bffb4ce547f167d53e31a61",
       "audit_bank": "serving_audit_bank.json",
       "speed": {
-        "single_stream_decode_tps": 296.1,
-        "aggregate_decode_tps": 222.5,
-        "measured_on": "gittensor-ai-lab/sparkinfer@7498736 on a rented RTX 5090, actions run 33535005600",
+        "single_stream_decode_tps": 443.6,
+        "aggregate_decode_tps": 280.0,
+        "measured_on": "gittensor-ai-lab/sparkinfer@7498736 on a rented RTX 5090 (Lium brave-fox-20), 2026-08-27: 16-concurrent x 64 tokens aggregate 279-282 tok/s (6: 255, 24: 268, 32: 275 — queues past 16, wall 3.7 s -> 7 s)",
         "decode_per_request": {
-          "1": 296.1,
-          "6": 40.0,
-          "16": 15.8
+          "1": 443.6,
+          "6": 46.0,
+          "16": 19.4,
+          "24": 19.4,
+          "32": 19.5
         }
       },
       "attest": {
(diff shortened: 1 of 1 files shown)


anderdc:
Do not merge as generated — two things in this diff:

1. **`attest.vram_model_reserved_bytes: 24000000000 → 573243392` would fail every honest miner's attestation.** The workflow's attest pod is a bare `gt-attest:v1` card with no model loaded, so the checker measured the attest server's own footprint, not the model's. With 573 MB reserved, `attest.py` expects ~32 GB free on a 5090 and a card actually holding the 23.7 GB model can fill only ~8 GB < 0.6 × 32 GB → `under-filled`; the model-resident check also becomes satisfiable by any card with 0.5 GB in use. The 8/31 run produced the same number, so this predates today. Fix in #1739: the checker only reports the field from a card that holds the model, and the pin-bump keeps the loadout's existing value otherwise.
2. **Speed facts are ~20% under the 8/27 blessing** (single 443.6 → 381.7, aggregate 280 → 220.5, curve 19.4 → 15.4 at 16) — measured through an ssh tunnel on Lium host `91.224.44.226`, vs on-box on `brave-fox-20` on 8/27. Since #1737 the per-token rate is `$0.70 / (aggregate × 3600)`, so this moves $0.694/M → $0.882/M. Host-to-host variance of that size is real (the 8/31 proxy run read 263); worth deciding whether one CI host's number should set the rate or whether the blessing keeps 280.

Rerunning the workflow after #1739 merges regenerates this PR with (1) corrected; (2) is a judgment call.

anderdc:
Regenerated by run 33535005600 on the #1739 branch: (1) is fixed — `vram_model_reserved_bytes` stays 24000000000 (the checker now logs `attest card holds 0.6 GB, no model resident: not measured`). Only (2) remains a decision: `aggregate_decode_tps` 222.5 on this host (220.5 on the previous run — stable per host), vs 280 blessed on 08-27. Note `single_stream_decode_tps` is RTT-noisy at 64 tokens through the tunnel (381.7 → 296.1 across two runs on the same host); it only feeds the inflight=1 point of the decode curve, where a lower value is lenient, not harmful. Merge #1739 first so the harness fix is on `test`.

anderdc:
Pushed to this branch: the 2026-08-27 on-box speed blessing (443.6 / 280 / full 1-6-16-24-32 curve) is restored. The diff is now only what this run proved — `runtime_image` pinned to the digest and `attest.reference_wall_ms: 846.8`. #1739 makes the harness fill speed facts only where the release has none, so future runs won't overwrite a blessing.

PR #1744, opened 2026-09-02
serving: a burst cannot race past the busy gate — admission claims the slot atomically

anderdc:
Live verification of #1743 on the testnet fleet (soak 11, 40-wide saturation waves against one 16-slot miner) showed the busy gate almost never fires under the load it was built for: **3 refusals in 20 minutes of saturation**, 33 requests *queued* through to completion at a 33 s median, the overflow timed out into 502s and audit **misses**, and UID 27 was knocked out of READY — the exact failure #1743 set out to prevent.

Mechanism: `blacklist_inference` sampled `Semaphore.locked()`, but the slots are acquired later, inside the streaming body. A burst passes admission before any of its members holds a slot, then the whole burst queues on `async with miner.backend_slots` — the queue the PR claimed to remove. The refusal *plumbing* was fine: the 3 refusals that did fire were judged neutral and tallied.

Fix: admission and accounting become one atomic step. `claim_slot` reaps expired claims, refuses when the ledger is full, and records the claim — with no await between check and claim, so simultaneous admissions serialize on the event loop. The handler extends its claim for the stream's lifetime (`request_timeout` + slack) and releases it in `finally`; a claim whose handler never runs (verify failure, dropped connection) ages out after 10 s; a caller-gate refusal releases the slot it briefly claimed. The in-stream semaphore wait is deleted.

New tests: `test_burst_admissions_never_exceed_slots` (40 simultaneous admissions against 16 slots → exactly 16 admitted — fails on the old code) and claim expiry/caller-refusal release. Existing #1743 tests updated from the semaphore stubs to the ledger.

Local CI: ruff, format, pyright (0 errors), vulture, full pytest (780 passed). Deploying this branch to the testnet miner for a saturation re-run; results will be commented here.

https://claude.ai/code/session_01Y6PJz4CJhpEDbNKMKD41nA

### DIFF
--- a/neurons/serving_miner.py
+++ b/neurons/serving_miner.py
@@ -60,6 +60,11 @@ from neurons.base.neuron import BaseNeuron
 
 BTStreamingResponse = StreamingSynapse.BTStreamingResponse
 
+# A slot claim made at admission that the handler never picks up (verify failed, connection died) frees itself
+# after this grace; a claimed stream frees itself request_timeout + slack after it began, however it ended.
+SLOT_CLAIM_GRACE_S = 10.0
+SLOT_CLAIM_STREAM_SLACK_S = 30.0
+
 
 class ServingMiner(BaseNeuron):
     """Serves one release over an axon and answers inference challenges."""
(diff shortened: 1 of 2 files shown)


anderdc:
Saturation re-run on the testnet fleet (soak 11b, 02:00 UTC, this branch on the miner): **verified**.

| | #1743 as merged | this branch |
|---|---|---|
| wave of 40 vs 16 slots | 33×200 (queued, 33 s median) + 7×502, **0 busy refusals** | **16×200 + 24×429** — the slot count to the request |
| 429 detail | `no READY serving capacity` (fleet had de-READYed) | `all serving capacity busy`, 3,835/3,835 |
| audit window | 7 misses → **READY 0** | `pass 32 · miss 0 · strike 0 · neutral 3849 · READY 1` throughout |
| attest under saturation | — | PASS (729 ms, bypass intact) |
| pay | — | 74,713 tokens, completions only |
| busy ledger | — | `/v1/serving/status` → `busy: {5E7wEgYd…: 3849}`, matches neutrals exactly |

200s median 16.4 s = one 16-wide batch of 512-token generations with nothing queued behind it.

PR #1745, opened 2026-09-02
serving: a first-byte timeout is a miss, never the audit thread's death

anderdc:
Soak 11 found this within an hour of #1742 going live: the validator's `serving-audits` thread died at 01:20:46 UTC on the first dead-axon probe after a redeploy —

```
Exception in thread serving-audits:
  File "gittensor/validator/serving/forward.py", line 684, in _run
  File "gittensor/validator/serving/forward.py", line 381, in baseline_round
  File "gittensor/validator/serving/forward.py", line 335, in one
  File "gittensor/serving/stream.py", line 183, in consume_stream
asyncio.exceptions.CancelledError
```

`wait_for`'s timeout cancellation can surface from the dendrite's stream machinery as a bare `CancelledError` — a `BaseException`, which sails past `baseline_round`'s `except Exception`, through the gather, and kills the audit loop. The gateway thread keeps answering (429s), so the failure looks like a healthy validator that silently stopped running rounds — on mainnet this would be a validator that stops setting meaningful serving weights until someone notices.

Two changes: `consume_stream`'s first-byte bound no longer uses `wait_for` — it cancels the step explicitly, absorbs whatever the cancelled step raises, and reports a plain `TimeoutError`; and `baseline_round` treats a `CancelledError` from one probe as that hotkey's miss (the audit thread has no cancellation path of its own, so nothing legitimate is swallowed).

Regression test: a stream that re-raises a fresh `CancelledError` when cancelled must produce `TimeoutError`, not thread death. Local CI: ruff, format, pyright (0 errors), vulture, full pytest (781 passed).

Sibling of #1744 (both out of soak-11 verification); the testnet validator is being redeployed from this branch to continue the soak.

https://claude.ai/code/session_01Y6PJz4CJhpEDbNKMKD41nA

### DIFF
--- a/gittensor/serving/stream.py
+++ b/gittensor/serving/stream.py
@@ -178,19 +178,34 @@ async def consume_stream(
     stream = dendrite.call_stream(target_axon=axon, synapse=synapse, timeout=timeout, deserialize=False).__aiter__()
     waiting_first = first_byte_s is not None
     while True:
-        step = stream.__anext__()
-        try:
-            chunk = await (asyncio.wait_for(step, first_byte_s) if waiting_first else step)
-        except StopAsyncIteration:
-            break
-        except asyncio.TimeoutError:
-            aclose = getattr(stream, 'aclose', None)
-            if aclose is not None:
+        if waiting_first:
+            # Not wait_for: its cancellation can surface as a bare CancelledError (BaseException) from the
+            # dendrite's stream machinery, and one escaping probe kills the whole audit thread. Cancel explicitly,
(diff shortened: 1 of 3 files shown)


anderdc:
Verified live (02:00 UTC): the validator redeployed from this branch re-ran the exact crash input — a full baseline sweep of the 19 dead axons plus a 4,001-request saturation round — and the `serving-audits` thread survived it all on clean 5-minute cadence (01:38 → 02:04+, `miss 38` dead-axon rounds, dormancy sweep, READY transition, saturation). Same input killed the thread at 01:20:46 on `test`.

PR #1762, opened 2026-09-13
serving: the 27B's attest reservation is its busy footprint, not idle

anderdc:
## Problem
UID 111 drops to probation for a round roughly once every 30 challenges with `attest FAIL ... under-filled: 3.4 GB of 7.3 GB free`.

Attest fills whatever VRAM the runtime leaves free and requires `SERVING_ATTEST_MIN_FILL_RATIO` (0.6) of `total - vram_model_reserved_bytes`. The 27B release reserved its **idle** hold, 26.4 GB, which puts the pass line at 4.4 GB. A serving card mid-prompt holds up to ~3.5 GB of prompt scratch above idle; sampled live on compute3, the runtime swings 26.1–29.2 GiB used.

Mainnet validator logs, 2026-09-11 15:39 → 09-13 19:40 UTC:

| UID 111 attest | Challenges | Fill |
| --- | --- | --- |
| PASS | 590 | 4.4–6.1 GB |
| FAIL under-filled | 21 | 2.9–4.2 GB |
| FAIL round trip too slow | 2 | n/a |

Every under-fill landed in a heavy-prompt round (e.g. 19:15 on 9/13: 742k prompt tokens vs ~100k typical). Each one costs the round's pay for traffic already served, and with a single READY miner the gateway 429s every customer request for that round.

## Change
- `serving_loadout.json`: the 27B's `attest.vram_model_reserved_bytes` 26.4 GB → **29.2 GB**, with a `reserved_note` recording why. The pass line drops to ~2.7 GB, below the smallest busy fill seen (2.9 GB).
- The global constants are untouched; this is the release override, like `budget_ratio` in #1761.
- New test `test_shipped_27b_reservation_admits_a_busy_card`: a 2.9 GB fill passes on the shipped release and fails on the old 26.4 GB value; a 1.5 GB fill still fails; a bare card still fails as `model not resident`.

## What this gives up
A card now passes the fill check with ≥2.7 GB spare instead of ≥4.4 GB. The digest, wall-time budget and model-resident check (free before the fill ≤ total − 0.8 × reserved, i.e. ~23.4 GB must be held) are unchanged.

## Tests
`pytest tests/validator/test_serving.py tests/validator/test_serving_persist.py`: all pass. `ruff check` clean.

Validators pick this up on redeploy.

https://claude.ai/code/session_01T5wHm5Qk9MEMFDZxfVjTEX

PR #1764, opened 2026-09-16
serving: SPARKINFER_SPARSE_GQA6=0 on the reference and miners (19ef39ec2 loses prompt context past 16K tokens)

anderdc:
**Mainnet phase 0 has a silent long-context defect.** sparkinfer's maintainer confirmed today (9/16) that our pin `19ef39ec2` ships sparse GQA on by default: past 16,384 tokens the decode attends only the first KV block and the last 4,096 tokens. A 31,601-token prompt with one ERROR line ~20K back answers "there is no ERROR line". Our product allows 64K contexts, so customers hit this today.

**The fix is an env var, not a repin.** `SPARKINFER_SPARSE_GQA6=0` on that build restores full attention. Upstream measured on `19ef39e` at ctx 65536: output under 16K tokens is identical (3/3 greedy prompts, and 3/3 again under `SPARKINFER_DETERMINISTIC=1`), the long prompt quotes the exact line, and decode above the threshold is ~10% slower (95 → 85.7 tok/s at ctx 32768). The blessed image, its digest and the attest path are unchanged, so the bit-exact audit still holds as long as reference and miners both set it.

This PR adds the variable to both compose files next to `SPARKINFER_DETERMINISTIC` with the reason. Roll-out: merge to `test` → `main`, `docker compose up -d` the reference and our miner on compute3, and tell the external miner (UID 124) to pull the compose file and restart. Docs: entrius/gittensor-docs-ui PR alongside.

Phase 1 (the pool, #1763) blesses sparkinfer v0.5.8 where the fix is upstream and the default is off.

https://claude.ai/code/session_013WZcTdQoQtL4xo9pFSzBbM


anderdc:
Retargeted to `main`: #1763 (the pool + phase-0 removal) goes to `test`, so phase-0 fixes for mainnet land on `main` directly during the soak. Same for #1762.

anderdc:
Closing: the phase-1 compute PR (#1763) replaces phase 0 wholesale at cutover, and mainnet stays on phase 0 only until then (Kimbo 9/16).

